PR #32, opened 2017-08-12
Add intro.ts demo that demos intro tutorial, fix intro to actually converge when training.

### LINKED ISSUE #22: Tutorial (training) problems
First off, while I understand the basic principles of NN, I must admit to being rather new to working with them, so maybe my problems are trivial.

I tried following the [tutorial](https://pair-code.github.io/deeplearnjs/docs/tutorials/intro.html), and copied the code [to jsbin](https://jsbin.com/pexojojesu/edit?html,js,output). The code is changed minimally in the attempt to make it work: assumed `labels` should be an array of shape `[1]` NDArrays as declared in `labelShape` and not array of shape `[3]` NDArrays as initialised in `const labels`, and that the second `const inputTensor` should really be `const labelTensor`. (Also, trivially, I removed type annotations and added the namespace to make it run directly in JavaScript.)

However, I can't get anything close to the expected label from this code.

I tried increasing the number of examples to 1000 (random autogenerated, with calculated labels). I tried increasing the batch size. I tried training in a loop. It all ended badly (i.e. with huge cost, and prediction very far from the expected value).

I tried to look up the documentation on `Session.prototype.train`; I couldn't work out which data is being processed, how many iterations are being done, or whether repeating it would add more training data or start from scratch (and if training is meant to be repeated, does the data need to be reshuffled)...

All in all, would it be possible to add an executable version of the training tutorial so we can see it in action (and how to write it correctly)? I can see the library itself works in Demos, but those are a bit too complex for basic study.

Thank you

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/32)

Fixes #22


dsmilkov:
:lgtm_strong:

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions.

---

*[demos/intro/index.html, line 1 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/32#-KrNGeiHcbrNWrKITr4b:-KrNGeiHcbrNWrKITr4c:b-204muu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d00de1ac6acc0b742efc5213547846daaf1367a7/demos/intro/index.html#L1)):*
> ```HTML
> <h1>Introduction tutorial code</h1>
> ```

Add copyright header

---

*[demos/intro/intro.ts, line 39 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/32#-KrNGkoUzchP_aAV5Lff:-KrNGkoUzchP_aAV5Lfg:b5dvm4y) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d00de1ac6acc0b742efc5213547846daaf1367a7/demos/intro/intro.ts#L39)):*
> ```XML
>     // Blocking call to actually read the values from average. Waits until the
>     // GPU has finished executing the operations before returning values.
>     console.log('mean squared difference: ' + average.get());  // average is a Scalar so we use .get()
> ```

wrap to 80 width

---

*[demos/intro/intro.ts, line 58 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/32#-KrNGpQRcqSxFKVLXuOX:-KrNGpQRcqSxFKVLXuOY:b-g27gij) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d00de1ac6acc0b742efc5213547846daaf1367a7/demos/intro/intro.ts#L58)):*
> ```XML
>   const labelTensor = g.placeholder('label', labelShape);
> 
>   // Variables are containers that hold a value that can be updated from training.
> ```

wrap line

---

*[demos/intro/intro.ts, line 133 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/32#-KrNHFgs9-dDnSGTHTtY:-KrNHFgs9-dDnSGTHTtZ:bwkwuf3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d00de1ac6acc0b742efc5213547846daaf1367a7/demos/intro/intro.ts#L133)):*
> ```XML
>     console.log('value: ' + testOutput.get(0));
>   });
> }
> ```

should we dispose all inputs and labels here? or too much detail for this tutorial?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/32#-:-KrNHRaRpupBJcAEMNzr:b-jbsf8z)*

PR #34, opened 2017-08-13
Fix 1) tanh underflow error, 2) scoping bug and add 3) debug mode for math

dsmilkov:
- Fix tanh overflow error where tanh(-100) was NaN.
- Fix scope bug where scope's return result of type NDArray[] was not kept in the parent scope
- Add debug mode for math, where it throws an error the first time a NaN entry is found.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/34)


nsthorat:
:lgtm_strong:

---

Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/math/math.ts, line 67 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34#-KrQG1FapFUS7tmDWSwq:-KrQG1FapFUS7tmDWSwr:b-n82qdy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/857568ad0a49a8327526150098557e6fadff7eac/src/math/math.ts#L67)):*
> ```XML
>    * and checked for NaNs. This significantly impacts performance.
>    */
>   enableDebugMode() {
> ```

I wonder if we should just call this enableNaNDetection or something more than just "enableDebugMode", unless you can think of some other things debug mode will do. For any other xxx detection we could have a bit for that as well. WDYT?

---

*[src/math/math.ts, line 159 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34#-KrP_hL7nUpXQsNMhTbB:-KrP_hL7nUpXQsNMhTbC:bpss9yc) ([raw file](https://github.com/pair-code/deeplearnjs/blob/857568ad0a49a8327526150098557e6fadff7eac/src/math/math.ts#L159)):*
> ```XML
>   }
> 
>   private checkForNaN(arr: NDArray): void {
> ```

can you add a unit test for this?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34#-:-KrQGhXCY5sazBtaimRo:b-jbsf8z)*

nsthorat:
Reviewed 1 of 4 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34)*

dsmilkov:
Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[src/math/math.ts, line 67 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34#-KrQG1FapFUS7tmDWSwq:-KrQUX7dy_hK1HmTUNv9:bxft5p1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8d05973fa2ddfb0c07a4af4455cecde62a65a1fc/src/math/math.ts#L67)):*

I wonder if we should just call this enableNaNDetection or something more than just "enableDebugMode", unless you can think of some other things debug mode will do. For any other xxx detection we could have a bit for that as well. WDYT?
</blockquote></details>

I can imagine later adding other things to debug mode and not having the user call several debug methods. For example, since we are already downloading all of the values to CPU, we might want to console log [min, 5%-tile, median, 95%-tile, max] for each tensor, log the operation of the last math call (obtained using console.trace()), log the execution time (when we add gpu callbacks in webgl 2.0) etc

---

*[src/math/math.ts, line 159 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34#-KrP_hL7nUpXQsNMhTbB:-KrQYypVFi0bR28AVCVx:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8d05973fa2ddfb0c07a4af4455cecde62a65a1fc/src/math/math.ts#L159)):*

can you add a unit test for this?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/34)*

PR #70, opened 2017-08-28
Update build-demo to take a path where bundle.js is written. This fixes the make-website script.

nsthorat:
Before, bundle.js was written in master to demos/$DEMO/bundle.js after the copy to $TMP. This means bundle.js never gets updated.

Now, build-demo takes an optional second param, where to store the bundle, so we can write the bundle to the TMP dir.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/70)


dsmilkov:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/70#-:-Kse9irR1S_8z8Pj1XKk:b-jbsf8z)*

Issue #55, opened 2017-08-18
Still errors on tutorial pages


amadanmath:
Continuing on #22... `intro.ts` got added to give an executable equivalent to the tutorial page code, and it works nicely (thank you!), now I have a problem with the [Guide for ML beginners page](https://pair-code.github.io/deeplearnjs/docs/tutorials/ml_beginners.html).

I replicated the current beginner page code on [jsbin](https://jsbin.com/pexojojesu/edit?js,output) again, with the only change (that I can see) being namespacing; but I am getting `Uncaught Error: Placeholder node "y label" [] not present in feed dictionary.` from the `session.train(...)` call.

By analogy with the working code, I figured that the feed entries should be `[{tensor: x, data: xProvider}, {tensor: yLabel, data: yProvider}]` instead of `[{tensor: x, data: xProvider}, {tensor: y, data: yProvider}]`; but this resulted in `Uncaught Error: tensor 12 not in array map.`, so maybe I got it wrong. (What is "tensor 12" anyway?)

nsthorat:
Ah I'm sorry, will fix this soon, I definitely should have actually ran these :)

As for your second comment, every Tensor gets a unique ID (the 'id' property). That error is telling us that we're trying to provide a Tensor that doesn't exist in the transitive dependencies of the Tensor we are evaluating. What you're saying is exactly right.

jeff3dx:
Yah first thing I did was try to run the code on page "Guide for non-ML experts" and got the same error: "Uncaught Error: Placeholder node "y label" [] not present in feed dictionary."  A fix would be appreciated.

nsthorat:
Working on this fix. There's a problem at head that is causing this error!

nsthorat:
This should be fixed.

PR #80, opened 2017-08-30
Remove split nodes.

### LINKED ISSUE #78: SplitNodes added multiple times if using different runtimes.
If a model has two different runtimes (evaluating cost vs y), and there's a SplitNode added, it will get added multiple times.

From our tutorial code:
```
  const graph = new Graph();
  // Make a new input in the graph, called 'x', with shape [] (a Scalar).
  const x: Tensor = graph.placeholder('x', []);
  // Make new variables in the graph, 'a', 'b', 'c' with shape [] and random
  // initial values.
  const a: Tensor = graph.variable('a', Scalar.new(Math.random()));
  const b: Tensor = graph.variable('b', Scalar.new(Math.random()));
  const c: Tensor = graph.variable('c', Scalar.new(Math.random()));
  // Make new tensors representing the output of the operations of the quadratic.
  const order2: Tensor = graph.multiply(a, graph.square(x));
  const order1: Tensor = graph.multiply(b, x);
  const y: Tensor = graph.add(graph.add(order2, order1), c);
  
  // When training, we need to provide a label and a cost function.
  const yLabel: Tensor = graph.placeholder('y label', []);
  // Provide a mean squared cost function for training. cost = (y - yLabel)^2
  const cost: Tensor = graph.meanSquaredCost(y, yLabel);
```

If y is evaluated, then cost is trained on, it will throw an error because two split nodes were added for the output of x.

The problem is that we add the nodes in the first pass, and then when constructing the unordered list we walk the Node DAG again. The old SplitNode from the other evaluation set is in that list, and thus another SplitNode is added, creating 2 SplitNodes.

nsthorat:
This change removes SplitNodes.

We now backprop through a node with multiple consumers by using a SummedTensorArrayMap. When gradients are set during backprop, we immediately sum.

This change also adds the demo code for the ML beginners guide now that this bug is fixed.

This fixes #78 and #55.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/80)


dsmilkov:
:lgtm_strong: Awesome work!

---

Reviewed 31 of 50 files at r1, 18 of 18 files at r2.
Review status: 49 of 50 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/session.ts, line 183 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspbUNowSmvnvV2bkZt:-KspbUNowSmvnvV2bkZu:begwl90) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/session.ts#L183)):*
> gradients.nullify(costTensor);

do we really need to nullify here, or is this just a pre-caution? In other words, is the gradient of the costTensor disposed somewhere in the previous train call?

---

*[src/session_test.ts, line 159 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspX6oXskpTp7xNv4A2:-KspX6oXskpTp7xNv4A3:b40urxu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/session_test.ts#L159)):*
> ·····

I see a few trailing whitespaces. Feel free to edit workspace settings for vscode and add: "files.trimTrailingWhitespace": true,
in this PR to avoid this in the future (I had it in my user settings, but not workspace settings)

---

*[src/tensor_array_map.ts, line 126 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspafQSjjiKNnopDQJM:-KspafQTbKpSTJKFU2X9:b-54dbg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/tensor_array_map.ts#L126)):*
> add(math: NDArrayMath,

What do you think about passing the math object in the c-tor so we don't have to pass it in on each add call. cleans up the API.

sorry this change might be annoying since it involves 40+ files :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-:-Kspay4e9HAxhrGdyyF7:bdil14w)*

nsthorat:
Review status: 49 of 50 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

*[src/session.ts, line 183 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspbUNowSmvnvV2bkZt:-KspmRFxzskKkw0lQN8g:bzcqr9h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/session.ts#L183)):*

> gradients.nullify(costTensor);

do we really need to nullify here, or is this just a pre-caution? In other words, is the gradient of the costTensor disposed somewhere in the previous train call?
</blockquote></details>

We do need to nullify. The last gradient will be the oneScalar from the previous run unless we null it out (then gradients.add will just be a set, instead of an add). It's not about disposing :)

---

*[src/session_test.ts, line 159 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspX6oXskpTp7xNv4A2:-Kspmg6uC0xueAtjwrFN:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/session_test.ts#L159)):*

> ·····

I see a few trailing whitespaces. Feel free to edit workspace settings for vscode and add: "files.trimTrailingWhitespace": true,
in this PR to avoid this in the future (I had it in my user settings, but not workspace settings)
</blockquote></details>

Done.

---

*[src/tensor_array_map.ts, line 126 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspafQSjjiKNnopDQJM:-KspnejValomafVf0fe8:bi39pou) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/tensor_array_map.ts#L126)):*

> add(math: NDArrayMath,

What do you think about passing the math object in the c-tor so we don't have to pass it in on each add call. cleans up the API.

sorry this change might be annoying since it involves 40+ files :(
</blockquote></details>

So I did it this way since it was the way most of our other functions handle math and if we do it via the constructor, we'll break that CPU / GPU switch. Maybe that's not a big deal? WDYT?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80)*

nsthorat:
Review status: 18 of 51 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[src/tensor_array_map.ts, line 126 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80#-KspafQSjjiKNnopDQJM:-KstCryUFD4oiSayEJH5:b93enso) ([raw file](https://github.com/pair-code/deeplearnjs/blob/365dedaac4bc49ed8d51259b47dbe270d0e88ac4/src/tensor_array_map.ts#L126)):*

So I did it this way since it was the way most of our other functions handle math and if we do it via the constructor, we'll break that CPU / GPU switch. Maybe that's not a big deal? WDYT?
</blockquote></details>

After talking offline, decided to go with the simpler API with math via constructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/80)*

PR #85, opened 2017-09-01
Enable single run to make sure exit

Lewuathe:
`singleRun` of karma is false as default. It causes stuck without exit when running `npm run test` in CLI. We can make sure to exit by enabling `singleRun`.

see: https://karma-runner.github.io/1.0/config/configuration-file.html

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/85)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -26,6 +26,7 @@ module.exports = function(config) {
       tsconfig: "tsconfig.json"
     },
     reporters: ['progress', 'karma-typescript'],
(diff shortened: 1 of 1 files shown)


nsthorat:
We don't use singleRun so that when we make a change to the code base it automatically recompiles and re-runs the test. This is super useful for quick development on unit tests and production code.

Lewuathe:
@nsthorat I got the purpose continuous testing without `singleRun`. Thanks.

PR #92, opened 2017-09-04
chore(yarn): add yarn.lock and support yarn

ivikash:
This will inherently support yarn.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/92)


dsmilkov:
Hi, is Yarn strictly better than Bower? If that's the case, can you remove bower from our scripts (package.json) and add yarn as a package dependency?

ivikash:
Hi @dsmilkov, I feel `yarn` is much more deterministic and handles nested dependencies easily with a lock file. I will also make the necessary changes in package.json. I am enclosing a few links for further read :-)
* https://bower.io/blog/2016/using-bower-with-yarn/
 * https://www.slant.co/versus/5094/17851/~bower_vs_yarn

ivikash:
@dsmilkov @nsthorat Any update on this?

CarmenPopoviciu:
if `bower` is completely removed as a dependency from the project (which is indeed a good move imo), shouldn't the bower dependencies (Polymer and all paper elements) be ported to `package.json` and have `bower.json` be removed as well? 

Perhaps it's better to have that tracked in a separate issue, so as to not mix concerns. I'd be more than happy to create an issue and submit a PR

Lewuathe:
Currently, we have seen a warning message when we run `npm run prep`.
```
npm WARN deprecated bower@1.8.2: ...psst! Your project can stop working at any moment because its dependencies can change. Prevent this by migrating to Yarn: https://bower.io/blog/2017/how-to-migrate-away-from-bower/
```

Basically, moving to yarn from bower is needed sooner or later, I think. 
It seems to be necessary to rebase the patch though.

nsthorat:
Thanks for putting this together. Because the change was a little deep in our build system we decided to take it on. We're now using yarn:
https://github.com/PAIR-code/deeplearnjs/commit/21dd65c756959def07c418e9c0c2604af1f71395

PR #110, opened 2017-09-08
Const and let are es6 not ts syntax

shancarter:
I was a little thrown by the original statement, as `const` and `let` are valid vanilla es6.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/110)


nsthorat:
:lgtm_strong:

Thank you, sorry this was confusing. That's right, let and const are valid :D too used to es5.

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/110#-:-KtYjRUUaJPV89Dsk03n:b1jcpc1)*

PR #115, opened 2017-09-10
Add benchmark for packed texture on GPU

Lewuathe:
Add missing benchmark of packed texture on GPU. It can show the speed that is significantly faster than unpacked texture.

screen shot 2017-09-10 at 23 09 02

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/115)

### DIFF
--- a/demos/benchmarks/math-benchmark-run-groups.ts
+++ b/demos/benchmarks/math-benchmark-run-groups.ts
@@ -26,7 +26,7 @@ import * as mulmat_gpu_benchmark from './mulmat_gpu_benchmark';
 export const BENCHMARK_RUN_GROUPS: BenchmarkRunGroup[] = [
   {
     name:
-        'Matrix Multiplication (CPU vs GPU): ' +
+        'Matrix Multiplication (CPU vs GPU vs GPU (packed texture)): ' +
             'matmul([size, size], [size, size])',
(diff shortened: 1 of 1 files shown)


nsthorat:
So this originally was in the benchmarks (which is why the actual file exists), but I removed it since we're going to implement this optimization in a different way. I don't want to get people's hopes up with the code as is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/115#-:-KtlcJOGD1J_kbV2DzMT:bc9hke0)*

Lewuathe:
I see. `MAX_POOL_POSNS_BENCHMARK_TEST` in max pool benchmark is not used either. Is this from same reason?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/115#-:-Ktow51e8GAdFJdkFvRi:ba7fajd)*

nsthorat:
Similar reason, this is just not useful to benchmark (it's only one part of the backprop).

nsthorat:
There are lots of other GPU methods we just haven't written benchmarks for which would be interesting to benchmark (unary ops, binary ops, resizing, batchnorm etc).

PR #127, opened 2017-09-14
Cache uniform locations in the GPGPUBinary so we don't query them every time we set an inputTextureMatrix

nsthorat:
This speeds things up in the model builder slightly, increase of 300 examples trained over 20 seconds on a macbook pro.

Setting arbitrary uniform values (floats, ints, etc) will come in a follow up PR. Right now the external demos just query uniform locations, but this will be centralized in the GPGPUBinary as well in the next PR which will move demos to logical sampling.

See this page for why caching uniform location is a good idea (TLDR it's slow):
https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/127)

### DIFF
--- a/demos/benchmarks/benchmark-demo.html
+++ b/demos/benchmarks/benchmark-demo.html
@@ -14,16 +14,17 @@ limitations under the License.
 ==============================================================================-->
 <!DOCTYPE html>
 <html>
+
 <head>
   <meta charset="UTF-8">
   <script src="../../bower_components/webcomponentsjs/webcomponents-lite.min.js"></script>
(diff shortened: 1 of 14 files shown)


dsmilkov:
:lgtm_strong:

---

Review status: 0 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/127#-:-KtyIA6mawuePYfboq_o:b-jbsf8z)*

PR #129, opened 2017-09-14
Vectorize (vec4) matmul and convolutions.

nsthorat:
When taking dot products for matmul / conv, break them into sum of vec4 dot products. We see large wins.

Matmul (before):
image
Matmul (after):
image

Conv (before):
image
Conv (after):
image

Max pool (before):
image

Max pool (after):
image

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/129)


dsmilkov:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[demos/benchmarks/avg_pool_cpu_benchmark.ts, line 20 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCJxfEWzNnWGfVNaOj:-KuCJxfEWzNnWGfVNaOk:brctpni) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_cpu_benchmark.ts#L20)):*
> mport * as conv_util from '../../src/math/conv_util';
> import {NDArrayMathCPU} from '../../src/math/math_cpu';
> import {Array3D} from '../../src/math/ndarray';

import public api whenever you can, here and elsewhere in this PR

---

*[demos/benchmarks/avg_pool_gpu_benchmark.ts, line 47 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCKSvFv8zpYYJ0X_x5:-KuCKSvFv8zpYYJ0X_x6:b-jjx6f0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_gpu_benchmark.ts#L47)):*
> gpgpu_math.runProgram(binary, [x], res);

to improve accuracy, how about calling gpgpu_math.runProgram once more, outside the for loop, before you start measuring time, as a warmup. here and in all other benchmarks. this will trigger upload of x and res to the gpu. Otherwise given 40 op runs, the first run might have a significant impact on the average (it might not, but worth a try).

---

*[demos/benchmarks/math-benchmark-run-groups.ts, line 62 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLOfcM735R008nx2E:-KuCLOfcM735R008nx2F:b-9isldl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/math-benchmark-run-groups.ts#L62)):*
> d1=16

seems like depth is 512 in maxpool_gpu_benchmark. can you also double-check the other numbers for the other benchmarks (important not to give users bad perf results :))

---

*[demos/benchmarks/max_pool_cpu_benchmark.ts, line 36 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLrq9Zgt82ebbyZLs:-KuCLrq9Zgt82ebbyZLt:b-fylabk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/max_pool_cpu_benchmark.ts#L36)):*
> 1

outputDepth 512 to match the GPU? and speaking of this, how about centralizing these benchmark params to one place, otherwise we get discrepancies between 
 3 places: cpu/gpu/run groups metadata - which is what users see

---

*[src/math/webgl/pool_gpu.ts, line 117 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCNPJrtQ-YfSbPDjiy:-KuCNPJrtQ-YfSbPDjiz:babgztl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L117)):*
> } else {

looks like you can early return on the if before, remove this else, and un-indent the block

---

*[src/math/webgl/pool_gpu.ts, line 145 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPh8JFdnn3nHiAaMW:-KuCPh8JFdnn3nHiAaMX:b-3z33j5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L145)):*
> int wR = f / ${filterWidth};

This can probably be optimized further, but I'm not 100% sure. Is there a way to conclude that `wR` will have the same value for `f`,`f+1`, `f+2`, and `f+3`. if yes, you can move it outside `getValue()` and avoid computing it 4 times.

---

*[src/math/webgl/pool_gpu.ts, line 151 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPBoNH-TH8I6LyoIs:-KuCPBoNH-TH8I6LyoIt:b-kwsec0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L151)):*
> if (xR < 0 || xR >= ${xNumRows} 

might improve perf: break this `if` in half, and move the part related to `xR`right after you compute `xR`. 

---

*[src/math/webgl/pool_gpu.ts, line 172 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRHYwSzHD12StEdJV:-KuCRHYx6fQpzlJ_a7xE:bu8ixde) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L172)):*
> ```XML
>           float avgValue = 0.0;
> 
>           for (int f = 0; f < ${filterNearestVec4}; f+=4) {
> ```

This might be too much to ask, but would be really good to understand the benefit of flattening out the weight indexing by benchmarking also the version where you keep two loops over `wR` and `wC` and convert to vec4 only the inner most loop.

If anything, this will help us understand how much room there is left to improve the flattened version, especially if the flat version is not faster than the un-flat version.

---

*[src/math/webgl/shader_compiler.ts, line 170 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRd5TX4GRY7IUWvcG:-KuCRd5UpnCYlBMFpnGj:bvudi3z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L170)):*
> return !equals.w || !equals.x || !equals.y || !equals.z;

a faster version might be: `return any(notEqual(values, values))`

---

*[src/math/webgl/shader_compiler.ts, line 173 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCS0gTiCd84Cbp5teO:-KuCS0gTiCd84Cbp5teP:bw93hi2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L173)):*
> float getNaN(vec4 values)

gosh this doesn't look great. if there was a way to produce NaN in glsl :) I researched this at one point and didn't find a solution. I tried again now and seems like a cleaner way would be to pass a NaN as uniform
`uniform float u_NaN` for every program.

other people claim success with `sqrt(-1.0);` but hard to tell without testing on all browsers.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129)*

nsthorat:
Review status: 5 of 22 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[demos/benchmarks/math-benchmark-run-groups.ts, line 62 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLOfcM735R008nx2E:-KuFoQwVT4RwSmBMAAxc:b-em322p) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/math-benchmark-run-groups.ts#L62)):*

> d1=16

seems like depth is 512 in maxpool_gpu_benchmark. can you also double-check the other numbers for the other benchmarks (important not to give users bad perf results :))
</blockquote></details>

Consolidated everything so param sharing is by design

---

*[demos/benchmarks/max_pool_cpu_benchmark.ts, line 36 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLrq9Zgt82ebbyZLs:-KuFoU6UlNFBVelR0VF4:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/max_pool_cpu_benchmark.ts#L36)):*

> 1

outputDepth 512 to match the GPU? and speaking of this, how about centralizing these benchmark params to one place, otherwise we get discrepancies between 
 3 places: cpu/gpu/run groups metadata - which is what users see
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 117 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCNPJrtQ-YfSbPDjiy:-KuFHjhXiFVhbBlJ0YmJ:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L117)):*

> } else {

looks like you can early return on the if before, remove this else, and un-indent the block
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 145 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPh8JFdnn3nHiAaMW:-KuFGF831X2MkqB7_JVp:b8pmp8s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L145)):*

> int wR = f / ${filterWidth};

This can probably be optimized further, but I'm not 100% sure. Is there a way to conclude that `wR` will have the same value for `f`,`f+1`, `f+2`, and `f+3`. if yes, you can move it outside `getValue()` and avoid computing it 4 times.
</blockquote></details>

Moved to just optimizing inner loop. That lets us do strided reads, so this is obsolute

---

*[src/math/webgl/pool_gpu.ts, line 151 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPBoNH-TH8I6LyoIs:-KuFob3dhuAXKFN3wxPl:b-5u5c5s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L151)):*

> if (xR < 0 || xR >= ${xNumRows} 

might improve perf: break this `if` in half, and move the part related to `xR`right after you compute `xR`. 
</blockquote></details>

I had done this originally, but it doesn't matter since all branches are taken. This just makes the code easier to read. Obsolete now.

---

*[src/math/webgl/pool_gpu.ts, line 172 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRHYwSzHD12StEdJV:-KuFogJ5kve6ZxZLYmkS:b-oqsj1a) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L172)):*

This might be too much to ask, but would be really good to understand the benefit of flattening out the weight indexing by benchmarking also the version where you keep two loops over `wR` and `wC` and convert to vec4 only the inner most loop.

If anything, this will help us understand how much room there is left to improve the flattened version, especially if the flat version is not faster than the un-flat version.
</blockquote></details>

It's pretty much the same in terms of performance, but I ended up doing 2 loops because we can optimize the read later.

---

*[src/math/webgl/shader_compiler.ts, line 170 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRd5TX4GRY7IUWvcG:-KuFokK8naXI-aPHGLua:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L170)):*

> return !equals.w || !equals.x || !equals.y || !equals.z;

a faster version might be: `return any(notEqual(values, values))`
</blockquote></details>

Done.

---

*[src/math/webgl/shader_compiler.ts, line 173 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCS0gTiCd84Cbp5teO:-KuFokwn3qKrXqQtSXYt:b5ghm4b) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L173)):*

> float getNaN(vec4 values)

gosh this doesn't look great. if there was a way to produce NaN in glsl :) I researched this at one point and didn't find a solution. I tried again now and seems like a cleaner way would be to pass a NaN as uniform
`uniform float u_NaN` for every program.

other people claim success with `sqrt(-1.0);` but hard to tell without testing on all browsers.
</blockquote></details>

Unfortunately we can't do this, for several reasons. First of all NaN is implementation specific, and second if you always upload a NaN, we  error out if the uniform isn't present (the chrome fragment shader compile transitively doesn't see a NaN in the program). This doesn't look great, but it's the best we can do (and vectorization is still much faster with this check).

---

*[demos/benchmarks/avg_pool_cpu_benchmark.ts, line 20 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCJxfEWzNnWGfVNaOj:-KuFIMwRyKyzZ91T89_F:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_cpu_benchmark.ts#L20)):*

> mport * as conv_util from '../../src/math/conv_util';
> import {NDArrayMathCPU} from '../../src/math/math_cpu';
> import {Array3D} from '../../src/math/ndarray';

import public api whenever you can, here and elsewhere in this PR
</blockquote></details>

Done.

---

*[demos/benchmarks/avg_pool_gpu_benchmark.ts, line 47 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCKSvFv8zpYYJ0X_x5:-KuFHulMgp5XlWF61p1c:b-e2b38) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_gpu_benchmark.ts#L47)):*

> gpgpu_math.runProgram(binary, [x], res);

to improve accuracy, how about calling gpgpu_math.runProgram once more, outside the for loop, before you start measuring time, as a warmup. here and in all other benchmarks. this will trigger upload of x and res to the gpu. Otherwise given 40 op runs, the first run might have a significant impact on the average (it might not, but worth a try).
</blockquote></details>

You can't do that - runProgram is non-blocking, if we leave the first one out, we're actually incorrectly timing ourselves (imagine CPU finishes in 10ms, but the GPU command takes 1 second, we would see a *slowdown*). Charles has always said upload is negligible, which is why he didn't do this in the first place (and the warmup is split here since compiling a shader program is blocking). I'm going to add disjoint query timers in a follow up CL, so I don't think it's worth optimizing this method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129)*

nsthorat:
Review status: 5 of 22 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[demos/benchmarks/math-benchmark-run-groups.ts, line 62 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLOfcM735R008nx2E:-KuGALgdc4B-u0-f7BFc:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/math-benchmark-run-groups.ts#L62)):*

Consolidated everything so param sharing is by design
</blockquote></details>

Done.

---

*[demos/benchmarks/max_pool_cpu_benchmark.ts, line 36 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCLrq9Zgt82ebbyZLs:-KuGAMCRyR7wU_ywCj7R:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/max_pool_cpu_benchmark.ts#L36)):*

Done.
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 117 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCNPJrtQ-YfSbPDjiy:-KuGAMjMS1_77QzY-Srz:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L117)):*

Done.
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 145 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPh8JFdnn3nHiAaMW:-KuGANIYCOqMj5bUcrKC:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L145)):*

Moved to just optimizing inner loop. That lets us do strided reads, so this is obsolute
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 151 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCPBoNH-TH8I6LyoIs:-KuGANnTY8HvhQpRCaCb:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L151)):*

I had done this originally, but it doesn't matter since all branches are taken. This just makes the code easier to read. Obsolete now.
</blockquote></details>

Done.

---

*[src/math/webgl/pool_gpu.ts, line 172 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRHYwSzHD12StEdJV:-KuGAOGZgOXlHaDxFxuk:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/pool_gpu.ts#L172)):*

It's pretty much the same in terms of performance, but I ended up doing 2 loops because we can optimize the read later.
</blockquote></details>

Done.

---

*[src/math/webgl/shader_compiler.ts, line 170 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCRd5TX4GRY7IUWvcG:-KuGAOljtXnKHOfuv5yJ:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L170)):*

Done.
</blockquote></details>

Done.

---

*[src/math/webgl/shader_compiler.ts, line 173 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCS0gTiCd84Cbp5teO:-KuGARjnHH9Nv6MInS_8:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/src/math/webgl/shader_compiler.ts#L173)):*

Unfortunately we can't do this, for several reasons. First of all NaN is implementation specific, and second if you always upload a NaN, we  error out if the uniform isn't present (the chrome fragment shader compile transitively doesn't see a NaN in the program). This doesn't look great, but it's the best we can do (and vectorization is still much faster with this check).
</blockquote></details>

Done.

---

*[demos/benchmarks/avg_pool_cpu_benchmark.ts, line 20 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCJxfEWzNnWGfVNaOj:-KuGAILXJi4rl-syN-3u:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_cpu_benchmark.ts#L20)):*

Done.
</blockquote></details>

Done.

---

*[demos/benchmarks/avg_pool_gpu_benchmark.ts, line 47 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-KuCKSvFv8zpYYJ0X_x5:-KuGAJ7Jlz_O5pkZ3X7S:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/236d1421278fcee5b690fd075faaa4f21d8706fa/demos/benchmarks/avg_pool_gpu_benchmark.ts#L47)):*

You can't do that - runProgram is non-blocking, if we leave the first one out, we're actually incorrectly timing ourselves (imagine CPU finishes in 10ms, but the GPU command takes 1 second, we would see a *slowdown*). Charles has always said upload is negligible, which is why he didn't do this in the first place (and the warmup is split here since compiling a shader program is blocking). I'm going to add disjoint query timers in a follow up CL, so I don't think it's worth optimizing this method.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129)*

dsmilkov:
:lgtm_strong:

---

Review status: 5 of 22 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/129#-:-KuGkQGVnMefbctHhsmy:b-jbsf8z)*

PR #133, opened 2017-09-16
Reorganize directory structure and move XhrDataset into src/data.

nsthorat:
This change was motivated by me needing XhrDataset for a new demo as part of the deeplearn top level exports.

This also renames "deeplearnjs.ts" alias module to "deeplearn.ts" - this is what the module name actually is.

In the process, I decided to reorganize the directory structure since we had lots of files just floating, and this makes it very clear what different classes are used for (optimizers, ops are all graph-related, for example). This new structure is as follows:

image

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/133)


dsmilkov:
:lgtm_strong: This is a great cleanup! Left just one question, which is not actionable, more like a discussion.

---

Reviewed 157 of 157 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/index.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/133#-KuCET4-paCjEFfzvdKb:-KuCET4-paCjEFfzvdKc:bqoqtmy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/9f095d5a91f45a0756411ee7d568bdf39b96dfd1/src/index.ts#L54)):*
> xhr_dataset

once we export this, we are making it part of public api. This is ok as long as we agree not to support backwards compatibility. We should plan at some point official dataset API. Also, since we export classes from xhr_dataset directly to top-level (line with export {XhrDataset.....}), why export xhr_dataset as a 2nd level? is it because of the getXhrDatasetConfig() method?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/133#-:-KuCG0XxsnZbtBwsSVnl:bt3h8ym)*

nsthorat:
Review status: 92 of 94 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/index.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/133#-KuCET4-paCjEFfzvdKb:-KuGD9I1hhZFmOY5TRok:bsbd1oe) ([raw file](https://github.com/pair-code/deeplearnjs/blob/9f095d5a91f45a0756411ee7d568bdf39b96dfd1/src/index.ts#L54)):*

> xhr_dataset

once we export this, we are making it part of public api. This is ok as long as we agree not to support backwards compatibility. We should plan at some point official dataset API. Also, since we export classes from xhr_dataset directly to top-level (line with export {XhrDataset.....}), why export xhr_dataset as a 2nd level? is it because of the getXhrDatasetConfig() method?
</blockquote></details>

Agreed about public API. I think with a version 1 release we'd have a centralized data format, and break backwards compatibility, and that's fine.

And yes, because of getXhrDatasetConfig()!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/133)*

PR #134, opened 2017-09-16
Clone dy when setting a gradient in as the same as gradient out.

nsthorat:
If we don't clone, the SummedGradientMap will dispose the old value, whose value may be read by another op.

FWIW this was so tricky to debug, even as a core author. An eager style execution will make this so much simpler.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/134)

### DIFF
--- a/src/ops/add.ts
+++ b/src/ops/add.ts
@@ -19,7 +19,7 @@ import {Tensor} from '../graph';
 import * as graph_util from '../graph_util';
 import {NDArrayMath} from '../math/math';
 import {NDArray, Scalar} from '../math/ndarray';
-import {TensorArrayMap, SummedTensorArrayMap} from '../tensor_array_map';
+import {SummedTensorArrayMap, TensorArrayMap} from '../tensor_array_map';
 import * as util from '../util';
 
 import {Operation} from './op';
@@ -75,7 +75,7 @@ export class Add extends Operation {
           gradientArrays.add(
               this.x1Tensor, math.divide(sum, this.dySizeScalar));
         } else {
-          gradientArrays.add(this.x1Tensor, dy);
+          gradientArrays.add(this.x1Tensor, math.clone(dy));
         }
       }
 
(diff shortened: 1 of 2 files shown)


dsmilkov:
:lgtm_strong:

---

---

*[src/ops/add.ts, line 78 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/134#-KuCVLcSAcFSSNEtCFXW:-KuCVLcSAcFSSNEtCFXX:b2di5ps) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e5996263b966f6ac6741760c71403f108b82404b/src/ops/add.ts#L78)):*
> math.clone(dy))

very curious: what situation led to this bug to manifest? Did it manifest in training and which demo? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/134#-:-KuCVsBG1CyOMpingI6A:b-jbsf8z)*

nsthorat:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/ops/add.ts, line 78 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/134#-KuCVLcSAcFSSNEtCFXW:-KuFESf57ZhY-ezD18LF:bkwwvh9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e5996263b966f6ac6741760c71403f108b82404b/src/ops/add.ts#L78)):*

> math.clone(dy))

very curious: what situation led to this bug to manifest? Did it manifest in training and which demo? 
</blockquote></details>

It happens in an unreleased demo (I sent you the name of it via chat) in an non-trivial wiring of tensors (after a matmul & bias add with a variable, then a loss function which takes a label in this case) but in the general case it's if two operations write to the same tensor (because of a split), however the original dy still needs to be read by yet another op who uses that for something.

I think it exposes something weird about the summed tensor array map
which assumes that it's a copy (when you add, and dispose old value)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/134)*

PR #153, opened 2017-09-25
Example: add a dataset to the model builder (Rune recognition) 

digitalheir:
[was asked](https://github.com/PAIR-code/deeplearnjs/issues/20#issuecomment-329345216) to write a tutorial for adding a data set to the model builder. hope you like it

I'm only a bit unsure about the example images I've prepared to the repo. They're about 650kb. I thought it might bloat the code repository a bit too much.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/153)


digitalheir:
I signed it!

nsthorat:
High level comments:

- This is so awesome. Thank you for putting this together. I think lots of people will benefit from this!
- It's fine to load the fonts externally
- Do we need to check the images in at all (hard for me to review, code review is freezing up because of so many files haha)? It seems like they're generated anyways, right? If that's the case, let's just leave them out. We can check the sprite image in, in fact we can host it on our CDN and include the runes in the public demo of the model builder.

Once the images are out of the repo I can review in a little more detail once reviewable.io doesn't freeze :D

---

Review status: 0 of 2018 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[demos/rune_recognition/generate_rune_data_for_model_builder.py, line 15 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/153#-Kv2x4z4ZLS_aSdBJaRR:-Kv2x4z5-SLvphENCpfw:b-djli6o) ([raw file](https://github.com/pair-code/deeplearnjs/blob/a13946f3a991a09eb18e6c2d09a88ac842691b47/demos/rune_recognition/generate_rune_data_for_model_builder.py#L15)):*
> ```Python
> import json
> 
> path_to_images = "runes/"
> ```

use single quotes for strings everywhere

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/153#-:-Kv2xCJHV-DiYR3XEYCd:bldq9we)*

PR #161, opened 2017-09-30
Add and synchronize typings

arthurjdam:
Makes applications leveraging TSHint (such as WebStorm) a bit happier :-)

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/161)

### DIFF
--- a/src/math/math_cpu.ts
+++ b/src/math/math_cpu.ts
@@ -233,7 +233,7 @@ export class NDArrayMathCPU extends NDArrayMath {
   }
 
   protected scaledArrayAddInternal<T extends NDArray>(
-      c1: Scalar, a: T, c2: Scalar, b: T) {
+      c1: Scalar, a: T, c2: Scalar, b: T): T {
     const newShape = util.assertAndGetBroadcastedShape(a.shape, b.shape);
     const newValues = new Float32Array(util.sizeFromShape(newShape));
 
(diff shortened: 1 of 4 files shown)


nsthorat:
:lgtm_strong:

Thanks for this PR!

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/161#-:-KvJYOfIWvPz5WweHKWH:b-3l5y2q)*

dsmilkov:
Hi Arthur,

What version of tslint and typescript compiler is the WebStorm IDE using? The typings you added a not necessary, since the typescript `2.4` compiler can infer them.

It would be nice if you figure out the typing inference in your setup, in order to stay in sync and not have to constantly update the code as the library progresses

arthurjdam:
Hi @dsmilkov ,

This'll resolve typings for even the latest version of WebStorm (`2017.2.4`), which uses TypeScript `2.4.2`. Though it doesn't produce a critical compiler error, it does require a user to disable autocompletion in order to not get into a fight with tsc.

dsmilkov:
Thanks for clarifying. Hope WebStorm resolves this on their side soon. We use Visual Studio Code and the auto-completion works well without the explicit typings. In the meantime, feel free to send us typings as you get out of sync with master.

Cheers

arthurjdam:
We should be set for now. Indeed a little strange that WS doesn't infer the types properly while VS does. Either way; I'll keep an eye on this as you guys progress! Thanks!

PR #194, opened 2017-10-08
Small optimization in sample snippets

reiinakano:
I noticed there was an opportunity to use a single dot product to perform a few consecutive multiplications and additions. I'm not sure if this was intentional so please close if there's a reason you didn't do it this way in the first place. I'm a bit doubtful about the efficiency of conversions between int and float but https://groups.google.com/forum/#!topic/comp.graphics.api.opengl/-FMnZy-yfxI seems to say it's fine.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/194)


reiinakano:
signed it

src/math/webgl/shader_compiler.ts
```diff
@@ -130,7 +130,9 @@ vec2 UVfrom2D(int texNumR, int texNumC, int numC, int row, int col) {
 const SAMPLE_3D_SNIPPET = `
 vec2 UVfrom3D(int texNumR, int texNumC, int stride0,
     int stride1, int row, int col, int depth) {
-  int index = row * stride0 + col * stride1 + depth;
+  vec3 v1 = vec3(row, col, depth);
```
dsmilkov: Unfortunately, dot() works only on floats, and for large `ndarray`s, the flat index can be high (>10 million), at which point float is loosing precision and causes index to be `10,000,001` instead of `10,000,000` for example. We had a real model affected by this.

reiinakano:
Ahh, I see. I've reverted the changes and left a single line comment. Let me know if you want me to reword it.

dsmilkov:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/194#-:-Kw2YMJRkQhCwYmh4Das:b-jbsf8z)*

dsmilkov:
All looks good. The branch is out of sync - once you merge with master, let me know and I'll push

reiinakano:
Done!

PR #201, opened 2017-10-10
Minor nitpick in model-builder.ts learingRate typo

reiinakano:
change `learingRate` to `learningRate`. This never caused problems before since the optimizer is always recreated with `learningRate` when Train is clicked. In fact, I may be wrong but I don't believe there is a need for the initial optimizer at all.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/201)


nsthorat:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/201#-:-Kw6CUzPEreNCixoJ-VM:b-jbsf8z)*

PR #202, opened 2017-10-10
Add NDArray.fromPixels, and use it in the SqueezeNet demo.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/202)


dsmilkov:
Reviewed 3 of 9 files at r1.
Review status: 3 of 9 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[demos/models/squeezenet.ts, line 113 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBqzRCRSFJxDFwWS84:-KwBqzRCRSFJxDFwWS85:b3wgnd9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/5faba47f2b3da9644ce2a20ffaf6e052ed4fedcc/demos/models/squeezenet.ts#L113)):*
> ```XML
>           [fire3.shape[0] - 1, (fire3.shape[1] - 1) * fire3.shape[2]]);
>       const fire3Sliced = fire3Sliced2d.as3D(
>           fire3.shape[0] - 1, fire3.shape[1] - 1, fire3.shape[2]);
> ```

nice! you moved to 'valid' padding

---

*[src/math/ndarray.ts, line 153 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBrLe7LakzSvV7x4qP:-KwBrLe7LakzSvV7x4qQ:b-4c96tm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/ndarray.ts#L153)):*
> ImageData|HTMLImageElement|HTMLCanvasElement

I think these might have width and height already, so maybe the user doesn't need to pass in `shape` at all? maybe just a bit that says `isGrayscale` ?

---

*[src/math/ndarray.ts, line 288 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBrnQijyUHGZ6CGs7k:-KwBrnQijyUHGZ6CGs7l:bpu6d62) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/ndarray.ts#L288)):*
> this.data.textureType = TextureType.DEFAULT;

`NDArray`s can be uploaded and downloaded from gpu many times. Shouldn't uploadToGPU() use the existing `this.data.textureType` instead of overwriting it, and calling different `uploadMatrixToTexture` implementations depending on `textureType`

---

*[src/math/webgl/shader_compiler.ts, line 533 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBtZiYNmJrtC04GEyX:-KwBtZiYNmJrtC04GEyY:bayfy5l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/webgl/shader_compiler.ts#L533)):*
> else if (inputInfo.shapeInfo.textureType === TextureType.RGBA_COLOR) {
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>
>       return `
>         float ${funcName}(int row, int col, int depth) {
>           vec2 uv = (vec2(col, row) + halfCR) / vec2(${tC}.0, ${tR}.0);
>           return sampleUVAndDepth(${texName}, uv, depth);
>         }
>       `;
</details>

I wouldn't change `getSampler3D`. Instead I would introduce `float sample${texName}Color()` so it's clear that it is sampling a color. I would generate the code for that method only if one of the inputs has textureType that is color

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202)*

nsthorat:
Review status: 3 of 9 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/ndarray.ts, line 153 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBrLe7LakzSvV7x4qP:-KwBzoqcs3zKfcnC59-L:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/ndarray.ts#L153)):*

> ImageData|HTMLImageElement|HTMLCanvasElement

I think these might have width and height already, so maybe the user doesn't need to pass in `shape` at all? maybe just a bit that says `isGrayscale` ?
</blockquote></details>

Done.

---

*[src/math/ndarray.ts, line 288 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBrnQijyUHGZ6CGs7k:-KwBzpbDw4Py0VNuVbqS:b7zqo3z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/ndarray.ts#L288)):*

> this.data.textureType = TextureType.DEFAULT;

`NDArray`s can be uploaded and downloaded from gpu many times. Shouldn't uploadToGPU() use the existing `this.data.textureType` instead of overwriting it, and calling different `uploadMatrixToTexture` implementations depending on `textureType`
</blockquote></details>

We always want this to be  DEFAULT when uploading, we dont upload packed textures, but this reminds me I need to get download to work.

---

*[src/math/webgl/shader_compiler.ts, line 533 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202#-KwBtZiYNmJrtC04GEyX:-KwBztma27pQ88fWeTRD:bfkvwf3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/54972f147d7720349b4de73980b3124b9aa2089f/src/math/webgl/shader_compiler.ts#L533)):*

> else if (inputInfo.shapeInfo.textureType === TextureType.RGBA_COLOR) {
>       return `
>         float ${funcName}(int row, int col, int depth) {
>           vec2 uv = (vec2(col, row) + halfCR) / vec2(${tC}.0, ${tR}.0);
>           return sampleUVAndDepth(${texName}, uv, depth);
>         }
>       `;

I wouldn't change `getSampler3D`. Instead I would introduce `float sample${texName}Color()` so it's clear that it is sampling a color. I would generate the code for that method only if one of the inputs has textureType that is color
</blockquote></details>

We want to override the existing sample method because want all existing shaders to work with this packed RGBA texture input to work out of the box.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/202)*

PR #203, opened 2017-10-11
Add math.ceil math.floor

reiinakano:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/203)


nsthorat:
Couple comments, just make sure tests pass! For some reason the clip NaN propagation was failing at head yesterday.

Thanks again for the contribution!! :) 

---

Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/unaryop_test.ts, line 275 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-KwG3vB-3z6waUpE_jDK:-KwG3vB-3z6waUpE_jDL:b18xdx1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/04304b2f7cf76bde8b4231a25672490bb4fc0c4f/src/math/unaryop_test.ts#L275)):*
> ```XML
> {
>   const tests: MathTests = it => {
>     it('exp', math => {
> ```

change the it from exp - maybe just 'basic'

---

*[src/math/unaryop_test.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-KwG3zLKb8fyxzRZNOaX:-KwG3zLKb8fyxzRZNOaY:bei5ud1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/04304b2f7cf76bde8b4231a25672490bb4fc0c4f/src/math/unaryop_test.ts#L287)):*
> ```XML
>     });
> 
>     it('exp propagates NaNs', math => {
> ```

change the it from exp - maybe just 'propagates NaNs'

---

*[src/math/unaryop_test.ts, line 309 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-KwG3yx4z5YFkB3LkS1r:-KwG3yx4z5YFkB3LkS1s:b18xdx1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/04304b2f7cf76bde8b4231a25672490bb4fc0c4f/src/math/unaryop_test.ts#L309)):*
> ```XML
> {
>   const tests: MathTests = it => {
>     it('exp', math => {
> ```

change the it from exp - maybe just 'basic'

---

*[src/math/unaryop_test.ts, line 321 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-KwG40nWL4sZVPw82XIk:-KwG40nWL4sZVPw82XIl:bei5ud1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/04304b2f7cf76bde8b4231a25672490bb4fc0c4f/src/math/unaryop_test.ts#L321)):*
> ```XML
>     });
> 
>     it('exp propagates NaNs', math => {
> ```

change the it from exp - maybe just 'propagates NaNs'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-:-KwG45b-dKBK2diH4xIu:b-laio4u)*

reiinakano:
Hmm, my bad about the tests. The thing is, on my laptop, even with a fresh install of master, 280+ tests fail. I have to backtrack and figure out when adding tests that I don't increase the number of failures.

To add to the complexity, I noticed that all tests running this setup `{'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 2}` fail on my machine. So I have to account for that as well. I believe I got a bit careless last time counting things. Sorry about that! I made sure this one is fine!

I also wrongly assumed that when Travis passes all is good, but looking closer, it looks like it only checks for successful compilation. Not sure what the options are (or if there are any), but you should look into adding CI that actually runs the tests. It'd be super difficult to verify PRs from strangers like me without them! ;)

I'd appreciate it if you could guide me in having all the tests pass on my machine. I definitely won't be making the same mistake again. Sorry!

nsthorat:
Ahh no worries at all, this is totally our fault. I think you're on a machine without WebGL 2.0, which is probably the issue. Can you check your WebGL version: http://webglreport.com/

We're going to add CI that actually runs tests soon, unfortunately Travis isn't good enough because they don't run GPU tests. We're going to set up browser stack at some point in the near future so that we actually get coverage, so this testing stuff is totally our fault, so sorry about that!

nsthorat:
:lgtm_strong:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/203#-:-KwGIHAzKK6LizLA7Qgu:b-jbsf8z)*

reiinakano:
You're right, both my broswers are on WebGL 1. No worries!

nsthorat:
Just curious what OS / machines you use to develop, could you post your WebGL reports here?

reiinakano:
Chrome
webglchrome

Mozilla
webglmozilla

nsthorat:
Thanks! And presumably your WebGL 2 tab doesn't show anything?

reiinakano:
Just this snippet! 

`× This browser does not support WebGL 2`

I tried turning it on it Firefox using this: https://wiki.mozilla.org/Platform/GFX/WebGL2, but no difference in tests failing.

For Chrome, same result. But I'll try updating it and see if that works.

nsthorat:
Cool, no problem, it may be your hardware.

What kind of performance do you get for 1024x1024 matmul on the benchmarks page, out of curiosity? What kind of graphics card do you have?

reiinakano:
512.455ms

I believe I'm just using the Intel one bundled with CPU. My laptop has a
proper GPU, but somehow I never actually got around to installing the
NVIDIA drivers since I do most neural network stuff on the cloud. Maybe
things'll run faster once I properly set that up.

nsthorat:
Cool, yeah that may help, but not totally sure. 500ms is pretty slow for the 1024x1024, I get ~120ms on my macbook pro with an integrated graphics card.

reiinakano:
I quickly switched over to my Windows dual boot and got 211ms on the
benchmark. It's also WebGL 2 compatible. It probably is hardware.

Issue #223, opened 2017-10-17
Add common missing ops.


nsthorat:
Flatten (covered by reshape, but would be a nice sugar method)
EmbeddingLookup
BatchNormalization2D
Conv1D, Conv3D
MaxPooling1D, MaxPooling2D (MaxPooling3D is already available)
Permute
Repeat
LeakyRELU, PRelu, ELU
Dropout (we could punt on this for now)

dnnsthnnr:
I would also add axis aligned Softmax in multidimensional Arrays to the list.

nsthorat:
We just committed the axis-aligned softmax:
https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/math.ts#L782

tycho01:
I'd found some node types listed on this [Neural Network Zoo](http://www.asimovinstitute.org/neural-network-zoo/) blog post of which I was wondering whether they're already covered among the current [graph methods](https://deeplearnjs.org/docs/api/classes/graph.html) as well.
I'm not sure all will correspond well since they're different layers of abstractions, but here's the impression I got comparing the two:
- [x] Input cell
- [ ] Back-fed input cell, listed as used as input of Hopfield Networks, (Restricted) Boltzmann Machine, Deep Belief Networks.
- [ ] Noisy input cell: listed as used as input of Denoising Auto-Encoders.
- [x] Hidden cell = Fully connected layer
- [ ] Probabilistic hidden cell: used in Markov Chains, (Restricted) Boltzmann Machine, Deep Belief Networks, Deep Convolutional Inverse Graphics Networks.
- [ ] Spiking hidden cell: used as the hidden cells of Liquid State Machines.
- [x] Output cell
- [ ] Match input-output cell: used in Auto Encoders, Deep Belief Networks, GANs.
- [x] Recurrent cell = ReLU
- [ ] Memory cell: used in LSTMs, Neural Turing Machines
- [ ] Different memory cell: used in Gated Recurrent Units
- [ ] Kernel cell: used in conv nets.
- [x] Convolution

I'm probably wrong about half of these as I know [conv nets](https://deeplearnjs.org/demos/model-builder/) and [GANs](https://reiinakano.github.io/gan-playground) were already being made with this, but maybe some others are still missing, if maybe not as common.

dsmilkov:
Hi,

[...]

PR #229, opened 2017-10-18
Add math.flatten

dnnsthnnr:
Adds sugar method flatten #223

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/229)


nsthorat:
Thanks for the contribution! Just one comment

---

---

*[src/math/math.ts, line 378 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/229#-Kwkw7I4gC0Te2-XUMDS:-Kwkw7I4gC0Te2-XUMDT:b-hrcny2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/43bde577e74801d7fee3fff91870b014ef96dd8b/src/math/math.ts#L378)):*
> ```XML
>    * @returns {Array1D}
>    */
>   flatten<T1 extends NDArray>(ndarray: T1): Array1D {
> ```

this should actually live on NDArray, where reshape lives as well, then the test should be in ndarray_test

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/229#-:-KwkwAFSmOXbgqpShLmy:b-5zyvns)*

dnnsthnnr:
Wasn't sure, but makes sense. Will change it.

dnnsthnnr:
@nsthorat while changing i found that flatten is exactly the same as the existing method NDArray.as1D().
This makes a flatten method redundant, or is it all about the naming?

nsthorat:
Ahh, looks like you are right.

Adding another method `flatten` to `NDArray` which just calls as1D() is all we need! Sorry about that. If you add that method I'll merge it in.

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/229#-:-Kwp-d1Ib6KVJVvtIGm0:b-jbsf8z)*

PR #231, opened 2017-10-19
add math.elu, math.leakyRelu

dnnsthnnr:
i got an elu implementation atm for the cpu only.
My question regarding the gpu implmentation is: would it be the best to implement it with the BinaryProgram?

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/231)


nsthorat:
Thanks for the contribution, few minor comments!

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/element_wise_arithmetic_test.ts, line 568 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-Kwp-o3xXH-_UjbJxRX_:-Kwp-o3xXH-_UjbJxRXa:b-kmbhup) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8936fe9961a65a25af58ab125c5f6911eda4f27c/src/math/element_wise_arithmetic_test.ts#L568)):*
> ```XML
> // math.elu
> {
>   const tests: MathTests = it => {
> ```

this should be in unary_op with leaky relu

---

*[src/math/math.ts, line 1107 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-Kwp5mKTBqUMPC2AePdI:-Kwp5mKTBqUMPC2AePdJ:bdwyu9u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8936fe9961a65a25af58ab125c5f6911eda4f27c/src/math/math.ts#L1107)):*
> ```XML
>    */
>   leakyRelu<T extends NDArray>(ndarray: T): T {
>     return this.executeOp('leaky-relu', () => this.leakyReluInternal(ndarray));
> ```

instead of dash, camel case: `leakyRelu`

---

*[src/math/math_cpu.ts, line 549 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-Kwp5j-GpL9omKR5YCiN:-Kwp5j-GpL9omKR5YCiO:be47fsy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8936fe9961a65a25af58ab125c5f6911eda4f27c/src/math/math_cpu.ts#L549)):*
> ```XML
>   }
> 
>   protected eluInternal<T extends NDArray>(ndarray: T, scale: Scalar): T {
> ```

TF doesn't use a scale:

https://www.tensorflow.org/versions/master/api_docs/python/tf/nn/elu

---

*[src/math/math_cpu.ts, line 572 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-Kwp07ONMxOtVUjiB1Ow:-Kwp07ONMxOtVUjiB1Ox:b4mhekt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8936fe9961a65a25af58ab125c5f6911eda4f27c/src/math/math_cpu.ts#L572)):*
> ```XML
>         resultValues[i] = v;
>       } else {
>         resultValues[i] = 0.01 * v;
> ```

this 0.01 should be passed in as `alpha`

https://www.tensorflow.org/versions/master/api_docs/python/tf/nn/leaky_relu

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-:-Kwp5pxoxBxeK94NKNlg:b-rqwgle)*

dnnsthnnr:
Will do :) Was looking for the definitions of the activation functions described in the papers, where there is an alpha in elu and the factor in leaky relu is fixed to 0.01. Although no problem. In the future, I will look up the tensorflow definitions.
If the leaky relu gets an alpha parameter, then whats the difference to pRelu?

Im new to shaders, so i dont know exactly how to implement the alpha for the leaky relu. would it be the best to do use a binary program or should I use a flexible unary program?

nsthorat:
One small comment about alpha, sorry to be pedantic. But this is so great, thank you!

---

Reviewed 6 of 6 files at r2.

---

*[src/math/math.ts, line 1104 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-KwpnIcVdmwe4cKy9KaI:-KwpnIcVdmwe4cKy9KaJ:bj07kg9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f08209803804e516f9c1fba6c21fa22023f58044/src/math/math.ts#L1104)):*
> ```XML
>    * @return {NDArray}
>    */
>   leakyRelu<T extends NDArray>(ndarray: T, alpha?:Scalar): T {
> ```

alpha will become a `number` after the other comment, so you can just initialize it here in the params as alpha = 0.2.

---

*[src/math/math_gpu.ts, line 331 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-Kwpmy0heaDNuBTNrVtL:-Kwpmy0heaDNuBTNrVtM:ber72cv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f08209803804e516f9c1fba6c21fa22023f58044/src/math/math_gpu.ts#L331)):*
> ```XML
>   }
> 
>   protected leakyReluInternal<T extends NDArray>(a: T, alpha: Scalar): T {
> ```

small comment: make alpha a regular `number` and not a `Scalar`. by calling `.get()` and splicing it in, we're introducing a CPU / GPU interlock. We don't need alpha to be a real `NDArray` here, it's fine to just leave it as a regular javascript number. The program you the construct here will have the alpha number spliced in directly, instead of uploaded as a texture, since that's also what TF does.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-:-KwpnfP9_DMwIxsFhd2S:bjf7qh9)*

dnnsthnnr:
Absolutely no problem. Made the changes.

nsthorat:
:lgtm_strong:

---

Reviewed 2 of 3 files at r3.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/231#-:-Kx04y5-_qMAP4Z6Hji2:b-jbsf8z)*

nsthorat:
BTW, thanks for this awesome contribution!

PR #248, opened 2017-10-20
Add axes to argMax/argMin/argMaxEquals. Add math.equal and batching to math.multinomial. Optimize compiler

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/248)


nsthorat:
:lgtm_strong:

Wonderful change, just a few comments

---

Reviewed 4 of 16 files at r1, 6 of 11 files at r2, 12 of 12 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/math/broadcast_util.ts, line 25 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx07ZJ2onxIAv614y5k:-Kx07ZJ2onxIAv614y5l:ba38u85) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/broadcast_util.ts#L25)):*
> ```XML
>  * inShape = [4, 1, 3]
>  * outShape = [5, 4, 3, 3]
>  * result = [1]. Dimension 1 (2nd dimension of input) gets broadcasted 1 => 4.
> ```

isn't this actually 1 => 3?

---

*[src/math/element_wise_arithmetic_test.ts, line 573 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx080mG-4wkPxyCeE7c:-Kx080mG-4wkPxyCeE7d:btzhn1g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/element_wise_arithmetic_test.ts#L573)):*
> ```XML
>       const b = Array1D.new([4, 5, -1]);
> 
>       const res = math.equal(a, b).getValues();
> ```

You should also assert the DType of "res" here and below

---

*[src/math/math.ts, line 271 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx08KWMGRG9cwHOLAxz:-Kx08KWMGRG9cwHOLAy-:bch6bqd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math.ts#L271)):*
> ```XML
>           `%c${paddedName}\t%c${time}\t%c${rank}D ${shape}\t%c${size}`,
>           'font-weight:bold', 'color:red', 'color:blue', 'color: orange');
>       this.checkForNaN(result.getValues(), result.dtype, name);
> ```

why did you move this? Now the performance.now() time will be wrong since it happens before you call getValues.

BTW - tangentially, I thought you had a PR where the timing code for debug mode to uses the query timer?

---

*[src/math/math.ts, line 1764 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx08nj2W0Y8cWbXS8Ih:-Kx08nj2W0Y8cWbXS8Ii:b-esj0eu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math.ts#L1764)):*
> ```XML
>    */
>   multinomial(
>       probabilities: Array1D|Array2D, numSamples: number,
> ```

why not make this a generic so the compiler knows statically what the ins / outs are?

---

*[src/math/math_cpu.ts, line 438 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx096-P-1gvU_UNR2_P:-Kx096-P-1gvU_UNR2_Q:b4sb3se) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math_cpu.ts#L438)):*
> ```XML
>       const loc = result.indexToLoc(i);
> 
>       const aLoc = loc.slice(-a.rank);
> ```

we talked about moving some of this logic out to common functionality, it may be worth doing that if we expect this to be a blueprint

---

*[src/math/ndarray.ts, line 845 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0A9o81QZwCO_0b6WD:-Kx0A9o81QZwCO_0b6WE:b-og50ep) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/ndarray.ts#L845)):*
> ```XML
>       if (util.isValNaN(val as number, 'bool')) {
>         bool[i] = util.getNaN('bool');
>       } else if (array[i]) {
> ```

else if (val)

---

*[src/math/reduction_ops_test.ts, line 141 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0AMfzUUQ0qlZiGz4z:-Kx0AMfzUUQ0qlZiGz5-:bny0ip2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/reduction_ops_test.ts#L141)):*
> ```XML
>     it('Array1D', math => {
>       const a = Array1D.new([1, 0, 3, 2]);
>       const result = math.argMax(a);
> ```

test the dtype here and below too

---

*[src/math/unaryop_test.ts, line 68 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0AT2DbD2yw9Rvar5L:-Kx0AT2DbD2yw9Rvar5M:bhk380o) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/unaryop_test.ts#L68)):*
> ```XML
>     });
> 
>     it('propagates NaNs, float32', math => {
> ```

You should be able to assert dtypes of the response here and below as well, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-:-Kx0BT1RMnT41x-50kRM:bdm0b8g)*

dsmilkov:
Reviewed 6 of 16 files at r1.
Review status: 15 of 22 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/math/broadcast_util.ts, line 25 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx07ZJ2onxIAv614y5k:-Kx1460DdO60eO4botSm:b1pzwm5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/broadcast_util.ts#L25)):*

isn't this actually 1 => 3?
</blockquote></details>

nice catch. done.

---

*[src/math/element_wise_arithmetic_test.ts, line 573 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx080mG-4wkPxyCeE7c:-Kx14QGidpNrUh1KJ55x:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/element_wise_arithmetic_test.ts#L573)):*

You should also assert the DType of "res" here and below
</blockquote></details>

Done.

---

*[src/math/math.ts, line 271 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx08KWMGRG9cwHOLAxz:-Kx14fDrIlJneIVnl4me:b-prbs9l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math.ts#L271)):*

why did you move this? Now the performance.now() time will be wrong since it happens before you call getValues.

BTW - tangentially, I thought you had a PR where the timing code for debug mode to uses the query timer?
</blockquote></details>

ups. great catch! I'm not sure why I moved it down (maybe while debugging something else).

Yes, I had a hacky PR where the timing code for debug uses query timer, but never cleaned the code and sent for review. In future PR we should do that.

---

*[src/math/math.ts, line 1764 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx08nj2W0Y8cWbXS8Ih:-Kx15EnMrZgApg0Orp7d:b85oljs) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math.ts#L1764)):*

why not make this a generic so the compiler knows statically what the ins / outs are?
</blockquote></details>

I tried all sorts of things, but failed to type it. The basic problem is that we can't yet strongly type both the rank and the dtype until we introduce an additional generic property in NDArray corresponding to the rank - which we will in future PR.

For example right now we can't type

`multinomial<T extends Array1D|Array2D>(...): T<'int32'> {...`

problem is `T<'int32'>`. But later when we add rank as generic to NDArray we can type:
`multinomial<NDArray<D,R>>(...): NDArray<'int32',R> {...` where `R` is the rank.

---

*[src/math/math_cpu.ts, line 438 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx096-P-1gvU_UNR2_P:-Kx1BmeTzn2uQrOCDjXK:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/math_cpu.ts#L438)):*

we talked about moving some of this logic out to common functionality, it may be worth doing that if we expect this to be a blueprint
</blockquote></details>

Done.

---

*[src/math/ndarray.ts, line 845 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0A9o81QZwCO_0b6WD:-Kx16cfZ5qm7O5ZYi4G4:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/ndarray.ts#L845)):*

else if (val)
</blockquote></details>

Done.

---

*[src/math/reduction_ops_test.ts, line 141 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0AMfzUUQ0qlZiGz4z:-Kx18IRkTcqcgzxA3Qgo:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/reduction_ops_test.ts#L141)):*

test the dtype here and below too
</blockquote></details>

Done.

---

*[src/math/unaryop_test.ts, line 68 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248#-Kx0AT2DbD2yw9Rvar5L:-Kx17Zt2zDut3eJ8p3L1:b-bfw56l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/3c08504fd96a417633b430f60fc5ed2fb99b444e/src/math/unaryop_test.ts#L68)):*

You should be able to assert dtypes of the response here and below as well, right?
</blockquote></details>

Done. Now I'm explicitly testing the dtype. Before I was implicitly testing the dtype since `expectArraysClose(res.getValues(), new Int32Array([...]))` will fail if `getValues()` doesn't return an `Int32Array`. Likewise for UInt8Array and Float32Array.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/248)*

PR #249, opened 2017-10-20
Add minpool benchmark

menglewis:
I added minPool to the benchmark demo.
I also noticed that the avgPool CPU benchmark was actually calling `maxPool` instead of `avgPool` so I added a method that will call the correct pool method based on the type param.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/249)

### DIFF
--- a/demos/benchmarks/math-benchmark-run-groups.ts
+++ b/demos/benchmarks/math-benchmark-run-groups.ts
@@ -71,49 +71,21 @@ export function getRunGroups(): BenchmarkRunGroup[] {
     params: convTransposedParams
   });
 
-  const maxPoolParams:
+  const poolParams:
       PoolBenchmarkParams = {depth: 8, fieldSize: 4, stride: 4, type: 'max'};
   groups.push({
-    name: 'Max pool',
+    name: 'Pool Op Benchmark: input [size, size]',
     min: 0,
     max: 1024,
     stepSize: 64,
     stepToSizeTransformation: (step: number) => Math.max(4, step),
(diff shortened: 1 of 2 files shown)


nsthorat:

---

*[demos/benchmarks/math-benchmark-run-groups.ts, line 92 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/249#-Kx0Bq9-1Jflv9XwldeP:-Kx0Bq9-1Jflv9XwldeQ:byhlcv3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca7a3bc1200b4dc55f0261f6e309b463fa420029/demos/benchmarks/math-benchmark-run-groups.ts#L92)):*
> ```XML
>       PoolBenchmarkParams = {depth: 8, fieldSize: 4, stride: 4, type: 'min'};
>   groups.push({
>     name: 'Min pool',
> ```

what do you think about all of the pooling ops being part of a dropdown, just like we do for reduction / unary?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/249)*

menglewis:
I think that makes sense. I'll update it.

nsthorat:
:lgtm_strong:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/249#-:-KxQj9wWAWELKTKXnd0_:b-jbsf8z)*

PR #262, opened 2017-10-24
Add CHANGELOG

glortho:
This adds a simple changelog according to GitHub convention to help users track what's happening in the project. I started this because I was trying to find out what changed in the 0.3.3 release you published to npm yesterday and was a bit lost without a changelog or tags (which would also be nice to add :-).

I ported over the information from your releases for now.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/262)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,12 +59,10 @@ of the library from [unpkg](https://unpkg.com):
 To use a specific version, add `@version` to the unpkg URL above
 (e.g. `https://unpkg.com/deeplearn@0.2.0`), which you can find in the
 [releases](https://github.com/PAIR-code/deeplearnjs/releases) page on GitHub.
-After importing the library, the API will be available as `deeplearn` in the
-global namespace. We recommend declaring the `dl` alias at the top of the file:
+After importing the library, the API will be available as `dl` in the global
+namespace.
 
 ```js
-var dl = deeplearn;
-
 var math = new dl.NDArrayMathGPU();
 var a = dl.Array1D.new([1, 2, 3]);
 var b = dl.Scalar.new(2);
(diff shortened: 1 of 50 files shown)


glortho:
I signed it!

nsthorat:
Thanks for taking the time to do this!

Would you be up for making an automated way to do this?

When we update the version, we run the `./scripts/make-version` script, which updates the version in our `version.ts` file.

What we could do is also automatically update the changelog when we run this version by querying github for the latest change between now and the last version. The determine the last commit between now and the last version, we would keep another file  with the commit ID.

We will from here on make sure that contributors use titles in PRs that are descriptive.

Let me know if you're up for this!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/262#-:-KxTzNoHpPZqzh9yZRLM:b602gng)*

glortho:
I’d be happy to do that, though if we’re going to automate I think it’s pretty critical that we use something like commitizen: https://github.com/commitizen/cz-cli/blob/master/README.md

Otherwise you end up with changelogs that are quite difficult to parse, even with good commit msgs. For example, with a major version bump will it be clear which commits relate to the breaking changes and which do not? commitizen makes this much easier to enforce. Let me know what you think.
________________________________

nsthorat:
HI @glortho,

Daniel and I took a look at commitizen and we're worried about adding another dependencies as well as trying to enforce the use of `git cz`.

How about we do this: keep a file storing the last commit ID of the last patch release, and when we run make-version we simply log to the console the descriptions + IDs of the PRs between now and the last release. We'll then manually put them into the releases tab as we publish the NPM.

How does that sound?

glortho:
Sure, I understand what you mean about dependencies/workflow changes.

If we `git tag` every >= patch release, we should be able to avoid storing
ids. We could just `git log` for pr merge commits since the last tag. I'll
take a stab at it and see what you think.

glortho:
Actually one question about your flow @nsthorat -- do you `git tag` before you run make-version or after?

dsmilkov:
We don't `git tag` programatically. We just manually release a new version via the [releases](https://github.com/PAIR-code/deeplearnjs/releases) page, which tags for us, but I like the idea of `git tag`ing all patch versions via the `make-version` script

dsmilkov:
Hi, first, thank you for your contribution and apologies for the late response.

The team has decided to use git tags instead of change log and have a fast release cycle using GitHub release notes to keep our users up to date.

PR #261, opened 2017-10-23
Add tests to Travis CI via BrowserStack

dsmilkov:
Fixes #8 

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/261)

PR #260, opened 2017-10-23
Implement AddSign and PowerSign Optimizers

principal-academy:
This is a GPU and CPU implementation of Quoc Le's optimizers for deep neural networks, dubbed AddSign and PowerSign. Per the specification with no decay per iteration, found at arXiv:1709.07417 [cs.AI]

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/260)

PR #263, opened 2017-10-24
Fix leakyRelu and elu tests

Lewuathe:
We can use toEqual() instead of toBe() because it checks only equivalence of its value. 

```
Chrome 61.0.3163 (Mac OS X 10.12.6) math_cpu.leakyRelu basic FAILED
	Expected [ 3 ] to be [ 3 ].
	    at src/math/unaryop_test.ts:280:6 <-
src/math/unaryop_test.js:1480:46
	    at src/test_util.ts:199:31 <- src/test_util.js:322:32
	    at Object.<anonymous> (src/test_util.ts:173:19 <-
src/test_util.js:276:34)
```

It fixed on the local machine. But we need to also set username and password of BrowserStack as said in #260.

see #261 [CI](https://travis-ci.org/PAIR-code/deeplearnjs/builds/291699345?utm_source=github_status&utm_medium=notification)

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/263)

### DIFF
--- a/src/math/unaryop_test.ts
+++ b/src/math/unaryop_test.ts
@@ -778,7 +778,7 @@ import {Array1D, Array2D, Scalar} from './ndarray';
       const a = Array1D.new([0, 1, -2]);
       const result = math.leakyRelu(a);
 
-      expect(result.shape).toBe(a.shape);
+      expect(result.shape).toEqual(a.shape);
       test_util.expectArraysClose(result.dataSync(),
           new Float32Array([0, 1, -0.4]));
     });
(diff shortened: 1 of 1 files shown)


dsmilkov:
Nice! Thanks!

Regarding BrowserStack testing: It turns out that for security reasons we can't run WebGL Browserstack tests via travis: https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions
and we are not allowed to expose the Browserstack user/pass publicly since anyone can spend resources using our quota.

So I'm updating the travis conf file (#266) to not run tests for external PRs for the time being. This is not ideal, but strictly better than what we had before, since at least we will know if the tests pass as soon as we merge the PR in master.

PR #269, opened 2017-10-27
Add a playground demo which uses github gist as a storage mechanism.

nsthorat:
Adds a very very basic playground.

If folks want to add featured playground demos, they can do it from index.html.

image

image

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/269)


dsmilkov:
:lgtm_strong: Great work, just a few FYI comments.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[demos/playground/index.html, line 20 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVS0cKDv_fCEcGwTTC:-KxVS0cKDv_fCEcGwTTD:bkd394z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/index.html#L20)):*
> README code

will this link always be visible to users? I'm wondering if we should remove it.

---

*[demos/playground/playground.ts, line 42 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVReT2vLpXJEm8o7Fr:-KxVReT2vLpXJEm8o7Fs:bmpa9ck) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/playground.ts#L42)):*
> ```XML
> 
>   const init: RequestInit = {method: 'POST', body: JSON.stringify(content)};
>   const result = await fetch('https://api.github.com/gists', init);
> ```

fyi, IE doesn't have fetch, but Edge does, so I think we are ok.
source: http://caniuse.com/#search=fetch

---

*[demos/playground/playground.ts, line 97 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVSat2hGzx9VV12U5S:-KxVSat2hGzx9VV12U5T:bckotz9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/playground.ts#L97)):*
> ```XML
>     // Eval in an async() so we can directly use await.
>     eval(`(async () => {
>       ${jscontentElement.innerText}
> ```

Someone can write a gist that freezes someones computer, and share that gist - I don't have  solution for this - something to be aware of.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-:-KxVTBbK0PhXIZMJdnqh:b-szwzcw)*

nsthorat:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[demos/playground/index.html, line 20 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVS0cKDv_fCEcGwTTC:-KxfMuCTqohoFJAGRNhA:b-702v7z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/index.html#L20)):*

> README code

will this link always be visible to users? I'm wondering if we should remove it.
</blockquote></details>

What do you mean by always visible? This is an analogy to the blocks listing:
https://bl.ocks.org/mbostock

I figure we can make the page prettier later, adding pictures etc.

---

*[demos/playground/playground.ts, line 42 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVReT2vLpXJEm8o7Fr:-KxfN5nfL6E2GeTPN6AK:bivw7ch) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/playground.ts#L42)):*

fyi, IE doesn't have fetch, but Edge does, so I think we are ok.
source: http://caniuse.com/#search=fetch
</blockquote></details>

Yeah I figured it was fine, IE users probably wont use this tool :D 

---

*[demos/playground/playground.ts, line 97 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269#-KxVSat2hGzx9VV12U5S:-KxfND3Jo0Dme0vQ7_R3:bs5aavl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b083735e2466819bc0a3fe16113a0a6b0a50a474/demos/playground/playground.ts#L97)):*

Someone can write a gist that freezes someones computer, and share that gist - I don't have  solution for this - something to be aware of.
</blockquote></details>

Yeah, I know about this. I don't really know of a solution, though I don't think it's really a big deal. I would be much more concerned if we were setting a cookie on deeplearnjs.org.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/269)*

PR #280, opened 2017-10-31
Fix adamax nan

Lewuathe:
- [x] : Fix AdamaxOptimizer divergence
- [x] : Add AdamaxOptimizer in model builder

Beta1 in updating value should be powered by t.
see -> https://arxiv.org/abs/1412.6980
screen shot 2017-10-31 at 22 04 50
screen shot 2017-10-31 at 22 04 45

And also since initial infinity norm is zero, it could cause the divergence of update value. Adding small value epsilon can prevent this case. This technique is also used in Keras.
https://github.com/fchollet/keras/blob/master/keras/optimizers.py

screen shot 2017-10-31 at 21 52 14

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/280)


nsthorat:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/graph/optimizers/adamax_optimizer.ts, line 31 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280#-KyC3CSOY0S80CUpwwwN:-KyC3CSPFHXCDcPxgm4c:b-ofnerg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fcc742b9092bfd6feb5b1a3cf90feb58bf41e363/src/graph/optimizers/adamax_optimizer.ts#L31)):*
> ```XML
>       private beta2: number, specifiedVariableList?: Node[]) {
>     super(learningRate, specifiedVariableList);
>     this.eps = Scalar.new(1e-8);    
> ```

trailing spaces here

---

*[src/graph/optimizers/adamax_optimizer.ts, line 100 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280#-KyC38YffLkBUwUIWIMv:-KyC38YgOqwWJH33sWUc:bwb6tz5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fcc742b9092bfd6feb5b1a3cf90feb58bf41e363/src/graph/optimizers/adamax_optimizer.ts#L100)):*
> ```XML
>         oldWeightedInfNorm.dispose();
>       });
>       this.accB1 = keep(math.multiply(this.accB1, this.b1));
> ```

I think you need to dispose the old accB1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280)*

Lewuathe:

---

*[src/graph/optimizers/adamax_optimizer.ts, line 100 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280#-KyC38YffLkBUwUIWIMv:-KyGY2xwGY2UBo0b-FAg:b-y44w71) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fcc742b9092bfd6feb5b1a3cf90feb58bf41e363/src/graph/optimizers/adamax_optimizer.ts#L100)):*

I think you need to dispose the old accB1
</blockquote></details>

How can we dispose of only old accB1? accB1 is used in next iteration so it's needed to be kept.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280)*

nsthorat:

---

*[src/graph/optimizers/adamax_optimizer.ts, line 100 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280#-KyC38YffLkBUwUIWIMv:-KySY5Rjr6RbkwgO1rPh:b-t14sp4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fcc742b9092bfd6feb5b1a3cf90feb58bf41e363/src/graph/optimizers/adamax_optimizer.ts#L100)):*

How can we dispose of only old accB1? accB1 is used in next iteration so it's needed to be kept.
</blockquote></details>

you'll have to do something like this, unfortunately:

const oldAccB1 = this.accB1;
this.accB1 = keep(math.multiply(this.accB1, this.b1));
oldAccB1.dispose();

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280)*

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/280#-:-KyY3RNRtUFxEXIdaJls:b-jbsf8z)*

PR #281, opened 2017-10-31
Demo: Learning XOR

ManrajGrover:
This pull request adds demo for the famous learning xor problem.

@nsthorat I'm using `meanSquaredCost` here which ideally should not be the case. We should be using `softmaxCrossEntropyCost`. I'm facing some issue with it. I feel it has not been documented properly or maybe I'm not using it correctly. Could you please look into it?

Also, I'm building an interface around this. Hence, work is in progress.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/281)


arthurjdam:
@ManrajGrover I sent you a PR which should fix the travis build!

arthurjdam:
Fine with me!

arthurjdam:
It was I! How do I confirm?

nsthorat:
Okay so softmax cross entropy cost is used for multi class outputs, actually what you want is a log loss if you're just predicting a single value. If you want to use softmax cross entropy, you need to make your targetArray and y one hot representations ([1, 0] representing 0, and [0, 1] representing 1).

Hope this helps.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-:-KyBvFzMkDId2hj0ERZO:bglkobd)*

ManrajGrover:
@nsthorat Yes. It makes sense. I don't think `logLoss` is implemented currently. Actually, I found none of [tf.losses](https://www.tensorflow.org/api_docs/python/tf/losses) is implemented. Should I open a separate issue for the same?

nsthorat:
We have `Graph.log` which you can use as a loss tensor if it's a Scalar (there's nothing special about loss except it emits a scalar).

ManrajGrover:
@nsthorat I'm not sure if this implementation is correct. (New to this library 😅) It is giving `Nan` as cost. I've tried the same with `math` and it seems to be working correctly.

```ts
const costTensor = graph.reduceSum(
  graph.add(
    graph.multiply(graph.constant([-1]), graph.multiply(y, graph.log(graph.add(output, graph.constant([EPSILON]))))),
    graph.multiply(graph.constant([-1]), graph.multiply(graph.subtract(graph.constant([1]), y), graph.log(graph.add(graph.subtract(graph.constant([1]), output), graph.constant([EPSILON])))))
  )
)
```

Please let me know if I should share the whole code.

> We have Graph.log which you can use as a loss tensor if it's a Scalar (there's nothing special about loss except it emits a scalar).

I feel out of box implementations would be a good addition. :)

arthurjdam:
Anything I need to do to confirm? Looks like this PR is pretty much good to go

ManrajGrover:
@arthurjdam 

> That's my commit throwing a wrench in the works.

Haha, yeah.

> Anything I need to do to confirm? Looks like this PR is pretty much good to go

I'm actually creating a UI for this with a few options of changing number of hidden units, optimiser, learning rate etc. which will get included in this PR.

nsthorat:
@ManrajGrover ping the thread when you're ready for me to look again!

arthurjdam:
@nsthorat the commit Googlebot is complaining about is mine (https://github.com/ManrajGrover/deeplearnjs/commit/6042974f3515abc8e046cfa39d8d37d863169951 to be exact), though it doesn't seem to be acknowledging my consent. Obviously I love a good XOR example; is there any way to force a merge once ManrajGrover finishes what he was up to?

nsthorat:
Not sure I understand, can you fix the lint errors? If we commit those, then our build will fail.

ManrajGrover:
@nsthorat I feel this can be reviewed now.

Points I wanted to discuss:
1. UI freezes while training the neural network. I'm not sure if this is the only demo which faces this issue. (explore service workers?)
2. We should definitely add more out of box implementations of loss functions to the library.

nsthorat:
1) Service workers unfortunately won't help (the problem isn't UI thread churn). I left comments inline.
2) Agreed! We're working on eager mode so training-related stuff is getting punted until we release the new UI :)

BTW, in general, I'm going to stop accepting demos (technical debt on our end is growing). Once the new API is in, we'll move stuff over to tutorials / inline playground-style examples.

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[demos/xor/learn-xor.ts, line 86 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-L-8AH__XokT252aU6iI:-L-8AH__XokT252aU6iJ:bevix5t) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2ac8e43c5a0f3ca885e513ee04a508db62d6a481/demos/xor/learn-xor.ts#L86)):*
> ```XML
>           costTensor, feedEntries, 4, optimizer, CostReduction.MEAN);
>     }
>     loss = cost.get();
> ```

here too, if you use "await .val()" you'll get UI to not freeze

---

*[demos/xor/learn-xor.ts, line 100 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-L-8A7qu6jjmkPTuU_2U:-L-8A7qu6jjmkPTuU_2V:b2e8aeb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2ac8e43c5a0f3ca885e513ee04a508db62d6a481/demos/xor/learn-xor.ts#L100)):*
> ```XML
>     const val = session.eval(output, [{tensor: input, data: inputData}]);
>     result.push({
>       input: inputData.dataSync(),
> ```

this is why your UI is freezing. If you make this whole learnXOR function async, and then you "await inputData.data()" you'll see the UI thread freeze up. In general, don't use "getValues" or "dataSync", use the async "data()"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-:-L-8AUO1Qmn91WNp6HCg:brqyd71)*

ManrajGrover:
@nsthorat I've made the required changes. It still blocks the UI thread.

> Service workers, unfortunately, won't help (the problem isn't UI thread churn). I left comments inline.

I'm not sure. Does this mean service workers won't help offload library computations? 

> Agreed! We're working on eager mode so training-related stuff is getting punted until we release the new UI :)

Awesome! :D I would love to help!

> BTW, in general, I'm going to stop accepting demos (technical debt on our end is growing). Once the new API is in, we'll move stuff over to tutorials / inline playground-style examples.

Please let me know how I can help. Although I'm new to WebGL, I can surely assist in some way.

nsthorat:
Service workers will basically give you another JS UI thread. Most of our real heavy lifting happens on the GPU (2 orders of magnitude faster). Eventually, when we push a WebAssembly backend, or we use swiftshader, we might try to parallelize with service workers, but right now it's not going to be of much help (we just need to focus on freeing up the UI thread, and not doing things like blocking it until the GPU is done. This is what `getValues()` and `dataSync()` do).

Great! I will absolutely let you know. We're putting together a design doc for the eager mode + new API now, and I will make a github issue that outlines all the important pieces of it.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-:-L-Cqen7gvL29Dg-WH6K:b-ht945o)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/281#-:-L-Cr2kmKiX7dJzHpVyU:b-jbsf8z)*

PR #289, opened 2017-11-02
Use a sandboxed iframe in the deeplearn.js playground and visual improvements.

nsthorat:
* `eval`s javascript in an iframe
* Comments out the save button for now, let people just start from the versions above (we will ask OSS to add github user login saving stuff)
* Match the font of the playground to VSCode.
* Add console.clear.
* Let console.log also log to the original console.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/289)


dsmilkov:
:lgtm_strong:

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[demos/playground/frame.html, line 78 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/289#-KyICIwuQTqWExw-Rql6:-KyICIwuQTqWExw-Rql7:b-556n0f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/a33a29119057cd19426d6466745e578bb95e57ef/demos/playground/frame.html#L78)):*
> ```HTML
>   </div>
> 
>   <script>
> ```

any complications on moving this code to its own js file?

also add `'use strict';` on right below the `<script>` tag to enforce strict mode

---

*[demos/playground/frame.html, line 82 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/289#-KyIDtHvv8tmYkXwxeFY:-KyIDtHvv8tmYkXwxeFZ:bpv6ny) ([raw file](https://github.com/pair-code/deeplearnjs/blob/a33a29119057cd19426d6466745e578bb95e57ef/demos/playground/frame.html#L82)):*
> ```HTML
>     const htmlconsoleElement = document.getElementById('html');
>     const consoleElement = document.getElementById('console');
>     const errorConsoleElement = document.getElementById('error');
> ```

add a TODO for us to check if it works on firefox/safari

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/289#-:-KyIE-TQLjwEmL_ZsMh2:b-jbsf8z)*

PR #290, opened 2017-11-02
Don't use `params` for shader key construction. Use `userCode` instead.

### LINKED ISSUE #75: Change how we do shader program cache keys to be less vulnerable to missing params in the cache key.
Right now, if you miss adding a params for the shader program cache key, two programs that have different source code may get cached to the same WebGLProgram.

A couple solutions:
1) Cache based on the entire program source. This is durable, but may be expensive.
2) Make every GPGPUProgram take a Params object as the only item in the constructor. Parse out elements from Params to build the shader source, and automatically generate the key from the Params object.

dsmilkov:
Fixes #75

Also improve string concatenation based on results in https://jsperf.com/string-concatenation/14

**Perf timeline for shader key at master when running model builder on MNIST (focus on % of time):**

shaderkey-master

**Perf timeline for shader key at PR:**

shaderkey-pr

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/290)

### DIFF
--- a/src/math/webgl/addscaledmat_gpu.ts
+++ b/src/math/webgl/addscaledmat_gpu.ts
@@ -20,7 +20,6 @@ import {GPGPUProgram} from './gpgpu_math';
 
 export class AddScaledMatProgram implements GPGPUProgram {
   variableNames = ['A', 'B', 'c1', 'c2'];
-  params: Array<{}> = [];
   outputShape: number[];
   userCode: string;
   supportsBroadcasting = true;
--- a/src/math/webgl/argminmax_gpu.ts
+++ b/src/math/webgl/argminmax_gpu.ts
@@ -45,12 +45,10 @@ export function getArgMinMaxSnippet(
 export class ArgMinMaxProgram implements GPGPUProgram {
   variableNames = ['A'];
   outputShape: number[];
-  params: Array<{}>;
   userCode: string;
   numBatchDims: number;
 
   constructor(shape: number[], axes: number[], opType: 'min'|'max') {
-    this.params = [opType];
     const [outShape, reduceShape] =
         axis_util.computeOutAndReduceShapes(shape, axes);
     this.outputShape = outShape;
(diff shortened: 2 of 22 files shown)


nsthorat:
:lgtm_strong:

Nice cleanup!

---

Reviewed 22 of 22 files at r1.

---

*[src/math/webgl/gpgpu_math.ts, line 173 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/290#-Kxy1Yrr-yIAYikBfFIk:-Kxy1Yrs3Lg-yRGRldgG:b-j0xcea) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2993e9920fa825568e5d71824e47f66f18c0540e/src/math/webgl/gpgpu_math.ts#L173)):*
> ```XML
>   const keyBroadcast = (program.supportsBroadcasting === true).toString();
>   let key = program.constructor.name;
>   key += '_' + keyBroadcast + '_' + keyInputs + '_' + keyUserCode;
> ```

wanna put a comment here saying this is faster so somebody in the future doesn't try to use join?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/290#-:-Kxy1gW_D2915JOmXIxl:b-7a6p2l)*

dsmilkov:
Review status: 21 of 22 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/math/webgl/gpgpu_math.ts, line 173 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/290#-Kxy1Yrr-yIAYikBfFIk:-Kxy37-EJxR33cwxfDsF:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2993e9920fa825568e5d71824e47f66f18c0540e/src/math/webgl/gpgpu_math.ts#L173)):*

wanna put a comment here saying this is faster so somebody in the future doesn't try to use join?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/290)*

PR #294, opened 2017-11-03
Pull SqueezeNet and KNNImageClassifier out into separate npm modules.

nsthorat:
This change does the following:
* Renames TopKImageClassifier => KNNImageClassifier
* Puts a package.json in each of the models (squeezenet and knn-image-classifier). These each now have `tsconfig.json`s, and `package.json`s. SqueezeNet has a `demo.html` (I'll file a bug for somebody in OSS to finish the other model to add a demo).
* Puts a package.json in `demos/`. This allows us to make demos depend on the public API of `deeplearn` as well as on the public API of any models they may use, like `deeplearn-squeezenet`.
* Converts benchmarks to use the public API. This means we test against the `NDArrayMath` layer, which greatly simplifies the tests. We always run a warmup query, so shader compilation time is not an issue.
* Removes `logSumExp` benchmark file. It's already benchmarked in reduction ops.
* Calls initializeGPU from the constructor of the SqueezeNet model. This is because when we load standalone versions of both SqueezeNet and deeplearn together, the global state about GPU initialization is not shared. Once we decouple NDArray from storage mechanism, this should go away. `initializeGPU` is exported in this PR: https://github.com/PAIR-code/deeplearnjs/pull/306
* For some reason, the linter is now complaning about performance_rnn adding three numbers.. I had to cast them, but VSCode didn't seem to have the problem. Really bizarre.

Notes:
* `./scripts/watch-demo` works as it did previously.
* `npm run prep` now also runs `npm install` in the demos directory.
* Think of `package.json`s as BUILD rules in Bazel.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/294)


dsmilkov:
:lgtm_strong:

---

Reviewed 47 of 47 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[package.json, line 18 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-KyIbxFP_eZztnmh0L2L:-KyIbxFP_eZztnmh0L2M:b-usb5rp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/00c7330936b9b8d36aaa6a1ec043dd20c4c5964f/package.json#L18)):*
> "bower": "~1.8.0",

move `bower`, `@types/polymer` and `polymer-bundler` since they are not dependencies of the core library, but dev deps the demos/

---

*[package.json, line 41 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-KyIbgwmP4qqAm_39sCN:-KyIbgwmP4qqAm_39sCO:b48ykbg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/00c7330936b9b8d36aaa6a1ec043dd20c4c5964f/package.json#L41)):*
> bower install

move this command to `demos/package.json` and change the relative html links to `../bower_components` instead of `../../bower_components`

---

*[tsconfig.json, line 22 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-KyIcmqeYh9nbsyKwk9t:-KyIcmqeYh9nbsyKwk9u:b-2dhv5j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/00c7330936b9b8d36aaa6a1ec043dd20c4c5964f/tsconfig.json#L22)):*
> ```JSON
>     "exclude": [
>       "starter/",
>       // Exclude models so we don't have to run npm install in every
> ```

remove comments since they are not supported by json

---

*[demos/package.json, line 8 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-KyIdIKmU7hgg4LyO7OU:-KyIdIKmU7hgg4LyO7OV:b-bxivi2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/00c7330936b9b8d36aaa6a1ec043dd20c4c5964f/demos/package.json#L8)):*
> ```JSON
>     "deeplearn": "~0.3.7",
>     "deeplearn-squeezenet": "~0.1.2",
>     "deeplearn-knn-image-classifier": "0.1.0"
> ```

add `~` to the version

---

*[demos/package.json, line 13 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-KyId5uNAGbNN_y7gJ9G:-KyId5uNAGbNN_y7gJ9H:b-sr8slw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/00c7330936b9b8d36aaa6a1ec043dd20c4c5964f/demos/package.json#L13)):*
> ```JSON
>     "type": "git",
>     "url": "https://github.com/PAIR-code/deeplearnjs.git"
>   },
> ```

add `"private": true` at the top level to prevent accidental publication. See https://docs.npmjs.com/files/package.json

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/294#-:-KyIfV_aroh_dyW2LRyL:b-jbsf8z)*

PR #298, opened 2017-11-03
Create git tag when running make-version script

glortho:
This clears the way for automated changelog diffs. See #262.

An enhancement for this workflow could be to let `make-version` accept an argument that matches the [npm version api](https://docs.npmjs.com/cli/version) -- namely `patch`, `minor`, or `major`.

We could then call `npm version ${arg}`, which would increment the package.json version, and create both a matching git tag and version commit.

Omitting the arg could just revert to the functionality covered by this PR. Let me know what you think.

For example:

```
$ npm run make-version -- patch
$ npm run make-version -- minor
$ npm run make-version -- major
$ npm run make-version
```

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/298)


dsmilkov:
:lgtm_strong: Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/298#-:-Ky2j8kshkR3qbeTEwbc:bfdlkw5)*

PR #299, opened 2017-11-04
Add LeakyRelu ops

Lewuathe:
Though `LeakyRelu` activation function was added in https://github.com/PAIR-code/deeplearnjs/pull/231, there is no corresponding ops function.

see: https://github.com/PAIR-code/deeplearnjs/issues/223

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/299)


nsthorat:
Thanks for adding this!

Unfortunately there's still more you need to do if you want to see this on the Graph.

You'll need to do the following:

1) Add a LeakyReluNode here: https://github.com/PAIR-code/deeplearnjs/blob/master/src/graph/graph.ts
2) Add a top level graph method that uses it
3) Add an emitter statement that converts the Node into an Operation: https://github.com/PAIR-code/deeplearnjs/blob/master/src/graph/operation_emitter.ts

I want to let you know, since you're a pretty active contributor, that this flow for adding differentiable ops is going to be made vastly simpler when we do the eager mode we're talking about. If you want to add training related stuff, you may want to wait a couple months for that.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/math/math.ts, line 1319 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyBwVcv5ZUXl92tLXYW:-KyBwVcv5ZUXl92tLXYX:bxs1689) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1056b2452a390ce7f37c91e82ba19b4ef4b4d4cc/src/math/math.ts#L1319)):*
> ```XML
>    * Computes step of the input NDArray element-wise, y=1 if x>0|alpha if x<=0.
>    *
>    * @param ndarray The input NDArray.
> ```

add alpha param here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-:-KyBxh5QmxFHIus8VrMF:beiyiyp)*

nsthorat:
Review status: 0 of 8 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/activation_functions.ts, line 58 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyBynXv2GwtgunTBvu8:-KyBynXv2GwtgunTBvu9:b38xhm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1056b2452a390ce7f37c91e82ba19b4ef4b4d4cc/src/math/activation_functions.ts#L58)):*
> ```XML
> 
> export class LeakyReluFunc implements ActivationFunction {
>   private alpha: number;
> ```

this will never get cleaned up, can you add a dispose method to `ActivationFunction` as well as override the `ElementWiseActivation` op's dispose to call `func.dispose()`

---

*[src/math/activation_functions.ts, line 65 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyByfq8YDjoNQmgY0_W:-KyByfq8YDjoNQmgY0_X:bm9m7s3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1056b2452a390ce7f37c91e82ba19b4ef4b4d4cc/src/math/activation_functions.ts#L65)):*
> ```XML
> 
>   output<T extends NDArray>(math: NDArrayMath, x: T) {
>     return math.scope(() => {
> ```

you actually dont need math.scope here - there's only 1 operation and it wont get cleaned up because of the return, same with in `der`. Sorry about that, this is my fault (I'll update the rest of this file)

---

*[src/math/webgl/unaryop_gpu.ts, line 68 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyBzHqFcS0M2PEEeNPd:-KyBzHqFcS0M2PEEeNPe:b-chwn7d) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1056b2452a390ce7f37c91e82ba19b4ef4b4d4cc/src/math/webgl/unaryop_gpu.ts#L68)):*
> ```XML
> export function STEP(alpha = 0.0) {
>   return `
>   return (x == x) ? (x > 0.0 ? 1.0 : float(${alpha})) : x;
> ```

can you indent this 2 more

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299)*

nsthorat:
One minor comment, then I will merge!

---

Reviewed 1 of 8 files at r1, 6 of 9 files at r2.
Review status: 7 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 1245 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyqWwzfKFV0YbcQ7poN:-KyqWwzfKFV0YbcQ7poO:bon2izl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/eb94142bf6ff6c97a0b5d349579295bbd0ee391f/src/math/math.ts#L1245)):*
> ```XML
>    * @return {NDArray}
>    */
>   leakyRelu<T extends NDArray>(ndarray: T, alpha = Scalar.new(0.2)): T {
> ```

We need to keep alpha as not a scalar, otherwise this Scalar is a global and we want to control this memory. By splicing the alpha value directly into the shader, we compile it directly into the program, and it becomes part of the shader key.

---

*[src/math/math_cpu.ts, line 699 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyqXLWg-rcEoyjhysu6:-KyqXLWg-rcEoyjhysu7:b6ia0xv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/eb94142bf6ff6c97a0b5d349579295bbd0ee391f/src/math/math_cpu.ts#L699)):*
> ```XML
>   }
> 
>   protected stepInternal<T extends NDArray>(ndarray: T, alpha = 0): T {
> ```

here too, we need the signature of alpha to be consistency across math types

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-:-KyqXS0o5aF96KCxrlW2:b-fyhrsq)*

Lewuathe:
Review status: 7 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 1245 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyqWwzfKFV0YbcQ7poN:-Kys2oFGvowO3uHl9FMI:bexbcr5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/eb94142bf6ff6c97a0b5d349579295bbd0ee391f/src/math/math.ts#L1245)):*

We need to keep alpha as not a scalar, otherwise this Scalar is a global and we want to control this memory. By splicing the alpha value directly into the shader, we compile it directly into the program, and it becomes part of the shader key.
</blockquote></details>

LeakyReluFunc disposes `alpha` (Scalar) parameter in its own `dispose` method. So I thought it's necessary to make this default parameter compatible with Scalar. 

We can pass `alpha` parameter as `number`. But in that case we don't need `dispose` method in LeakyReluFunc . Is that assumption right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299)*

nsthorat:
Review status: 7 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/math/math.ts, line 1245 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-KyqWwzfKFV0YbcQ7poN:-Kyw5qVG3jBKKOVwKVw-:bfy5rqo) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d609af4ffa75e23465820f63711f76e4ed918293/src/math/math.ts#L1245)):*

LeakyReluFunc disposes `alpha` (Scalar) parameter in its own `dispose` method. So I thought it's necessary to make this default parameter compatible with Scalar. 

We can pass `alpha` parameter as `number`. But in that case we don't need `dispose` method in LeakyReluFunc . Is that assumption right?
</blockquote></details>

Great point, we need to make LeakyReluFunc also take a number.

Technical details:
If we call "get" here, we introduce a CPU / GPU interlock in library code, which we don't want to do. There are two options 1) it's an NDArray argument - in this case we have to make it an official input sampler, with a texture, etc. This would not get spliced into the shader source, but rather it would be a texture sampler or 2) it's spliced in as a number, this is what TensorFlow does. If we do it that way, we don't want to let it be an NDArray because that could be in GPU memory and then we introduce a download before splicing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299)*

nsthorat:
:lgtm_strong:

---

Reviewed 1 of 8 files at r1, 3 of 9 files at r2, 2 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/299#-:-Kz-6c2Y1p1PwNAk4kdr:b-jbsf8z)*

nsthorat:
Thanks for bearing with me. Once we add new API functionality, we cant go back :)

PR #301, opened 2017-11-04
Don't check for `WEBGL_color_buffer_float` since `OES_texture_float` implicitly enables it

dsmilkov:
Fixes a bug in the Environment where we would always fall back to byte packing whenever Web 2.0 was not enabled, even though Web 1.0 supports floats.

Checking if `OES_texture_float` is available is enough, since that implicitly enables `WEBGL_color_buffer_float`, according to the WebGL specs. See https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_color_buffer_float

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/301)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -108,17 +108,12 @@ function isFloatTextureReadPixelsEnabled(webGLVersion: number): boolean {
 
   const gl = getWebGLRenderingContext(webGLVersion);
 
-  let floatExtension;
-  let colorBufferFloatExtension;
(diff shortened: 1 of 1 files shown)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/301#-:-Ky6pbK7DkBnyIwgXlsS:b-jbsf8z)*

PR #307, opened 2017-11-06
Fast style transfer demo

reiinakano:
Here's my initial PR for fast style transfer demo.

Few things:
1. I have no idea why the demo-header and demo-footer won't show up on my demo... I've checked for any difference with the other demos and I can't figure it out. Can't shake the feeling I'm missing something super obvious.
2. The checkpoints are not included in this commit. Should probably upload it to your Google Cloud instead of directly to this repository.
3. Let me know if the FAQ and credits below the whole thing should be removed to make it more similar to the other demos.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/307)


nsthorat:
This is a wonderful contribution, thank you for taking the time to check it into demos. Great work!

A few high level things:
1) Polymer sucks. I really regret adding it in the first place. Daniel and I will debug header / footer stuff after this is checked in.
2) Yup, uploading to Google Cloud SGTM. Can you upload the weights / manifest somewhere zipped up for me to download (Google Drive should work).
3) FAQ seems reasonable to me! :)

Other high level comments:
* Unfortunately the typescript files need license headers, can you copy the license headers from the other TS files?
* Once you check this in - do you care if I change a few things? We have a designer in house that we can consult to make it pretty. I can also see if live webcam style transfer is reasonable (even if it's laggy, it will still be cool, and on beefy graphics card it may not be super unreasonable).

---

Reviewed 2 of 13 files at r1.
Review status: 2 of 13 files reviewed at latest revision, 7 unresolved discussions.

---

*[demos/fast-style-transfer/fast-style-transfer.ts, line 274 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGTxcfr4dhJHxb2uq8:-KyGTxcfr4dhJHxb2uq9:boi0cze) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/fast-style-transfer.ts#L274)):*
> ```XML
>   }
> 
>   private drawOnCanvas(ndarray: Array3D) {
> ```

Ah, here is where you're going to see a bottleneck. We have utilities that let you draw a texture directly to canvas, without ever doing a CPU download. This should improve performance. We do this in a couple demos, `imagenet` & `nn-art`

What you should do is the following:

1) Construct your GL from the canvas that you want to draw to like this:
https://github.com/PAIR-code/deeplearnjs/blob/master/demos/nn-art/cppn.ts#L75

2) In the constructor, call this function and save the shader program:
https://github.com/PAIR-code/deeplearnjs/blob/64814ca08c5c4724bdc91c9ccde82893169dcd26/src/math/webgl/render_ndarray_gpu_util.ts#L22
Something like this:
`this.renderShader = render_ndarray_gpu_util.getRenderRGBShader(this.gpgpu, destinationWidth` where destinationWidth is the width in pixels of the output canvas. If this changes, you may have to build a new render shader.

3) Now, instead of `drawOnCanvas`, you just have to call `render_ndarray_gpu_util.renderToCanvas(this.gpgpu, this.renderShader, ndarray.getTexture());`
Where `ndarray` is the Array you want to draw. You don't have to give it the canvas because you constructed the WebGLRenderContext (this.gl) from the canvas and this method will bind the canvas to the framebuffer and render.

---

*[demos/fast-style-transfer/net.ts, line 8 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGVgCWSCkebX1zlDmp:-KyGVgCWSCkebX1zlDmq:b8oappk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/net.ts#L8)):*
> ```XML
>     document.URL.substr(0,document.URL.lastIndexOf('/')) + '/ckpts/';
> 
> export class TransformNet {
> ```

can you make this extend `Model`? Then `loadVariables` should become `load` and `infer` should become `predict`.

---

*[demos/fast-style-transfer/net.ts, line 8 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGVov_sFQDIQHabCox:-KyGVov_sFQDIQHabCoy:b-9i7f1z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/net.ts#L8)):*
> ```XML
>     document.URL.substr(0,document.URL.lastIndexOf('/')) + '/ckpts/';
> 
> export class TransformNet {
> ```

can you link to the checkpoint you used / an arxiv paper / website?

---

*[demos/fast-style-transfer/net.ts, line 17 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGWrFSclR_SHgf7Ru6:-KyGWrFSclR_SHgf7Ru7:b-cxxsbh) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/net.ts#L17)):*
> ```XML
>    * variables have all been loaded.
>    */
>   loadVariables(): Promise<void> {
> ```

Can you short circuit if the variables have been loaded (this.variables != null)?

I noticed in your demo that you have to download weights again every time you press style transfer.

---

*[demos/fast-style-transfer/styletransfer.html, line 161 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGWQhpAnrSK0bmbBQh:-KyGWQhpAnrSK0bmbBQi:b-467x9j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/styletransfer.html#L161)):*
> ```HTML
>   <input type="file" id="fileSelect" style="display: none" accept="image/x-png,image/gif,image/jpeg"/>
> 
>   
> ```

remove the comment

---

*[demos/fast-style-transfer/styletransfer.html, line 181 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGWYIP2s35XEBCZCXG:-KyGWYIP2s35XEBCZCXH:b-8ebk7u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/styletransfer.html#L181)):*
> ```HTML
>   <div class="l--body">
>     <h2>What is this?</h2>
>     <p>This is an implementation of the Fast Neural Style Transfer algorithm running purely on the browser using the Deeplearn.JS library. Basically, a neural network attempts to "draw" one picture, the Content, in the style of another, the Style.</p>
> ```

"Deeplearn.JS" => "deeplearn.js"

---

*[demos/fast-style-transfer/styletransfer.html, line 202 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyGWe6i-gx-lBrv4jU1:-KyGWe6i-gx-lBrv4jU2:b-8ebk7u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/7bccd0d73bc73188a2654989ea02e22760714bfa/demos/fast-style-transfer/styletransfer.html#L202)):*
> ```HTML
>       <li>The authors of the paper introducing <a href="https://arxiv.org/abs/1603.08155">Real-Time Style Transfer</a>.</li>
>       <li>The author of the <a href="https://github.com/lengstrom/fast-style-transfer">fast-style-transfer</a> Github repository.</li>
>       <li>The authors of <a href="https://github.com/PAIR-code/deeplearnjs">Deeplearn.JS</a></li>
> ```

"Deeplearn.JS" => "deeplearn.js"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-:-KyGTL_7ZtyZPh3P_scI:b-mrax5k)*

reiinakano:
> Once you check this in - do you care if I change a few things? We have a designer in house that we can consult to make it pretty. I can also see if live webcam style transfer is reasonable (even if it's laggy, it will still be cool, and on beefy graphics card it may not be super unreasonable).

Sure, no problem! If you go with live webcam style transfer, I would suggest to keep the cam snapshot option. I think the majority of low-power laptops will have trouble if it's continuously processing.

Anyway, I'm currently having a problem with rendering straight to the canvas. Pretty sure I followed all your steps correctly. Right now I'm trying to render not the result image, but the Content image just to debug things. It renders into the canvas but the picture is in black and white so there's definitely something wrong.. If I try to render the actual result image, I don't get anything at all.

nsthorat:
The render code expects value between 0 - 1, can you make sure you've normalized the output? If that doesn't work, could you stage the HTML / JS into a github pages so I can step through it?

Sure, I'll make sure we keep static image loading from the webcam.

---

Review status: 0 of 13 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-:-KyHm3RBVue1BRuq-Mjf:bmy07p2)*

reiinakano:
Ahh that makes sense, the previous code rendered with 0-255 so I assumed the same. It works now.

I think I've addressed all your comments. Have one problem though, when the short circuit for variable loading "activates", the button for "Start Style Transfer" doesn't become disabled anymore. I've noticed the same problem before - while the image is running through the net, I couldn't change parts of the UI to show the status. I think this has something to do with not being able to render the UI while GPU processing is taking place? Do you have any suggestions to handle this?

PS. Not sure where the Travis error is from, when I run `npm run lint --silent` on my computer it doesn't error out.

reiinakano:
Here's the Drive link with the checkpoints. 
https://drive.google.com/drive/folders/1mbC_TC50CIJqOhHPaKnpj9ObJZD6sJsu?usp=sharing

nsthorat:
Ah sorry about that, yeah we updated the workflow. Demos have their own package now, so you just have to run `npm run prep` which will do 2 `npm install`s. Apologies.

Checkpoints are uploaded under http://storage.googleapis.com/learnjs-data/checkpoint_zoo/transformnet/

For example:
http://storage.googleapis.com/learnjs-data/checkpoint_zoo/transformnet/la_muse/manifest.json

LGTM after this - will merge.

Thank you so much for bearing with me :)

---

Reviewed 10 of 13 files at r1, 3 of 3 files at r5.

---

*[demos/fast-style-transfer/fast-style-transfer.ts, line 259 at r5](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyNjQ_54yc3ybykdJCw:-KyNjQ_54yc3ybykdJCx:bqx6ynr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2875798d6bb580507ee3a3c546d4741caaf2bf72/demos/fast-style-transfer/fast-style-transfer.ts#L259)):*
> ```XML
> 
>   async runInference() {
>     
> ```

remove blank line

---

*[demos/fast-style-transfer/net.ts, line 68 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyIH5mvQoRNIExpZFBM:-KyIH5mvQoRNIExpZFBN:b-4jg0qf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4f1abb3e4ca6a2c5aeb9e5b5003c3454ee3145f4/demos/fast-style-transfer/net.ts#L68)):*
> ```XML
>       const outTanh = this.math.tanh(convT3);
>       const scaled = this.math.scalarTimesArray(Scalar.new(150), outTanh);
>       const shifted = this.math.scalarPlusArray(Scalar.new(255./2), scaled);
> ```

can you construct these scalars in the constructor, this allocates and uploads a new NDArray each pass through and will leak textures and cause performance issues

---

*[demos/fast-style-transfer/net.ts, line 43 at r5](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-KyMHNA4teBaqgoeQGhV:-KyMHNA4teBaqgoeQGhW:b5wnuas) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2875798d6bb580507ee3a3c546d4741caaf2bf72/demos/fast-style-transfer/net.ts#L43)):*
> ```XML
>   async load(): Promise<void> {
>     if (this.variableDictionary[this.style] == null) {
>       console.log('downloading weights for first time');
> ```

remove console log

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-:-KyMNA8SOKCkUPn4H5IS:b-hld3e)*

reiinakano:
Done.

Do you have any suggestions for this comment?

> Have one problem though, when the short circuit for variable loading "activates", the button for "Start Style Transfer" doesn't become disabled anymore. I've noticed the same problem before - while the image is running through the net, I couldn't change parts of the UI to show the status. I think this has something to do with not being able to render the UI while GPU processing is taking place? Do you have any suggestions to handle this?

nsthorat:
:lgtm_strong:

There are lots of reasons this can happen, but I don't see any obvious reasons in your code as is.

I can merge this, then I'll start a demo / debug it and we can iterate. Does that sound okay to you?

---

Review status: 11 of 13 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/307#-:-KyOo0X6cSMVynVTAZ1J:bp1213p)*

reiinakano:
Sure!

PR #314, opened 2017-11-07
Add math.tile

reiinakano:
Here's my initial commit of `math.tile` mirroring https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.tile.html and https://www.tensorflow.org/api_docs/python/tf/tile.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/314)


nsthorat:
This is an awesome contribution, thank you! Few comments below.

---

Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 934 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-KySaJKG3BgRiWk723eG:-KySaJKHc4GIdVyTlf0q:b-kwztex) ([raw file](https://github.com/pair-code/deeplearnjs/blob/56e422c95881377c380b42c16883d6c25b1668b7/src/math/math.ts#L934)):*
> ```XML
>    */
>   tile<D extends keyof DataTypes, T extends NDArray<D>>(
>       a: T, multiples: number[]): T {
> ```

can you call this reps for consistency with numpy?

---

*[src/math/tile_test.ts, line 24 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-KySd0iqjQy54WKXED6R:-KySd0iqjQy54WKXED6S:bm0her9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/56e422c95881377c380b42c16883d6c25b1668b7/src/math/tile_test.ts#L24)):*
> ```XML
> 
> // math.tile
> {
> ```

can you write some tests for ints and bools since you also support tiling those?

---

*[src/math/webgl/tile_gpu.ts, line 37 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-KySda7sJK3asap9pigR:-KySda7sJK3asap9pigS:blwxljd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/56e422c95881377c380b42c16883d6c25b1668b7/src/math/webgl/tile_gpu.ts#L37)):*
> ```XML
> 
>     this.userCode = `
>     void main() {
> ```

indent this block 2 more

---

*[src/math/webgl/tile_gpu.ts, line 63 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-KyS_pqkNlz43_7TjDJR:-KyS_pqkNlz43_7TjDJS:bp17lul) ([raw file](https://github.com/pair-code/deeplearnjs/blob/56e422c95881377c380b42c16883d6c25b1668b7/src/math/webgl/tile_gpu.ts#L63)):*
> ```XML
> }
> 
> function getDataType(rank: number): string {
> ```

since we use this in another place, can you move this to a common place: `shader_compiler.ts`?

You should also rename this to something like `getCoordsDataType`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-:-KySdivXQAg8KiTF1R19:b-i7wewc)*

reiinakano:
I've addressed your other comments except for the last one pending some clarifications.

Do you want `getCoordsDataType` to be some kind of utility function that could be called with `shader_compiler.getCoordsDataType`?

Or do you want it to be included in the compiler code itself? Would calling `${dtype} resRC = getOutputCoords();` outside the `main` block work? This way we'd be able to directly call `resRC` without `getOutputCoords`. Although there's probably a reason you made it a function instead of a constant.

nsthorat:
getCoordsDataType should be identical to what you're doing, just move the function location itself to `shader_compiler` as a util function. It should be a javascript function that returns a string for use in the spliced shader.

You can't put getOutputCoords() outside the main block, the main() block is the actual execution function of the fragment shader itself. Yep, it's a function because it's dynamically computed based on gl_FragCoord!

We're going to write a little more about how we actually do some of the WebGL stuff in a tutorial soon, will keep you posted.

Hope this helps.

---

Reviewed 2 of 5 files at r2, 1 of 1 files at r3.
Review status: 3 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/tile_test.ts, line 78 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-Kya_fd62YxGiZxwMVgm:-Kya_fd62YxGiZxwMVgn:b-x2uzgk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2ee99512a95f105982d40dbd8c84f8428ff48ec4/src/math/tile_test.ts#L78)):*
> ```XML
>       expect(t2.shape).toEqual([6]);
>       expect(t2.dtype).toBe('bool');
>       expect(t2.getValues()).toEqual(new Uint8Array([1, 0, 1, 1, 0, 1]));
> ```

sorry to be a pain, can you just replicate the Float tests for Int / Bool (make sure we also are checking for NaNs, there are subtle NaN details for non-floats)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-:-Kya_rIr8uu9iDG5HqRl:bb8ligj)*

reiinakano:
Done!

nsthorat:
:lgtm_strong:

---

Reviewed 6 of 6 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/314#-:-KygOex_f8ombQRV-eh2:b-jbsf8z)*

PR #327, opened 2017-11-09
Add math.gather

reiinakano:
Initial commit for `math.gather` mirroring https://www.tensorflow.org/api_docs/python/tf/gather.

I was going to call it `embedding_lookup` but saw that it was closer to `gather`, since the lookup from TF supports sharding.

My biggest issue with this is in the GPU implementation,  I create a new `Array1D` for the `indices` parameter. I need it to be in a texture to work as a dynamic lookup table since GLSL arrays apparently can't do this (see [this](https://stackoverflow.com/questions/19529690/index-expression-must-be-constant-webgl-glsl-error) and [this](https://stackoverflow.com/questions/6247572/variable-array-index-not-possible-in-webgl-shaders)). Of course, this means there is a memory leak as I'm not sure where I can dispose of the texture. 

This can be solved by requiring `indices` to be an `Array1D` itself and letting the user handle disposing it properly. What do you think?

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/327)


nsthorat:
Hey @reiinakano I apologize for dropping this. We've updated the way we do kernels (math calls into a backendEngine with a typed object that configures the kernel execution), would you be up for updating to the new way and we can review this?

Let us know if you need help understanding the new way to execute kernels.

reiinakano:
Hi yeah sure, I'll study the new way next week when I'm off work.

nsthorat:
Hey @reiinakano, just curious if you're going to finish this off. Sorry about the delay, and thanks so much for this change :)

PR #332, opened 2017-11-09
srcObject fix for safari webcam

jameswex:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/332)

### DIFF
--- a/demos/teachable_gaming/teachable_gaming.ts
+++ b/demos/teachable_gaming/teachable_gaming.ts
@@ -194,7 +194,11 @@ export class TeachableGamingDemo extends TeachableGamingDemoPolymer {
       navigator.getUserMedia(
           {video: true},
           (stream) => {
-            this.webcamVideoElement.src = window.URL.createObjectURL(stream);
+            try {
+              this.webcamVideoElement.srcObject = stream;
+            } catch (error) {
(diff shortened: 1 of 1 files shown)


dsmilkov:
:lgtm_strong:

---

---

*[demos/teachable_gaming/teachable_gaming.ts, line 198 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/332#-KyX_nBGl0gjI4nJru7E:-KyX_nBHRns88tJCgFFF:b-9rmbiv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/bfb9fef8cc1facbbc54f258b3e0d7bcbce555c11/demos/teachable_gaming/teachable_gaming.ts#L198)):*
> ```XML
>           (stream) => {
>             try {
>               this.webcamVideoElement.srcObject = stream;
> ```

is there a way to test for this without a try/catch? e.g. test if this.webcamVideoElement.srcObject != null ? if not no worries.

I just feel like try/catch should be last resort since it's a catch all and we might miss other errors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/332#-:-KyXa5-nMD1RSAVT0XVn:b-jbsf8z)*

PR #339, opened 2017-11-12
Fix missing space in matMul rank assert

mikelseverson:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/339)


dsmilkov:
Thanks!

Issue #267, opened 2017-10-26
Import Caffe models to deeplearn.js


chaosmail:
To make it as simple as possible for the user, it would be great to parse both `deploy.prototxt` (model layout) and `*.caffemodel` (model weights) files directly within the browser. As both files are simply protos this should be possible. After playing around, I found that the actual model architecture is embedded in the original `*.caffemodel` file itself - hence we need to only load this file. Heck, even the training mean is included!

The Caffe proto definition can be found here: https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto

The following steps would be necessary:

- [x] Parse the `caffe.proto` to Typescript module using dcodeIO (https://github.com/dcodeIO/ProtoBuf.js/#pbts-for-typescript)

```sh
$ pbjs -t static-module -w commonjs -o compiled.js file1.proto file2.proto
$ pbts -o compiled.d.ts compiled.js
```
An alternative would be to serve the `caffe.proto` file and parse it at runtime (discussion of trade-offs: https://github.com/dcodeIO/ProtoBuf.js/#reflection-vs-static-code), which would not require a build step. I would prefer a build/compile step during dev than more dependencies at runtime. What do you guys think?

- [x] Add `protobuf.js` as a dependency. Maybe even the light version is enough for us.

- [x] Fetch the binary `*.caffemodel` file, parse it to NetParameter and convert it to a deeplearn.js model

regarding 1.:

As a proof-of-concept, I converted the `caffe.proto` file into a `caffe.json` proto (using `cli/bin/pbjs -t json tmp/caffe.proto > tmp/caffe.json`) and successfully loaded the `squeezenet.caffemodel` (both weights and structure) from the original file only. Here is a working demo:

```html
<html>
  <head>
    <!-- https://github.com/dcodeIO/protobuf.js -->
    <script src="libs/protobuf.min.js"></script>
  </head>
  <body>
    <script>
  
    function fetchArrayBuffer(uri) {
      var req = new Request(uri);
      return fetch(req).then(function(res) {
        return res.arrayBuffer();
      });
    }

    protobuf.load("caffe.json", function(err, root) {
      
      // Read caffe proto
      var caffe = root.lookup('caffe');
 
      // Fetch caffemodel
      fetchArrayBuffer('squeezenet_v1.1.caffemodel').then(function(res){

        var model = caffe.NetParameter.decode(new Uint8Array(res));
        console.log(model);
      });
    });
    </script>
  </body>
[...]

PR #346, opened 2017-11-13
Add caffemodel loading and parsing

chaosmail:
Implements #267 

I decided to create a PR to merge the current status of development, as the next steps (parsing `deploy.prototxt` and creating a static operations DAG) require some bigger changes and more input from you. This PR adds the ability to load and parse `*.caffemodel` files directly within JavaScript (using protobuf decode) and extract the weights for usage in deeplearn.js.

~~In `scripts/build-caffe-proto.sh`, the latest `caffe.proto` file is pulled from https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto and converted into a JavaScript and a Typescript definition file using protobuf.js and stored in `src/contrib/caffe`.~~

In `models/caffe/scripts/build-proto.sh`, the latest `caffe.proto` file is pulled from https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto and converted into a JavaScript and a Typescript definition file using protobuf.js and stored in `models/caffe/proto`.

## Caffemodel

The `*.caffemodel` file contains the model definition, layer parameters and weights after a specific phase (train, test), as well as the preprocessing offset (training mean, crop parameters, etc.). Unfortunately, many `*.caffemodel` files contain all layer definitions (and model graph) from the training phase (e.g. cross validation splits, labels, etc.) without the correct information about the phase (train, test). The `*.caffemodel` is a binary proto and can be parsed with the caffe.proto definition and protobuf.js as a dependency.

## Prototxt

The `*.prototxt` file contains the model definition and parameters for a specific phase (train, test). Most of the times, authors provide a `deploy.prototxt` file containing the test definition of the model - which contains only the relevant layers for inference. The `*.prototxt` is a textual proto definition and can be parsed with a custom parser (e.g. https://github.com/brotchie/protobuf-textformat/blob/master/src/parser.js) and the parser dependency (e.g. Parsimmon).

## Recommendation for Deeplearn.js

In my opinion, the best way to support Caffe (models) would be to integrate caffemodel parsing into the core of deeplearn.js (as implemented in this PR) - such that properly generated caffemodel files (that contain the correct phase variables per layer) can be loaded directly. It would be nice to also provide a script to convert caffemodels with missing phase variables to correct caffemodels e.g. using python, and eventually host these models on https://storage.googleapis.com/learnjs-data/checkpoint_zoo/ as well.

As a second option, one could provide a separate module (e.g. deeplearn-caffe-utils.min.js), to load and parse `*.prototxt` files and apply the test structure to the caffemodel. This way, the parser dependency would not affect the size of deeplearn.js while making it possible to easily load all available caffe models directly within JavaScript (without compiling).

## Testing the SqueezenetCaffe model

Compile `models/squeezenet` and replace the following line in `demos/imagenet/imagenet.ts`

```ts
import {SqueezeNet} from 'deeplearn-squeezenet';
```

with

```ts
import {SqueezeNetCaffe as SqueezeNet} from 'deeplearn-squeezenet';
```

> Unfortunately I couldn't run this anymore after the rebase on master.

## A simple example for debugging

This is a simple snippet to load SqueezeNet weights from the original model:

```html
<!doctype html>
<html>
  <head>
    <script src="dist/deeplearn.js"></script>
  </head>
  <body>
    <script>
      var GITHUB_CDN = 'https://rawgit.com/DeepScale/SqueezeNet/master/SqueezeNet_v1.1/';
      var model = new deeplearn.CaffeModel(GITHUB_CDN + 'squeezenet_v1.1.caffemodel');
      model.loadVariables();
    </script>
  </body>
</html>
```
Now you can open the developer tools and see the parsed model definition in `model.caffemodel`.

## TODOs

There a few TODOs that I marked directly in the source code, e.g.
- [x] storing variables as Maps instead of objects
- [x] naming conventions for the weights (besides conv), e.g. batchnorm, etc.
- ~~support more general preprocessing, such as cropping~~

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/346)


nsthorat:
Wow, this is wonderful. Thank you for the detailed notes, it is really helpful with this large PR.

The comments I have are high level:

* I'm a little hesitant putting the typings and parser in `src/` because of bloat (we compile every file and serve it with every bundle request). We've moved models out to their own npm packages so they can be used independently, and it seems like we'll only need to use the caffe files when we've got a model that's being ported from caffe. Would it be possible to put these utils in models somewhere to be shared among multiple caffe models?
* Can you put the caffe squeezenet in its own npm package? I don't want to bloat that one either since people will be directly using it.
* caffe.js does automatic model parsing and execution right? I think it would be nice to see an end to end demo of that of some model (something other than squeezenet would be cool). I can put the weights on GCP for you.

Overall this is amazing. Thank you for putting this together!

---

Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[scripts/build-caffe-proto.sh, line 16 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-C7wDSPiUoo1iw9qy:-Kz-C7wDSPiUoo1iw9qz:base36m) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/scripts/build-caffe-proto.sh#L16)):*
> ```Shell
> # limitations under the License.
> # =============================================================================
> set -e
> ```

can you put a big comment at the top here explaining what this is for

---

*[src/index.ts, line 57 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CB9p82Q-wggooozf:-Kz-CB9p82Q-wggooozg:b-rvyaix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/src/index.ts#L57)):*
> ```XML
> export {GPGPUContext} from './math/webgl/gpgpu_context';
> export {Model} from './model';
> export {CaffeModel} from './contrib/caffe_model';
> ```

this is what worries me since we'd be unconditionally compiling this into the bundle

---

*[src/contrib/caffe_model.ts, line 26 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CTJ-WMMBjY8TrHMg:-Kz-CTJ-WMMBjY8TrHMh:b4jgya9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/src/contrib/caffe_model.ts#L26)):*
> ```XML
> 
>   // TODO Map structure not compatible with object structure used in
>   // models.Squeezenet
> ```

what exactly isn't compatible?

---

*[src/contrib/caffe_model.ts, line 31 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CKDz3y04SAfsnIHV:-Kz-CKDz3y04SAfsnIHW:bvel09a) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/src/contrib/caffe_model.ts#L31)):*
> ```XML
>    * @type {Map<string, NDArray>}
>    */
>   protected variables: Map<string, NDArray>;
> ```

we should try to avoid map, you can just use a regular associative array:

`protected variables: {[varName: string]: NDArray};`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-:-Kz-9hFjmC6bi27VAe2_:bxa87da)*

nsthorat:
Follow up thought:

We should make a new directory in models for `Caffe` with its own package.json. Any caffe model will depend directly on that public API. This makes dependencies much saner and allows us to version the Caffe logic.

Does this make sense?

---

Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-:-Kz-F0OBKy0QJlz6QkGc:bffxpkv)*

nsthorat:
One more follow up thought:

I think we're going to separate our repository into several repositories under a deeplearnjs github organization. We were thinking all of your Caffe porting logic would live in its own deeplearn-caffe repo, with its own npm package that we can let you own.

Would that make sense?

---

Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-:-Kz-b7TTrjbHEe9k_Bjk:b-k2bu7l)*

chaosmail:
Hi @nsthorat 

Thanks for reviewing the code and for your suggestions.

I think putting this code in a separate module is a very good idea (e.g. similar to the model code), as it contains a lot of code that most users might not need (the proto compiled src/contrib/caffe/caffe.js file is gigantic). Then Caffe models can depend on this code and use it when needed - I like this a lot. I think, separating into different projects is definitely good idea in the long term. 

However, from my point of view I think it might be too early to separate code into different repos instead of a monorepo (with multiple modules). Just think about that this code and all models will strongly depend on a common model interface and a common build tool chain. I am pretty sure that both things will change quite a bit in the common weeks and synchronous changes throughout multiple repos are a nightmare.

Hence, my recommendation would be to put the code of this MR into a models/caffe directory and build a models/squeezenet_caffe that uses the caffe weights, but in the same repo. It can be separated out anytime once the interfaces are stable. What do you think about this? Of course if you want to split it to a separate repos I am perfectly fine ;)

One more thing that I am struggeling with. Could you point me into the right direction of naming conventions for layer weights when dumping the weights to files (`scripts/dump_checkpoints/tensorflow_checkpoint_dumper.py#var_name_to_filename`) which you also  use in model::variables? I am a bit lost especially for non Conv/FC layers. Thanks a lot!

Best,
Christoph

chaosmail:
Review status: 0 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[models/caffe/caffe_model.ts, line 26 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CTJ-WMMBjY8TrHMg:-KzKOdo4X34XSrQ2rulM:b79rqdz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/models/caffe/caffe_model.ts#L26)):*

what exactly isn't compatible?
</blockquote></details>

I used ES6 Maps, but replaces them with associative arrays

---

*[models/caffe/caffe_model.ts, line 31 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CKDz3y04SAfsnIHV:-KzKOcwWsbSNZgrPCyEX:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/models/caffe/caffe_model.ts#L31)):*

we should try to avoid map, you can just use a regular associative array:

`protected variables: {[varName: string]: NDArray};`
</blockquote></details>

Done.

---

*[models/caffe/scripts/compile-proto.sh, line 16 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-C7wDSPiUoo1iw9qy:-KzKMWOaw4l-81xX0e_s:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/models/caffe/scripts/compile-proto.sh#L16)):*

can you put a big comment at the top here explaining what this is for
</blockquote></details>

Done.

---

*[src/index.ts, line 57 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-Kz-CB9p82Q-wggooozf:-KzKMi-hANKi3H5XGb_J:bwdgksw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfab42e0eedea0844085fb7cc1a5cd9790d3f46e/src/index.ts#L57)):*

this is what worries me since we'd be unconditionally compiling this into the bundle
</blockquote></details>

Done, moved to models/caffe for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346)*

nsthorat:
We're actually going to be splitting the toolchain up, it's getting really out of hand to figure out where things intersect. It seems like most ML libraries are making a separate models repository, so I think we're going to follow suit. It will also logically separate them so when we try to build things we realize that we should try to test them with unit tests and not directly with demos.

For checkpoint dumping, we just replace any non-alphanumeric characters with "_". As long as you follow that you should be okay.

Quick question - is there an example of automatic model structure porting? Am I missing that somehow?

---

Reviewed 2 of 30 files at r3.
Review status: 1 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[models/caffe/caffe_model.ts, line 22 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-KzP1Yb3lPSZNlUVGyuZ:-KzP1Yb3lPSZNlUVGyu_:bghql71) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e568d61ef6b4b94d95c22dd04282ae5a1f2729fa/models/caffe/caffe_model.ts#L22)):*
> ```XML
> import {Model, NDArray} from 'deeplearn';
> 
> export class CaffeModel implements Model {
> ```

can this be abstract?

---

*[models/caffe/caffe_model.ts, line 26 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-KzP1RU4OD0Vdncp3NEg:-KzP1RU4OD0Vdncp3NEh:b-3w2hen) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e568d61ef6b4b94d95c22dd04282ae5a1f2729fa/models/caffe/caffe_model.ts#L26)):*
> ```XML
>   /**
>    * Model weights per layer
>    * @type {{[varName: string]: NDArray}}
> ```

no need for type

---

*[models/caffe/caffe_model.ts, line 32 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-KzP1SRSwhk6dtDQghcL:-KzP1SRSwhk6dtDQghcM:b-3w2hen) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e568d61ef6b4b94d95c22dd04282ae5a1f2729fa/models/caffe/caffe_model.ts#L32)):*
> ```XML
>   /**
>    * Preprocessing Offset
>    * @type {NDArray}
> ```

no need for type

---

*[models/squeezenet-caffe/demo.html, line 15 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-KzP1lBpbjjL-ASypSn2:-KzP1lBpbjjL-ASypSn3:b-8xkt8m) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e568d61ef6b4b94d95c22dd04282ae5a1f2729fa/models/squeezenet-caffe/demo.html#L15)):*
> ```HTML
>     resultElement.innerText = 'Loading squeezenet...';
> 
>     // If dl isn't loaded, wait 1 second.
> ```

you can remove this now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/346#-:-KzP3zKVPtgG6hrhqy1d:b-qjlqcg)*

chaosmail:
Hi guys, here is a short update on my (very good) progress:

* I added support for parsing prototxt models and protobinary files (e.g. mean files)
* I added Caffe model parsing to Deeplearn.js math operations (using the math module)
* I added a CPU implementation for LRN (which is needed for many models of Caffe Model Zoo)
* I added a few models to test (GoogLeNet, GoogLeNet with Batchnorm, AgeNet, GenderNet, VGG16 crashes the browser unfortunately)
* I moved a lot of functionality to separate reusable packages (https://www.npmjs.com/package/prototxt-parser, https://www.npmjs.com/package/dag-iterator, https://www.npmjs.com/package/caffe-proto)

I had/still have a lot of problems with the issue that Caffe computes the output dimensions of Conv and Pool layer (https://github.com/BVLC/caffe/issues/1318 and https://github.com/BVLC/caffe/issues/4252) differently to Deeplearn.js (almost in every model). This is something that we should fix somehow.

I actually believe, that the code of this branch should go into a separate repo, so I can faster fix some rough edges, use demos on gh-pages, include the caffe_models in the same repo etc. What do you think about this idea?

One more thought: I think, that we could easily add pure support to a lot of other DL frameworks (Caffe2, TF, Keras) in the same way this library does (like deeplearn-kears, deeplearn-tensorflow, etc.) - as many of them also use protobuf under the hood. I will put together a demo of Tensorflow once I am through with this commit.

Here are to samples to show you how cool and easy the caffe models can be used:

```typescript
// AgeNet from https://gist.github.com/GilLevi/c9e99062283c719c03de
// Weights are served locally, mean and model are served from Github
export class AgeNet extends CaffeModel {

  // download model from 
  // https://gist.github.com/GilLevi/c9e99062283c719c03de
  static caffemodel = 'age_net.caffemodel'; 
  static prototxt = 'GilLevi/c9e99062283c719c03de/raw/ddd9cf649c323041ee6e4731ff45636a09261597/deploy_age.prototxt';
  static meanfile = 'GilLevi/AgeGenderDeepLearning/master/models/mean.binaryproto';

  // Target labels
  static labels = ['(0, 2)','(4, 6)','(8, 12)','(15, 20)','(25, 32)','(38, 43)','(48, 53)','(60, 100)'];

  constructor(math: NDArrayMathGPU) {
    super(math, MODEL_DIR + AgeNet.caffemodel, GITHUB_GIST + AgeNet.prototxt, GITHUB_CDN + AgeNet.meanfile);
  }
}

// SqueezeNet from https://github.com/DeepScale/SqueezeNet/tree/master/SqueezeNet_v1.1
// Served entierly from Github
export class SqueezeNet extends CaffeModel {

  // model served entirely from github.com
  static caffemodel = 'DeepScale/SqueezeNet/master/SqueezeNet_v1.1/squeezenet_v1.1.caffemodel';
  static prototxt = 'DeepScale/SqueezeNet/master/SqueezeNet_v1.1/deploy.prototxt';

  // Target labels
  static labels = IMAGENET_CLASSES;

  constructor(math: NDArrayMathGPU) {
    super(math, GITHUB_CDN + SqueezeNet.caffemodel, GITHUB_CDN + SqueezeNet.prototxt);
  }
}
```

Cheers guys!

nsthorat:
Hey Christoph,

Thanks for the detailed notes. I just wanted to let you know we haven't forgotten about you, we've just been busy with NIPS / vacations / etc. Daniel and I are going to both be back in the office and Tuesday and we'll chat about this work in person since it's a large change. I'm so sorry about the delay, but I promise we'll sync soon!

chaosmail:
no worries, I assumed that you guys must be crazy busy catching up with everyone after NIPS; congrats btw. also from me for the best demo award ;) let me know when you are back and want to discuss current state and next steps for this feature!

chaosmail:
Closing this PR as discussed. PRs #514 and #517 add Caffe compatibility to Deeplearn.js core; this branch will be moved to a separate repo.

PR #358, opened 2017-11-16
Math.relu() GPU should use NaN check + precision fix for Math.asin().

nkreeger:
This fixes all unit-tests on my G-Linux thinkpad X1 carbon V3. Win64 is another topic and I'll tackle those in another CL when I'm on a different machine.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/358)

### DIFF
--- a/src/math/unaryop_test.ts
+++ b/src/math/unaryop_test.ts
@@ -591,7 +591,7 @@ import {Array1D, Array2D, Scalar} from './ndarray';
       for (let i = 0; i < a.size; i++) {
         expected[i] = Math.asin(values[i]);
       }
-      test_util.expectArraysClose(result.getValues(), expected, 1e-2);
+      test_util.expectArraysClose(result.getValues(), expected);
 
       a.dispose();
     });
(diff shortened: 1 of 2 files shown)


dsmilkov:
:lgtm_strong: This is awesome. Thanks for diving deep and fixing this. Now you know what we are dealing it trying to get WebGL consistency :)

---

---

*[src/math/unaryop_test.ts, line 594 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/358#-Kz9wSGWaZ6Dg7oyMjkV:-Kz9wSGWaZ6Dg7oyMjkW:b-dtlsag) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ba171dbff6a3bfb5408e473dd97c60d6cc8fcf9a/src/math/unaryop_test.ts#L594)):*
>  1e-2

let's remove this param altogether, which will result in using the global `test_util.TEST_EPSILON` (which is the same value)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/358#-:-Kz9weDT4n540b0Jj2Bf:bgrx9zc)*

PR #360, opened 2017-11-18
Improve build system (npm -> yarn, remove bower, add demos/tsconfig.json)

dsmilkov:
- Replace npm with yarn
- Remove bower as a dependency (yarn handles it)
- Add `tsconfig.json` in `demos/`, so demos compile separately
- Update base `tsconfig.json` to not compile demos (15% faster compile time)
- Make all `tsconfig.json` files extend from the base `tsconfig.json` to avoid repetition
- move `@types/polymer` and `tone-piano` deps to demos
- update typescript compiler from `2.5.3` to `2.6.1`

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/360)


nsthorat:
:lgtm_strong:

This is such a great cleanup. Few comments inline.

---

Reviewed 16 of 26 files at r1, 31 of 36 files at r2, 1 of 3 files at r3, 4 of 5 files at r4.

---

*[package.json, line 2 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKUWEnz7AxLwHztaYJ:-KzKUWEnz7AxLwHztaYK:b-oz7x58) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f9e626c4d32725735a86adee38f886c2b1970a3f/package.json#L2)):*
> ```JSON
> {
>   "name": "deeplearn",
> ```

what do you think about using a scoped package? We will have to create a "deeplearn" organization, and then deprecate the old package. We can do this in a follow up, just wanted to put these notes down.

https://docs.npmjs.com/misc/scope

Check out the first response here about deprecation (users can get a message when installing):

https://stackoverflow.com/questions/43007471/how-to-migrate-npm-package-to-an-organization-scope

---

*[tsconfig.json, line 10 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKWmnG7X3o9Oxnfw-Z:-KzKWmnG7X3o9Oxnfw-_:b-xltlcm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4cf065abd76671e935d88f851dd091ca60f91b8b/tsconfig.json#L10)):*
> ```JSON
>     "declaration": true,
>     "target": "es5",
>     "lib": ["es2015", "dom"],
> ```

I thought we wanted to target es6 as well? Does that happen somehow? 

---

*[models/knn_image_classifier/package.json, line 28 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKWypsHfPFOk6u97D0:-KzKWyptW5BoH482Z428:b8cxtfp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4cf065abd76671e935d88f851dd091ca60f91b8b/models/knn_image_classifier/package.json#L28)):*
> ```JSON
>     "prep": "yarn && mkdirp dist",
>     "build": "browserify --standalone knn_image_classifier knn_image_classifier.ts -p [tsify] -o dist/bundle.js",
>     "publish-npm": "tsc && yarn build && npm publish"
> ```

sourceMap --false for consistency, and in the other package.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-:-KzKX2IQjOIRR9xPJZEt:b-tzxlmr)*

dsmilkov:
Thanks!

---

Reviewed 26 of 26 files at r1, 35 of 36 files at r2, 3 of 3 files at r3.

---

*[package.json, line 2 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKUWEnz7AxLwHztaYJ:-KzLW18XhD4ugLndPzW2:b-neo8f9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f9e626c4d32725735a86adee38f886c2b1970a3f/package.json#L2)):*

what do you think about using a scoped package? We will have to create a "deeplearn" organization, and then deprecate the old package. We can do this in a follow up, just wanted to put these notes down.

https://docs.npmjs.com/misc/scope

Check out the first response here about deprecation (users can get a message when installing):

https://stackoverflow.com/questions/43007471/how-to-migrate-npm-package-to-an-organization-scope
</blockquote></details>

Yup, let's do scope package in another PR

---

*[tsconfig.json, line 10 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKWmnG7X3o9Oxnfw-Z:-KzLW4Ec3epx-xO90Jev:bwnqfil) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4cf065abd76671e935d88f851dd091ca60f91b8b/tsconfig.json#L10)):*

I thought we wanted to target es6 as well? Does that happen somehow? 
</blockquote></details>

we will produce a separate es6 build (maybe under dist/es6/...). We shouldn't change the current target because we will break existing users.

---

*[models/knn_image_classifier/package.json, line 28 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-KzKWypsHfPFOk6u97D0:-KzLWWFYIdCT7zv5HBiT:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4cf065abd76671e935d88f851dd091ca60f91b8b/models/knn_image_classifier/package.json#L28)):*

sourceMap --false for consistency, and in the other package.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360#-:-KzLWWhTOSLZrAi5algH:b-ucfknh)*

dsmilkov:
Reviewed 1 of 5 files at r4, 20 of 20 files at r5.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/360)*

PR #369, opened 2017-11-21
Remove unnecessary scope in activation functions

Lewuathe:
Since `math` functions are tracked by default, ops in activation functions are not needed to be wrapped by `scope`.

Follow-up of #299 

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/369)


nsthorat:

---

*[src/math/activation_functions.ts, line 37 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/369#-KzU1JGNm8rPXA-47L2q:-KzU1JGNm8rPXA-47L2r:bngmcs9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8eeefdbd9c8142c8f356d1f0ccbfce223954f128/src/math/activation_functions.ts#L37)):*
> ```XML
>   der<T extends NDArray>(math: NDArrayMath, x: T, y: T) {
>     return math.scope(() => {
>       const ySquared = math.elementWiseMul(y, y);
> ```

you still need a scope here to clean ySquared (and other functions where you do 2 things in the scope)

You don't need math.scope() with a single math function inside that returns because we only clean up intermediates that aren't the return value

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/369)*

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/369#-:-L-Y9tPJzlJLfmMbJvbH:b-jbsf8z)*

PR #371, opened 2017-11-21
Add types to NDArray rand(), randNormal(), and randUniform().

nkreeger:
Bunch of things going on here, with a couple of TODOs. I started to deprecate the randGauss() method in utils to ensure that the proper generated pair is returned in order. If we want to switch to another more complicated algorithm we can easily do it with the interface I introduced. 

- Tests to come for truncated rand normal. I need to research more for proper validation of truncated normality. I'll port that functionality to the new MPRandGauss implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/371)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: node_js
 node_js: "8"
-install:
-  - yarn prep
 script:
-  - yarn build
-  - yarn lint
+  - ./scripts/build_and_lint_all
   - if [ "$BROWSERSTACK_KEY" ]; then yarn test-travis; fi
--- a/demos/benchmarks/math-benchmark-run-groups.ts
+++ b/demos/benchmarks/math-benchmark-run-groups.ts
@@ -26,6 +26,7 @@ import {MatmulCPUBenchmark, MatmulGPUBenchmark} from './matmul_benchmarks';
 import {PoolBenchmarkParams, PoolCPUBenchmark, PoolGPUBenchmark} from './pool_benchmarks';
 // tslint:disable-next-line:max-line-length
 import {ReductionOpsCPUBenchmark, ReductionOpsGPUBenchmark} from './reduction_ops_benchmark';
+// tslint:disable-next-line:max-line-length
 import {UnaryOpsCPUBenchmark, UnaryOpsGPUBenchmark} from './unary_ops_benchmark';
 
 export function getRunGroups(): BenchmarkRunGroup[] {
--- a/demos/package.json
+++ b/demos/package.json
@@ -4,9 +4,6 @@
   "description": "deeplearn.js demos - not actually published",
   "private": true,
   "dependencies": {
-    "deeplearn": "~0.3.11",
-    "deeplearn-knn-image-classifier": "~0.1.2",
-    "deeplearn-squeezenet": "~0.1.2",
     "@bower_components/font-roboto": "PolymerElements/font-roboto#^1.0.1",
     "@bower_components/iron-a11y-announcer": "PolymerElements/iron-a11y-announcer#^1.0.0",
     "@bower_components/iron-a11y-keys-behavior": "PolymerElements/iron-a11y-keys-behavior#^1.0.0",
@@ -53,7 +50,11 @@
     "@bower_components/paper-tooltip": "PolymerElements/paper-tooltip#1.*.*",
     "@bower_components/polymer": "Polymer/polymer#1.*.*",
     "@bower_components/web-animations-js": "web-animations/web-animations-js#^2.2.0",
-    "@bower_components/webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.24"
+    "@bower_components/webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.24",
+    "deeplearn": "~0.3.11",
+    "deeplearn-knn-image-classifier": "~0.1.2",
+    "deeplearn-squeezenet": "~0.1.2",
+    "tslint": "~5.8.0"
   },
   "repository": {
     "type": "git",
@@ -70,7 +71,9 @@
     "watchify": "~3.9.0"
   },
   "scripts": {
-    "prep": "yarn"
+    "prep": "yarn",
+    "build": "tsc",
+    "lint": "tslint -p . -t verbose"
   },
   "license": "Apache-2.0",
   "engines": {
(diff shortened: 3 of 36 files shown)


dsmilkov:
:lgtm_strong: This is great! Tiny comments thus LGTM. Feel free to submit after.

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/util.ts, line 60 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzV5sxEOsr1TSt3k6Id:-KzV5sxEOsr1TSt3k6Ie:bv2x7j8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/util.ts#L60)):*
> ```XML
>  * @param stdDev The standard deviation. Default is 1.
>  */
> // TODO(kreeger): Deprecate this method for generation in rand.ts.
> ```

tip: you can mark it as deprecated using jsdoc, that is:
`/** @deprecated Please use this other method instead */`.
Code editors/linters can take advantage of that.

---

*[src/math/ndarray.ts, line 473 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzVkpP-C-z5dPrVMKIB:-KzVkpP-C-z5dPrVMKIC:b-wksdes) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/ndarray.ts#L473)):*
> ```XML
>       values = new Float32Array(size);
>     } else if (dtype === 'int32') {
>       values = new Int32Array(size);
> ```

for my own understanding: calling NDArray.randUnform with dtype of int32 will reach this state, which means uniformly sampled floats will get floored (due to Int32Array()), which is want we want in order to avoid skewing the distribution, correct?

I was looking here for uniform ints generator: https://stackoverflow.com/questions/26795084/generating-uniform-distribution-using-math-random

---

*[src/math/rand.ts, line 18 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzV6XwqxzC_SBAMjx8O:-KzV6XwqxzC_SBAMjx8P:bftgns7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/rand.ts#L18)):*
> import {RandNormalDataTypes} from './ndarray';

There a circular dep (ndarray -> rand -> ndarray): https://travis-ci.org/PAIR-code/deeplearnjs/builds/305389339#L517

You'd have to move `RandNormalDataTypes` to be in here.

---

*[src/math/rand.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzViniAG_wTdmgOKfED:-KzViniAG_wTdmgOKfEE:b-vhv55f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/rand.ts#L54)):*
potentially useful: I think truncated random is nothing fancy. According to the [TensorFlow doc](https://www.tensorflow.org/api_docs/python/tf/truncated_normal):

> values whose magnitude is more than 2 standard deviations from the mean are dropped and re-picked.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-:-KzVlIsg-DvvWNF9K0z2:b-9a0b2)*

nkreeger:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/util.ts, line 60 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzV5sxEOsr1TSt3k6Id:-KzWuZA-saWxWmLKDkfT:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/util.ts#L60)):*

tip: you can mark it as deprecated using jsdoc, that is:
`/** @deprecated Please use this other method instead */`.
Code editors/linters can take advantage of that.
</blockquote></details>

Done.

---

*[src/math/ndarray.ts, line 473 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzVkpP-C-z5dPrVMKIB:-KzWudkZmnqGNhQzbehq:b-ea3b4a) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/ndarray.ts#L473)):*

for my own understanding: calling NDArray.randUnform with dtype of int32 will reach this state, which means uniformly sampled floats will get floored (due to Int32Array()), which is want we want in order to avoid skewing the distribution, correct?

I was looking here for uniform ints generator: https://stackoverflow.com/questions/26795084/generating-uniform-distribution-using-math-random
</blockquote></details>

Pretty much - the array gets initialized here, but the value is set in the randFunction param. I use Math.random() instead of floor() because in testing I see better JB values from skewness and kurtosis values (closer to 0). This probably works better than the github example because we're generating gaussian values from the Marsaglia polar method.

---

*[src/math/rand.ts, line 18 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzV6XwqxzC_SBAMjx8O:-KzWvG-Bouc9PtRhl4F8:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/rand.ts#L18)):*

> import {RandNormalDataTypes} from './ndarray';

There a circular dep (ndarray -> rand -> ndarray): https://travis-ci.org/PAIR-code/deeplearnjs/builds/305389339#L517

You'd have to move `RandNormalDataTypes` to be in here.
</blockquote></details>

Done.

---

*[src/math/rand.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371#-KzViniAG_wTdmgOKfED:-KzWvHpB3p13DZY4kKAh:b-22sgtt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8a99c30407b82f292937931c932b4dc8e96bbb64/src/math/rand.ts#L54)):*

potentially useful: I think truncated random is nothing fancy. According to the [TensorFlow doc](https://www.tensorflow.org/api_docs/python/tf/truncated_normal):

> values whose magnitude is more than 2 standard deviations from the mean are dropped and re-picked.
</blockquote></details>

Thanks I'll look into that for the next PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/371)*

PR #373, opened 2017-11-21
Adding error handling if the image is unusable.

dcunited08:
I wrapped the processing in a try/catch so if an error occurs, the user is notified and can recover.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/373)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ npm-debug.log
 .DS_Store
 dist/
 .idea/
-demos/bower_components/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ npm-debug.log
 .DS_Store
 dist/
 .idea/
+demos/bower_components/
--- a/demos/imagenet/imagenet.ts
+++ b/demos/imagenet/imagenet.ts
@@ -152,27 +152,37 @@ export class ImagenetDemo extends ImagenetDemoPolymer {
     const isWebcam = this.selectedInputName === 'webcam';
 
     await this.math.scope(async (keep, track) => {
-      const image = track(Array3D.fromPixels(
-          isWebcam ? this.webcamVideoElement : this.staticImgElement));
 
-      const inferenceResult = await this.squeezeNet.predict(image);
-      const namedActivations = inferenceResult.namedActivations;
+      this.clearOutput();
 
-      this.layerNames = Object.keys(namedActivations);
+      try {
+        const image = track(Array3D.fromPixels(
+            isWebcam ? this.webcamVideoElement : this.staticImgElement));
 
-      const topClassesToProbability = await this.squeezeNet.getTopKClasses(
-          inferenceResult.logits, TOP_K_CLASSES);
+        const inferenceResult = await this.squeezeNet.predict(image);
+        const namedActivations = inferenceResult.namedActivations;
 
-      let count = 0;
-      for (const className in topClassesToProbability) {
-        if (!(className in topClassesToProbability)) {
-          continue;
+        this.layerNames = Object.keys(namedActivations);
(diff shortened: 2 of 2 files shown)


nsthorat:

---

*[.gitignore, line 11 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373#-Kzo10OpmwtWtAa8DgmZ:-Kzo10OpmwtWtAa8Dgm_:b-h86x9g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df0015498a6bc29bd59b2ee4937c79f27dd7ae10/.gitignore#L11)):*
> ```
> dist/
> .idea/
> demos/bower_components/
> ```

you shouldnt have to do this now that we use yarn (remove bower_components from your directory)

---

*[demos/imagenet/imagenet.ts, line 156 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373#-Kzo0fiweuSPg7wjaueD:-Kzo0fiweuSPg7wjaueE:b-2i0ub9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df0015498a6bc29bd59b2ee4937c79f27dd7ae10/demos/imagenet/imagenet.ts#L156)):*
> ```XML
>     await this.math.scope(async (keep, track) => {
> 
>       this.clearOutput();
> ```

I think the underlying problem is that we actually call animate() and then fromPixels() before the webcam is actually ready. Instead of catching the error, can we fix the root cause (not animating when the webcam / image element is not ready yet)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373)*

dcunited08:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[demos/imagenet/imagenet.ts, line 156 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373#-Kzo0fiweuSPg7wjaueD:-Kzy_Gai7gtA03tpGmki:bwi3qed) ([raw file](https://github.com/pair-code/deeplearnjs/blob/44867ae1d3f5872da0b01b9ea01d0369167e81c4/demos/imagenet/imagenet.ts#L156)):*

I think the underlying problem is that we actually call animate() and then fromPixels() before the webcam is actually ready. Instead of catching the error, can we fix the root cause (not animating when the webcam / image element is not ready yet)?
</blockquote></details>

The issues doesn't appear to be with the webcam but with the images. I noticed the beer image doesn't work no matter what you do, we are receiving an error: Error: The output # of rows (55.5) must be an integer. Change the stride and/or zero pad parameters.  I will look into that more.  The webcam is working fine from my testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373)*

dcunited08:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[.gitignore, line 11 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373#-Kzo10OpmwtWtAa8DgmZ:-Kzyb6ZYwEFVFo8QUMV0:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/44867ae1d3f5872da0b01b9ea01d0369167e81c4/.gitignore#L11)):*

you shouldnt have to do this now that we use yarn (remove bower_components from your directory)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/373)*

nsthorat:
Sorry to drop this - we fixed this in a PR that was related! Thank you for the contribution :)

PR #376, opened 2017-11-22
Migrate math backend kernels to a separate file and interface defining inputs / outputs.

nsthorat:
This is the first PR preparing us for eager mode.

This PR separates `NDArrayMath` from backends. Backends now need to implement `NDArrayMathBackend`, which contains a set of core kernels. The `NDArrayMath` layer is backend-agnostic - it does shape validation and can compute combinations of kernels.

This PR also:
* Wraps `NDArrayMath` - `NDArrayMathCPU` and `NDArrayMathGPU` are now thin wrappers that initialize NDArrayMath with different backends. The API remains the same.
* Separates the optimizer tests into their own files, away from `session_test.ts`.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/376)


dsmilkov:
:lgtm_strong: This is a great step towards eager! Nice work!!

---

Reviewed 2 of 9 files at r1, 119 of 120 files at r2.
Review status: 80 of 81 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/math/math.ts, line 23 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/376#-KzofGHhZUPy4t5WeEPs:-KzofGHhZUPy4t5WeEPt:butfl0l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e0d19512604fcec51b8d64bda4c89fed571128e7/src/math/math.ts#L23)):*
> ```XML
> import * as axis_util from './axis_util';
> import {MatrixOrientation, NDArrayMathBackend} from './backends/math_backend';
> import {NDArrayMathBackendCPU} from './backends/math_backend_cpu';
> ```

I think we should move the legacy `NDArrayMathGPU` and `NDArrayMathCPU` c-tors to their backends: `NDArrayMathGPU` goes to `math/backends/math_backend_webgl.ts`, and analogously for `NDArrayMathCPU`

Then math.ts won't import any backends directly other than the interface `NDArrayMathBackend`. This won't cause any cyclic deps since the CPU/GPU backend will depend on math, but math depends only on backend interface.

This allows us to ship only webgl backend, or only cpu backend later on, with treeshaking.

---

*[src/math/math.ts, line 44 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/376#-KzoeQxnuyWPn7qxRt-N:-KzoeQxnuyWPn7qxRt-O:b-nx0jo1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e0d19512604fcec51b8d64bda4c89fed571128e7/src/math/math.ts#L44)):*
> abstract
 
remove `abstract`. We should be able to create a global NDArrayMath without going through the legacy NDArrayMathGPU/NDArrayMathGPU

---

*[src/math/math.ts, line 2082 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/376#-KzoepV5GnI7KyAPLAFw:-KzoepV5GnI7KyAPLAFx:b-7yal20) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e0d19512604fcec51b8d64bda4c89fed571128e7/src/math/math.ts#L2082)):*
> ```XML
> }
> 
> export class NDArrayMathCPU extends NDArrayMath {
> ```

add jsdoc on top
`/** @deprecated Please use NDArrayMath */`
here and in `NDArrayMathGPU` below

---

*[src/math/backends/math_backend_cpu.ts, line 33 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/376#-Kzoft_4NzbrMZqO0djX:-Kzoft_4NzbrMZqO0djY:bgmeg9q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e0d19512604fcec51b8d64bda4c89fed571128e7/src/math/backends/math_backend_cpu.ts#L33)):*
> NDArrayMathBackendCPU
optional: Consider calling it just `NDArrayBackend/NDArrayBackendWebGL/NDArrayBackendCPU` (math is implicit) And rename the files to `backend_cpu.ts` and `backend_webgl.ts`. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/376#-:-Kzodvnh21j_p4xw-UKg:bfzxbmk)*

PR #377, opened 2017-11-23
Add selu math

Lewuathe:
We can use same scaling parameters with [the ones TensorFlow uses](https://github.com/tensorflow/tensorflow/blob/c2ce4f68c744e6d328746b144ff1fcf98ac99e6c/tensorflow/compiler/tf2xla/kernels/elu_op.cc#L72-L75), In short,`alpha` and `scaledAlpha` instead of `lambda` as in original paper.

```
alpha = 1.0507009873554804934193349852946
lambda = 1.0507009873554804934193349852946
scaledAlpha = alpha * lambda = 1.7580993408473768599402175208123
```

see: https://arxiv.org/abs/1706.02515

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/377)

### DIFF
--- a/src/math/math.ts
+++ b/src/math/math.ts
@@ -1245,6 +1245,15 @@ export abstract class NDArrayMath {
   protected abstract eluDerInternal<T extends NDArray>(ndarray: T): T;
 
   /**
+   * Computes scaled exponential linear element-wise.
+   * @hidden
+   */
+  selu<T extends NDArray>(ndarray: T): T {
+    return this.executeOp('selu', () => this.seluInternal(ndarray));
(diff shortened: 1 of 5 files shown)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/377#-:-Kzo01fMRg5IF4ZP9fPw:b-jbsf8z)*

PR #378, opened 2017-11-23
Add batching to `math.conv2d` and `{min,max,avg}pool2d`

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/378)

### DIFF
--- a/src/graph/ops/convolution.ts
+++ b/src/graph/ops/convolution.ts
@@ -76,13 +76,16 @@ export class Convolution2D extends Operation {
   backProp(
       math: NDArrayMath, inferenceArrays: TensorArrayMap,
       gradientArrays: SummedTensorArrayMap) {
-    const weights = inferenceArrays.get(this.wTensor) as Array4D;
+    const filter = inferenceArrays.get(this.wTensor) as Array4D;
     const x = inferenceArrays.get(this.xTensor) as Array3D;
     const dy = gradientArrays.get(this.yTensor) as Array3D;
 
     math.scope(() => {
-      const {dw, db, dx} =
-          math.conv2dBackProp(x, dy, weights, this.stride, this.zeroPad);
+      const dw =
+          math.conv2dDerFilter(x, dy, filter.shape, this.stride, this.zeroPad);
+      const db = math.conv2dDerBias(dy);
+      const dx =
+          math.conv2dDerInput(x.shape, dy, filter, this.stride, this.zeroPad);
       gradientArrays.add(this.wTensor, dw);
       gradientArrays.add(this.bTensor, db);
       gradientArrays.add(this.xTensor, dx);
--- a/src/math/conv2d_der_test.ts
+++ b/src/math/conv2d_der_test.ts
@@ -18,9 +18,9 @@
 import * as test_util from '../test_util';
 import {MathTests} from '../test_util';
 
-import {Array3D} from './ndarray';
+import {Array3D, Array4D} from './ndarray';
 
-// math.conv2dDerWeights
+// math.conv2dDerFilter
 {
   const tests: MathTests = it => {
     it('input=3x3x1,d2=1,f=2,s=1,p=0', math => {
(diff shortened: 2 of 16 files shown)


nsthorat:
:lgtm_strong:

---

Reviewed 16 of 16 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/math/conv_util.ts, line 55 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjwDkh0yp7wORL1Cei:-KzjwDkh0yp7wORL1Cej:b-jy7k0o) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/conv_util.ts#L55)):*
> ```XML
>     filterSize: [number, number]|number, strides: number|[number, number],
>     pad: 'same'|'valid'|number,
>     dataFormat: 'channelsFirst'|'channelsLast' = 'channelsLast'): Conv2DInfo {
> ```

can we use the same terminology as TF and move this out to a type somewhere: 'NHWC' | 'NCHW'?

https://www.tensorflow.org/api_docs/python/tf/nn/conv2d

---

*[src/math/math.ts, line 1434 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjuAgse4aw0ldlrL3o:-KzjuAgse4aw0ldlrL3p:bo6omv1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/math.ts#L1434)):*
> ```XML
>   conv2d(
>       input: Array3D|Array4D, filter: Array4D, bias: Array1D|null,
>       strides: [number, number]|number, pad: 'valid'|'same'|number): Array3D
> ```

Can you make input a generic so that you statically pipe the input rank through to the result, and elsewhere?

---

*[src/math/math.ts, line 1697 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjwnrleKtYb9F8QoR4:-KzjwnrleKtYb9F8QoR5:blg1b7x) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/math.ts#L1697)):*
> ```XML
> 
>     const convInfo = conv_util.computeConv2DInfo(
>         input4D.shape, filter.shape, strides, pad, true);
> ```

name this "true" as a variable

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-:-KzjxrsUEzqPoMJoG7Ec:b-jbsf8z)*

dsmilkov:
Review status: 15 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

*[src/math/conv_util.ts, line 55 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjwDkh0yp7wORL1Cei:-KzkJqplLseQgj2Xh4Gg:bw1vrk8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/conv_util.ts#L55)):*

can we use the same terminology as TF and move this out to a type somewhere: 'NHWC' | 'NCHW'?

https://www.tensorflow.org/api_docs/python/tf/nn/conv2d
</blockquote></details>

I thought about this for a while. tf.layers decided to use `channelsFirst`|`channelsLast`, to be compatible to Keras and other APIs. I like `channelsFirst`|`channelsLast` since it's more readable and user friendly. I think it will not cause problems during migration. People that explicitly setup their models with `NCHW` will know it's `channelsFirst`.

Also I haven't yet exposed the dataFormat at the math API yet, so not visible to users yet. When we do that, we can, for sake of compatibility, take also `NCHW`|`NHWC` and map it to `channelsFirst`|`channelsLast`.

---

*[src/math/math.ts, line 1434 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjuAgse4aw0ldlrL3o:-Kzl3fRybqWKrvKC-3TT:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/math.ts#L1434)):*

Can you make input a generic so that you statically pipe the input rank through to the result, and elsewhere?
</blockquote></details>

Done.

---

*[src/math/math.ts, line 1697 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378#-KzjwnrleKtYb9F8QoR4:-KzkL1-_54xa0NXMzLm8:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8e5985bd8281ac18b018aaaf1602cb115f809f84/src/math/math.ts#L1697)):*

name this "true" as a variable
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/378)*

PR #381, opened 2017-11-24
Add type, support, and tests for NDArray.randTruncatedNorm()

nkreeger:
This gets the implementation very close. I think seeding math.random() might get us a little closer with precision. Means and Std deviation shift quite a bit when truncating using the Marsaglia polar method. I think the next PR I'll take a shot at implementing https://en.wikipedia.org/wiki/Ziggurat_algorithm.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/381)


nkreeger:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[a discussion](https://reviewable.io:443/reviews/pair-code/deeplearnjs/381#-KzxzDaLle9XAJdpGIBa:-KzxzDaLle9XAJdpGIBb:by5uhh4) (no related file):*
Friendly ping :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/381)*

dsmilkov:
:lgtm_strong: Great work! Left one FYI comment

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/math/ndarray_test.ts, line 1563 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/381#-Kzy0Vk3XFMgb4UC5z5a:-Kzy0Vk3XFMgb4UC5z5b:bx09yii) ([raw file](https://github.com/pair-code/deeplearnjs/blob/0ff8af5e9ef83b958c729906e81c263438001fbe/src/math/ndarray_test.ts#L1563)):*
> ```XML
>   // Expect higher variances for truncated values.
>   // TODO(kreeger): Consider a different gausiann method or seeding JS
>   // Math.rand() for precision.
> ```

fyi, plain javascript doesn't have seeded rand generator, but we have seedrandom as a dependency of the project, which is the go-to JS library for seeded rand generators, written by a Googler.

see: https://github.com/PAIR-code/deeplearnjs/blob/04fb4353854e6ab9a8f81b25036a819fa675d1ab/src/math/backends/backend_cpu.ts#L18

and https://github.com/davidbau/seedrandom

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/381#-:-Kzy1Jqwp-J7A_IprBgi:b-1df4fc)*

nkreeger:
Thanks - I was looking at seedrandom and forgot to check if the project is already using it. I'll create a followup PR that uses that and hopefully irons out any flaky normalization tests.

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/381#-:-KzyG1GgsmQLcnpIr55b:b-s742x2)*

PR #383, opened 2017-11-25
Simplify benchmarks & enable them on all devices.

nsthorat:
This PR:
* Enables benchmarks on all devices (whitelisted in support.js)
* Makes buttons mobile friendly with MDL.
* Adds viewport meta to the page so it scales for phones
* Moves common benchmark logic to `benchmark_util.warmupAndBenchmarkGPU`. This function warms up the GPU, times, and returns the result, and it changes functionality based on whether timers are enabled.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/383)


dsmilkov:
:lgtm_strong: Great cleanup!

---

Reviewed 12 of 13 files at r1.
Review status: 12 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[demos/benchmarks/benchmark.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/383#-Kzom1wMDH1nze56otwp:-Kzom1wNqz5_g89DJoh7:bsfqcit) ([raw file](https://github.com/pair-code/deeplearnjs/blob/5eb76978886b898c04c466d9b755ca3b7e04a3f7/demos/benchmarks/benchmark.ts#L19)):*
> 5000
i'd lower this to 2500 given that matmul tests are N^3. If we allow the last test to be 4800 for example, the next one will be at least double. 

also maybe rename to `LAST_RUN_CPU_CUTOFF_MS`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/383#-:-Kzomyu9cJtqVwoMOLeM:bq6sxv)*

PR #395, opened 2017-11-28
Add extra precision to NaN checks in Shader compiler.

nkreeger:
This additional multiplication step ensures that GPUs don't take optimizations for close-to-zero values (typically NaN on win64). This makes all tests pass on my win64 machine (AMD-FX-8350 eight-core @ 4.00ghz + NVIDIA GeForce GTX 1050 4GB)

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/395)

### DIFF
--- a/src/math/backends/webgl/shader_compiler.ts
+++ b/src/math/backends/webgl/shader_compiler.ts
@@ -238,11 +238,15 @@ const SHADER_PREFIX = `
   const vec2 halfCR = vec2(0.5, 0.5);
 
   bool isNaN(float val) {
-    return val == val ? false : true;
+    float v1 = val * val;
+    float v2 = val * val;
+    return v1 == v2 ? false : true;
   }
 
   bool hasNaN(vec4 values) {
-    return any(notEqual(values, values));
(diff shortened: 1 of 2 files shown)


dsmilkov:
:lgtm_strong: Oh boy, that's just sad, that we have to do these things.

Do you have any sense on the performance impact of this change for shaders where we check for NaNs?

We should think about adding production mode where we strip out the NaN checking stuff and make the shaders leaner.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/395#-:-L-4hY0DylatjmpH9XIe:b-l8eic4)*

nkreeger:
I don't have great visibility into performance impact yet . We might be able to add a NaN sniffer before compiling the shader source - or when webgl3 is here rely on isnan().

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/395#-:-L-4iJwvnnF-eH49-n61:b-glpf6l)*

PR #398, opened 2017-11-29
Correct backprop for add operation with broadcast

minsukkahng:
Hi Daniel and Nikhil,

Although I'm not 100% sure, I think when we compute backprop for add involving broadcast, we need to use sum, not average. Could you check if my PL is correct? I might miss something.

I identified this issue while I perform 2d+1d in my tool. When I update code to this PL, it works better. I have also tried to derive this backpropagation equation. I hope you can double-check with this.

To be short, I want you to check whether the answer for the following line is 7 (avg) or 42 (sum).
https://github.com/minsukkahng/deeplearnjs/blob/71e7ed105591f958a642313d9590f5cd3142c8df/src/graph/ops/add_test.ts#L146

Thanks,
Brian

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/398)

### DIFF
--- a/src/graph/ops/add.ts
+++ b/src/graph/ops/add.ts
@@ -76,18 +76,10 @@ export class Add extends Operation {
             this.x2Tensor.shape.length === 2 &&
             this.x1Tensor.shape[0] === this.x2Tensor.shape[1]) {
           const sum = math.sum(dy as Array2D, 0);
-          if (this.dySizeScalar == null) {
-            this.dySizeScalar = Scalar.new(this.x2Tensor.shape[0]);
-          }
-          gradientArrays.add(
-              this.x1Tensor, math.divide(sum, this.dySizeScalar));
+          gradientArrays.add(this.x1Tensor, sum);
         } else if (util.isScalarShape(this.x1Tensor.shape)) {
           const sum = math.sum(dy);
-          if (this.dySizeScalar == null) {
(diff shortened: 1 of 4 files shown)


nsthorat:
You are totally right, I just did the math out too.

Could you also make the change here:
https://github.com/PAIR-code/deeplearnjs/blob/master/src/graph/ops/subtract.ts

Then LGTM, sorry about that, and nice find!

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/398#-:-L-CpqT-aAl6WvkSfxk5:b236mqi)*

minsukkahng:
Great! Thanks for the check. My original commit also includes correcting subtract (for scalar - ndarray). I can add a new case of 2darray - 1darray if you like.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/398#-:-L-DffyWhSuYzv2xIiAU:b-yrof33)*

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/398#-:-L-I0nb7zaMZLRWQU_3l:b-jbsf8z)*

PR #406, opened 2017-12-03
Update SqueezeNet and KNN Image Classifier API to not do async predictions.

nsthorat:
This PR:
* Updates SqueezeNet predict function to return the logits directly.
* Adds a separate method to SqueezeNet if you want a particular activation. This makes "predict" a cleaner API.
* Updates KNN Image Classifier model to only asynchronously convert predictions to the class / probabilities. Calling `addImage` is immediate now, no need to await.
* Fixes the webcam imagenet demo. We now wait for image / video element to be ready before inferring. This ensures we don't throw errors trying to upload a [0x0] texture to the GPU.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/406)


dsmilkov:
:lgtm_strong:

---

Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions.

---

*[demos/imagenet/imagenet.ts, line 117 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RlpdyWmv_LTCFNAD-:-L-RlpdyWmv_LTCFNAD0:b-9xwwqt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L117)):*
> ```XML
>     } else {
>       this.initWithoutWebcam();
>     }
> ```

How about wrapping this whole block (line 100-117) into `setupCameraInput()` which returns a Promise when it decided that it has/doesn't have webcam.

Squeezenet.load() is another promise and you want to execute them in parallel.

```
const camSetup = setupCameraInput();
const modelSetup = this.squeezeNet.load();
await Promise.all([camSetup, modelSetup]);
// Now you know both uiInitialized and squeezeNetLoaded are true.
```

---

*[demos/imagenet/imagenet.ts, line 126 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-Rk73KdOTa_c0xTrJo:-L-Rk73KdOTa_c0xTrJp:b6p9bos) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L126)):*
> requestAnimationFrame(() => this.animate());

maybe cleaner to call `this.startAnimating();`

---

*[demos/imagenet/imagenet.ts, line 139 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RkfkDSEF_3ELY-gT0:-L-RkfkDSEF_3ELY-gT1:b-5ptzi3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L139)):*
> setUiInitialized

rename to `setUIInitializedAndStartAnimating` to be clearer

---

*[demos/imagenet/imagenet.ts, line 192 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RhlKDegnKQpCwBA1z:-L-RhlKEThPAD5ZwKW8t:bze02ex) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L192)):*
> ```XML
>       }
> 
>       const image = track(Array3D.fromPixels(element));
> ```

Wouldn't it be amazing if the ready logic above gets incorporated into our API, so it's not on the user to do different checks for VideoElement vs ImgElement.

I'm not sure what the best API is, maybe something like:

```
await dl.pixelsReady(element);
imArray3D.fromPixels(element);
```

and ideally you want to do this `awaiting` before calling `startAnimating()`, though not sure if that complicates logic when user switches between webcam and image input.

---

*[models/knn_image_classifier/knn_image_classifier.ts, line 122 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RpTHnhQb5RLF5t3kX:-L-RpTHnhQb5RLF5t3kY:bc2yoo5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/models/knn_image_classifier/knn_image_classifier.ts#L122)):*
> ```XML
>   }
> 
>   predict(image: Array3D): Array1D {
> ```

Add jsdoc explaining what the result is. In this case, if I understand correctly, it's the cosine distance between the image and all images in the database.

If true, then this method as a standalone might not be very useful to the user. Because to use the prediction, you need to also get the database and also the `classIndex` associated with every row in the database, at which point you basically need predictClass.

so I would rename `predictClass -> predict` and `predict -> computeLogits` and maybe keep `computeLogits` internal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-:-L-RskgNt6Fy1WpU3W-3:b-jbsf8z)*

nsthorat:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

*[demos/imagenet/imagenet.ts, line 117 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RlpdyWmv_LTCFNAD-:-L-SKnCaUxICe-rO76ds:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L117)):*

How about wrapping this whole block (line 100-117) into `setupCameraInput()` which returns a Promise when it decided that it has/doesn't have webcam.

Squeezenet.load() is another promise and you want to execute them in parallel.

```
const camSetup = setupCameraInput();
const modelSetup = this.squeezeNet.load();
await Promise.all([camSetup, modelSetup]);
// Now you know both uiInitialized and squeezeNetLoaded are true.
```

</blockquote></details>

Done.

---

*[demos/imagenet/imagenet.ts, line 126 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-Rk73KdOTa_c0xTrJo:-L-SKywsQFFoIbT2R_LZ:b-t9w44o) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L126)):*

> requestAnimationFrame(() => this.animate());

maybe cleaner to call `this.startAnimating();`
</blockquote></details>

Removed all the other places `requestAnimationFrame` can be called from, so we just start the loop here.

---

*[demos/imagenet/imagenet.ts, line 139 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RkfkDSEF_3ELY-gT0:-L-SL3ws8nP7_95gi_uE:b-ebhsct) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L139)):*

> setUiInitialized

rename to `setUIInitializedAndStartAnimating` to be clearer
</blockquote></details>

Removed this entirely.

---

*[demos/imagenet/imagenet.ts, line 192 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RhlKDegnKQpCwBA1z:-L-SL5fz3GEE__vkAvvg:bcrnzp8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/demos/imagenet/imagenet.ts#L192)):*

Wouldn't it be amazing if the ready logic above gets incorporated into our API, so it's not on the user to do different checks for VideoElement vs ImgElement.

I'm not sure what the best API is, maybe something like:

```
await dl.pixelsReady(element);
imArray3D.fromPixels(element);
```

and ideally you want to do this `awaiting` before calling `startAnimating()`, though not sure if that complicates logic when user switches between webcam and image input.


</blockquote></details>

Possibly, but I think maybe we don't want to get in the business of UI-related stuff. 

Actually, I should have just added load events before calling this (this was only a little weird because of the animate loop).

What we could do, though, is simply throw if the image isn't ready inside `fromPixels`. I can do that in a follow up.

---

*[models/knn_image_classifier/knn_image_classifier.ts, line 122 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406#-L-RpTHnhQb5RLF5t3kX:-L-SLLNRv9WY2qaT6dik:b-2a9fhd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b4590883fb2b44be96caff11e2c7b70a403ee68d/models/knn_image_classifier/knn_image_classifier.ts#L122)):*

Add jsdoc explaining what the result is. In this case, if I understand correctly, it's the cosine distance between the image and all images in the database.

If true, then this method as a standalone might not be very useful to the user. Because to use the prediction, you need to also get the database and also the `classIndex` associated with every row in the database, at which point you basically need predictClass.

so I would rename `predictClass -> predict` and `predict -> computeLogits` and maybe keep `computeLogits` internal.
</blockquote></details>

Spoke offline - because of the CPU indices computations this would end up returning a promise. For now, we decided to keep this as is. I put a big comment block explaining what is going on and I will file 2 issues for the 2 kernels we need.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/406)*

PR #412, opened 2017-12-04
Added cholesky

viktorpergjoka:
Added cholesky decomposition with additional functions in new linalg file

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/412)


dsmilkov:
First thank you for your contribution!

Since we want all operations to be fast and utilize the GPU via WebGL, we won't be accepting this since it is using plain javascript.

PR #425, opened 2017-12-08
Add the pow operator

pyu10055:
Added pow operator for both cpu and gpu
discovered some discrepancies between WebGL implementation of pow function.
the current implementation is when the base is negative or base is 0 and power is negative then return NaN. That is the current OpenGL [spec](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/pow.xhtml).

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/425)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,8 @@ $ yarn prep # Installs dependencies.
 ```
 
 We recommend using [Visual Studio Code](https://code.visualstudio.com/) for
-development. Make sure to install [TSLint VSCode extension](https://marketplace.visualstudio.com/items?itemName=eg2.tslint) and the `clang-format` command line tool with the [Clang-Format VSCode extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) for auto-formatting.
+development. Make sure to install [TSLint VSCode extension](https://marketplace.visualstudio.com/items?itemName=eg2.tslint) and the `clang-format` command line tool with the [Clang-Format VSCode extension](https://marketplace.visualstudio.com/items?itemName=xaver.clang-format) for auto-formatting (Be sure to use the [NPM/Yarn clang-format binary](https://github.com/angular/clang-format)).
+
 
 To interactively develop any of the demos (e.g. `demos/nn-art/`):
 
--- a/README.md
+++ b/README.md
@@ -48,9 +48,11 @@ console.log(result.dataSync());
 #### ES3/ES5 JavaScript
 
 You can also use **deeplearn.js** with plain JavaScript. Load the latest version
-of the library from [unpkg](https://unpkg.com):
+of the library from [jsDelivr](https://www.jsdelivr.com/) or [unpkg](https://unpkg.com):
 
 ```html
+<script src="https://cdn.jsdelivr.net/npm/deeplearn"></script>
+<!-- or -->
 <script src="https://unpkg.com/deeplearn"></script>
 ```
 
--- a/demos/package.json
+++ b/demos/package.json
@@ -66,6 +66,7 @@
     "d3-array": "~1.2.1",
     "d3-format": "~1.2.1",
     "d3-scale": "~1.0.7",
+    "minimist": "~1.2.0",
     "mkdirp": "~0.5.1",
     "poi": "~9.5.4",
     "poi-preset-typescript": "~9.0.2",
--- a/demos/performance_rnn/index.html
+++ b/demos/performance_rnn/index.html
@@ -202,6 +202,12 @@ limitations under the License.
       bottom: 0;
       background-color: green;
     }
+    #resettingText {
+      font-size: 20px;
+      transition: opacity 1s;
+      text-align: center;
+      margin-top: 15px;
+    }
   </style>
 </head>
 <body>
@@ -342,6 +348,9 @@ limitations under the License.
       </div>
     </div>
     <div id="keyboard" class="hidden"></div>
+    <div id="resettingText" style="opacity: 0">
+      Resetting...
+    </div>
     <div class="text-block">
       <a href="https://magenta.tensorflow.org/performance-rnn">Performance RNN</a> was trained in TensorFlow on MIDI from piano performances. It was then ported to run in the browser using only Javascript in the <a href="https://deeplearnjs.org/">deeplearn.js</a> environment. Piano samples are from <a href="https://archive.org/details/SalamanderGrandPianoV3">Salamander Grand Piano</a>.
     </div>
--- a/demos/performance_rnn/performance_rnn.ts
+++ b/demos/performance_rnn/performance_rnn.ts
@@ -52,6 +52,8 @@ const NOTES_PER_OCTAVE = 12;
 const DENSITY_BIN_RANGES = [1.0, 2.0, 4.0, 8.0, 16.0, 32.0, 64.0];
 const PITCH_HISTOGRAM_SIZE = NOTES_PER_OCTAVE;
 
+const RESET_RNN_FREQUENCY_MS = 30000;
+
 let pitchHistogramEncoding: Array1D;
 let noteDensityEncoding: Array1D;
 let conditioned = false;
@@ -109,7 +111,7 @@ const SALAMANDER_URL = 'https://storage.googleapis.com/learnjs-data/' +
 const CHECKPOINT_URL = 'https://storage.googleapis.com/learnjs-data/' +
     'checkpoint_zoo/performance_rnn_v2';
 
-const isDeviceSupported = demo_util.isWebGLSupported() && !demo_util.isSafari();
+const isDeviceSupported = demo_util.isWebGLSupported();
 
 if (!isDeviceSupported) {
   document.querySelector('#status').innerHTML =
@@ -120,6 +122,8 @@ if (!isDeviceSupported) {
 }
 
 const math = new NDArrayMathGPU();
+
+let modelReady = false;
 
 function start() {
   piano.load(SALAMANDER_URL)
@@ -149,6 +153,7 @@ function start() {
 
         fullyConnectedBiases = vars['fully_connected/biases'] as Array1D;
         fullyConnectedWeights = vars['fully_connected/weights'] as Array2D;
+        modelReady = true;
         resetRnn();
       });
 }
@@ -594,6 +599,7 @@ function midiInNoteOn(midiNote: number, velocity: number) {
 
   // Turn on conditioning when a note is pressed/
   if (!conditioned) {
+    resetRnn();
     enableConditioning();
   }
 
@@ -601,6 +607,7 @@ function midiInNoteOn(midiNote: number, velocity: number) {
   setTimeout(() => {
     if (performance.now() - lastNotePressedTime > CONDITIONING_OFF_TIME_MS) {
       disableConditioning();
+      resetRnn();
     }
   }, CONDITIONING_OFF_TIME_MS);
 
@@ -637,7 +644,10 @@ function playOutput(index: number) {
         if (activeMidiOutputDevice != null) {
           try {
             activeMidiOutputDevice.send(
-                [MIDI_EVENT_ON, noteNum, currentVelocity * globalGain],
+                [
+                  MIDI_EVENT_ON, noteNum,
+                  Math.min(Math.floor(currentVelocity * globalGain), 127)
+                ],
                 Math.floor(1000 * currentPianoTimeSec) - pianoStartTimestampMs);
           } catch (e) {
             console.log(
@@ -662,7 +672,10 @@ function playOutput(index: number) {
 
         if (activeMidiOutputDevice != null) {
           activeMidiOutputDevice.send(
-              [MIDI_EVENT_OFF, noteNum, currentVelocity * globalGain],
+              [
+                MIDI_EVENT_OFF, noteNum,
+                Math.min(Math.floor(currentVelocity * globalGain), 127)
+              ],
               Math.floor(timeSec * 1000) - pianoStartTimestampMs);
         }
         piano.keyUp(noteNum, timeSec);
@@ -680,8 +693,10 @@ function playOutput(index: number) {
                 `seconds which is over ${MAX_NOTE_DURATION_SECONDS}, will ` +
                 `release.`);
             if (activeMidiOutputDevice != null) {
-              activeMidiOutputDevice.send(
-                  [MIDI_EVENT_OFF, noteNum, currentVelocity * globalGain]);
+              activeMidiOutputDevice.send([
+                MIDI_EVENT_OFF, noteNum,
+                Math.min(Math.floor(currentVelocity * globalGain), 127)
+              ]);
             }
             piano.keyUp(noteNum, currentPianoTimeSec);
             activeNotes.delete(noteNum);
@@ -700,3 +715,19 @@ function playOutput(index: number) {
   }
   throw new Error(`Could not decode index: ${index}`);
 }
+
+// Reset the RNN repeatedly so it doesn't trail off into incoherent musical
+// babble.
+const resettingText = document.getElementById('resettingText');
+function resetRnnRepeatedly() {
+  if (modelReady) {
+    resetRnn();
+    resettingText.style.opacity = '100';
+  }
+
+  setTimeout(() => {
+    resettingText.style.opacity = '0';
+  }, 1000);
+  setTimeout(resetRnnRepeatedly, RESET_RNN_FREQUENCY_MS);
+}
+setTimeout(resetRnnRepeatedly, RESET_RNN_FREQUENCY_MS);
--- a/demos/tsconfig.json
+++ b/demos/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../tsconfig.json",
   "include": ["**/*"],
   "compilerOptions": {
-    "declaration": false
+    "declaration": false,
+    "target": "es2017"
   }
 }
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Hardware-accelerated JavaScript library for machine intelligence",
   "private": false,
   "main": "dist/index.js",
+  "jsdelivr": "dist/deeplearn.js",
   "unpkg": "dist/deeplearn.js",
   "types": "dist/index.d.ts",
   "repository": {
--- a/scripts/watch-demo
+++ b/scripts/watch-demo
@@ -18,11 +18,13 @@ const path = require('path');
 const spawn = require('cross-spawn');
 const opn = require('opn');
 const shell = require('shelljs');
+var argv = require('minimist')(process.argv.slice(2));
 
 /** List of vue demos. */
 const VUE_DEMOS = new Set(['vue-demo', 'latent-space-explorer']);
 
-const demoPath = process.argv[2];
+const demoPath = argv._[0];
+
 const demoFolder = path.basename(demoPath);
 const outputPath = path.join(demoPath, 'bundle.js')
 const tsPath = path.join(demoPath, demoFolder + '.ts');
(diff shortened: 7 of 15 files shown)


nkreeger:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/math/element_wise_arithmetic_test.ts, line 259 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-rngKqZ8h5vyNJkyRZ:-L-rngKqZ8h5vyNJkyR_:b-eewjsa) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/element_wise_arithmetic_test.ts#L259)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>
>     it('pow throws when passed ndarrays of different shapes',
>         math => {
>           const a = Array2D.new([2, 3], [1, 2, -3, -4, 5, 6]);
>           const b = Array2D.new([2, 2], [5, 3, 4, -7]);
</details>
>  
>           expect(() => math.pow(a, b)).toThrowError();
>           expect(() => math.pow(b, a)).toThrowError();
>  
>           a.dispose();
>           b.dispose();
>         });
Looks like the spacing is off in this CL. I had some trouble with the correct clang-format - see the README for instructions on how to use it. Are you using VS Code w/ the extensions supplied there? It makes your life much easier :-)

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-ro6NEaucoH2djypgr:b-k54cyg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*
> ```XML
> // ensure the pow op return NaN for all negative base cases.
> export const POW = `
>   if (a < 0.0 ) return pow(a, 0.5);
> ```

You'll want to use the isNaN() method here - look at EQUAL and others.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/math/element_wise_arithmetic_test.ts, line 259 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-rngKqZ8h5vyNJkyRZ:-L-sGTLzoZGYUS5xLYGn:baw72ga) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/element_wise_arithmetic_test.ts#L259)):*
<details><summary><i>Previously, nkreeger (Nick Kreeger) wrote…</i></summary><blockquote>

>     it('pow throws when passed ndarrays of different shapes',
>         math => {
>           const a = Array2D.new([2, 3], [1, 2, -3, -4, 5, 6]);
>           const b = Array2D.new([2, 2], [5, 3, 4, -7]);
>  
>           expect(() => math.pow(a, b)).toThrowError();
>           expect(() => math.pow(b, a)).toThrowError();
>  
>           a.dispose();
>           b.dispose();
>         });
Looks like the spacing is off in this CL. I had some trouble with the correct clang-format - see the README for instructions on how to use it. Are you using VS Code w/ the extensions supplied there? It makes your life much easier :-)
</blockquote></details>

thanks, done.

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-sGDcnQg7pKcDeEqXe:bjt4c4a) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*
<details><summary><i>Previously, nkreeger (Nick Kreeger) wrote…</i></summary><blockquote>

You'll want to use the isNaN() method here - look at EQUAL and others.
</blockquote></details>

my goal is to return an NaN value if the base is less than 0. But seems there is no way
to do that directly, I was trying to use the pow's own implementation.
any suggestions on how to do that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 1005 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sLu4zmKEXXxO84o8j:-L-sLu4zmKEXXxO84o8k:b-m4cuvz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/23492c0c1ca4fc9a0bac5ce5c3bfeda1e7e39265/src/math/math.ts#L1005)):*
> ```XML
>    * @param b The second NDArray to pow element-wise.
>    */
>   pow<G extends keyof DataTypes>(a: NDArray<G>, b: NDArray<G>):
> ```

can you also add a `powStrict`? `powStrict` enforces that both NDArrays are the same shape so we can statically infer the return rank is the same as the input type (see `subStrict`)

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-sO_rxPgPXhMvtNK9c:bpobcqj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

my goal is to return an NaN value if the base is less than 0. But seems there is no way
to do that directly, I was trying to use the pow's own implementation.
any suggestions on how to do that?
</blockquote></details>

Even though the GLSL spec has bases < 0 returning NaN, we definitely want to support negative bases. You can do this by saving the sign of a, taking absolute value of a, taking the pow of that result, and then negating the value of the result depending on whether the sign of a is negative and the power is odd.

Separately, if all the NaN tests fail for pow, you should have an explicit isNaN(a) check at the top of this method.

One more thing: can you make sure to add {}

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-sUT5gsyxmyrmgs5-R:b-myvny7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*

Even though the GLSL spec has bases < 0 returning NaN, we definitely want to support negative bases. You can do this by saving the sign of a, taking absolute value of a, taking the pow of that result, and then negating the value of the result depending on whether the sign of a is negative and the power is odd.

Separately, if all the NaN tests fail for pow, you should have an explicit isNaN(a) check at the top of this method.

One more thing: can you make sure to add {}
</blockquote></details>

My understanding is,with float point, it does not make sense to check odd/even. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-sX8s2HOcVM5rM4mxw:bamjtth) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

My understanding is,with float point, it does not make sense to check odd/even. 
</blockquote></details>

I'm not sure I understand.

Thinking this through, I think we want to force 'int32' as the dtype on y. If it's floating point, and x is negative, we'll get complex numbers, and I think negative base and integer power is much more common of a use case.

So what you should do is round b, and use the approach from above (make sure you support negative numbers).

Does this make sense?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 1005 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sLu4zmKEXXxO84o8j:-L-spdHzcmRM9y_Np_KC:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/23492c0c1ca4fc9a0bac5ce5c3bfeda1e7e39265/src/math/math.ts#L1005)):*

can you also add a `powStrict`? `powStrict` enforces that both NDArrays are the same shape so we can statically infer the return rank is the same as the input type (see `subStrict`)
</blockquote></details>

Done.

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L-spSMEW-2vgiZxG_Hh:bovnj5s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*

I'm not sure I understand.

Thinking this through, I think we want to force 'int32' as the dtype on y. If it's floating point, and x is negative, we'll get complex numbers, and I think negative base and integer power is much more common of a use case.

So what you should do is round b, and use the approach from above (make sure you support negative numbers).

Does this make sense?
</blockquote></details>

make sense, I have the same concern, but make the second param int32 strictly could solved this problem. thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
LGTM after this!

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L-sqdXUirG4PC1Ws8fI:b-lhodm4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*
> ```XML
> }
> 
> // element-wise pow
> ```

can you add a test to make sure we throw when we're not calling pow with an int32 y?

---

*[src/math/element_wise_arithmetic_test.ts, line 279 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-stWUeVLnL8VDvVaK-:-L-stWUeVLnL8VDvVaK0:be58j1k) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L279)):*
> ```XML
> 
>       a.dispose();
>       b.dispose();
> ```

can you add a basic powStrict test?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-:-L-st_wmkifhuoYf73Gt:b-5earjc)*

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L01hmA8ailIGKenXSdc:b8b7vu9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*

can you add a test to make sure we throw when we're not calling pow with an int32 y?
</blockquote></details>

would not that be guaranteed by the compiler type check?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L01kSdC-DwSPyxxNGqz:bxm4rp5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

would not that be guaranteed by the compiler type check?
</blockquote></details>

In TypeScript, yes, but most of our users are in JS.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L023hW8FaoTRS7xTug-:berr55u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*

In TypeScript, yes, but most of our users are in JS.
</blockquote></details>

make sense, but I am not sure how to test this check with unit test file in ts. 

---

*[src/math/element_wise_arithmetic_test.ts, line 279 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-stWUeVLnL8VDvVaK-:-L024ZSdhHHIRKRGFfak:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L279)):*

can you add a basic powStrict test?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L02l-yOP7GnPxKxO0xZ:b-r7b3l2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

make sense, but I am not sure how to test this check with unit test file in ts. 
</blockquote></details>

Here's an example of testing for 'bool':
https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/tile_test.ts#L89

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L02ltqb2i8zfvfIHWoH:b-4dk7xz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

make sense, I have the same concern, but make the second param int32 strictly could solved this problem. thanks.
</blockquote></details>

quick question: how would this ever be 2? Because of precision issues?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
By the way, not sure if you saw "Reviewable" for code reviews yet, but if you're not using it check out the link in your PR description (github reviews aren't so great).

pyu10055:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L034o_jIK1HZydBVP8a:bcj7ubn) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*

Here's an example of testing for 'bool':
https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/tile_test.ts#L89
</blockquote></details>

my issue is how to test the negative case, meaning provide a float32 exponent, and expect the method to throw error.

---

*[src/math/backends/webgl/binaryop_gpu.ts, line 27 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-ro6NEaucoH2djypgq:-L035E0Yv3P16dMPQOg6:bnbm2f9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/feafcf72849c7a8c537e8dd71ced093d6713dbbb/src/math/backends/webgl/binaryop_gpu.ts#L27)):*

quick question: how would this ever be 2? Because of precision issues?
</blockquote></details>

yes, if the precision causes the value to be slightly less than 2, the round function would make it 2.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/math/element_wise_arithmetic_test.ts, line 231 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-L-sqdXUirG4PC1Ws8fH:-L05KR7w09sSXD_MvX3c:b-yg9qsn) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25ce69e340cdfadd6dc762ea8e1be2eaf5d77f4f/src/math/element_wise_arithmetic_test.ts#L231)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

my issue is how to test the negative case, meaning provide a float32 exponent, and expect the method to throw error.
</blockquote></details>

Ah, so just cast the second param as any:
`const result = math.pow(a, b as any);`

Where b is 'float32'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425)*

nsthorat:
:lgtm_strong:

---

Reviewed 4 of 6 files at r3, 1 of 2 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/425#-:-L05xllowO0uk3nFJb9h:b-jbsf8z)*

PR #426, opened 2017-12-09
Add kernel input configurations as the interface between Math and Backends via a BackendEngine.

nsthorat:
* Adds kernel configurations which have: `Node`s, `InputConfig`s and `InputArray`s. `Node`s will represent entire executed Node in our kernel graph, `InputConfig`s will hold the args, and `InputArray`s will own the inputs themselves (so we can iterate through them during backprop). These are separated for type enforcement.
* Adds a `BackendEngine` which calls into a `kernel_registry.executeKernel` utility method. This method looks up the kernel by name in an object that maps it to its actual backend function. This allows us to have a tape as well as serialize the kernel graph.

Note:`BackendEngine` will eventually hold the tape.

Other things:
* Deletes copy2d. It's not used anywhere and it's not functional.
* Removes scaledArrayAdd kernel.
* Attempt to name kernels similar to TensorFlow (capitalized, short, etc)
* Makes executeOp a pass through. We are eventually going to make this log a combination of ops and other kernels, so we should keep this structure around.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/426)


dsmilkov:
:lgtm_strong: Awesome change! Left a few comments, but nothing major. The only bigger comment is to rename `backends/kernels` to `backends/types`

---

Reviewed 6 of 18 files at r1, 27 of 28 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/index.ts, line 51 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-L03GWPI5D_iy1gxd_zw:-L03GWPI5D_iy1gxd_zx:b-pmtxij) ([raw file](https://github.com/pair-code/deeplearnjs/blob/447db4b4080769cecf66a7f54cee3af393f0d059/src/index.ts#L51)):*
> ```XML
> export {NDArrayMathCPU} from './math/backends/backend_cpu';
> export {NDArrayMathGPU} from './math/backends/backend_webgl';
> export {MatrixOrientation} from './math/backends/kernels/matmul';
> ```

looking at all files in math/backends/kernels/... , there is no actual implementation, they are only typings that go away at compile time.

To might lead to initial user confusion, so I think we should rename the folder `kernels` to `types`. WDYT?

---

*[src/math/math.ts, line 259 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-L03E2uusr-8Gd2PGHnL:-L03E2uusr-8Gd2PGHnM:bwkqqb0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/447db4b4080769cecf66a7f54cee3af393f0d059/src/math/math.ts#L259)):*
> ```XML
>   private executeOp<G extends keyof DataTypes, T extends NDArray<G>>(
>       name: string, f: () => T): T {
>     // TODO(nsthorat): Do operation logging and performance profiling.
> ```

I can see `executeOp` and `scope` converging, where `executeOp` is just a named scope, and regular scope is unnamed. That is:
`dl.scope('some name', () => {...});`

In debug mode you get grouping and performance aggregation for every scope, named and unnamed, for every user code, not just internal math.

---

*[src/math/math.ts, line 1302 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-L03HS9K0CIQtDXGCVYI:-L03HS9K0CIQtDXGCVYJ:b-tba8kk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/447db4b4080769cecf66a7f54cee3af393f0d059/src/math/math.ts#L1302)):*
> this.add(this.multiply(c1, a), this.multiply(c2, b))

add a todo to add SGEMM to backend and use it here.

---

*[src/math/backends/backend.ts, line 49 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-L039WKwWU2k6HAAckkB:-L039WKwWU2k6HAAckkC:b-7eby1g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/447db4b4080769cecf66a7f54cee3af393f0d059/src/math/backends/backend.ts#L49)):*
> ```XML
>  * this can be done gradually (throw an error for unimplemented methods).
>  */
> export interface MathBackend {
> ```

We should file an issue to document (contributions welcome) the kernels themselves, so people have clear guidelines for new backend implementations.

---

*[src/math/backends/kernel_registry.ts, line 103 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-L03A6t75gHKmWJ-DlJ2:-L03A6t75gHKmWJ-DlJ3:bejn8e5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/447db4b4080769cecf66a7f54cee3af393f0d059/src/math/backends/kernel_registry.ts#L103)):*
> config

Not 100% sure, but I feel that it would be better at some point if the kernels themselves (the functions in `MathBackend`) take list of arguments, instead of a single config, just like before this PR. This will make the backend interface immediately readable, instead of looking at typings of Config. In that case, here, we would have to call the backend with arguments:

```
=> backend.matMul(config.inputs.a, config.inputs.b,
      config.args.aOrientation, config.args.bOrientation)
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/426#-:-L03JDk-_ItjZI9vXr0I:bpm0gpd)*

PR #431, opened 2017-12-12
Handle error in case of 404 not found variable

Lewuathe:
`XMLHttpRequest` does not trigger `onerror` in case of 404. But even such case, `CheckpointLoader` should throw exception as well as `onerror` case.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/431)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/431#-:-L0AI3wRNeiNkeNEEYoe:b-jbsf8z)*

PR #432, opened 2017-12-12
Vue Demos: Fix UglifyJS errors while making website

ManrajGrover:
Currently, when we run `./scripts/make-website.sh`, UglifyJS throws an error while optimizing assets because it does not currently support ES6 and above. This PR removes the target as `es2017` and thus fixes the errors. This also makes sure that the demos work.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/432)


nsthorat:
Ah apologies.

Can you check out what I did here:
https://github.com/PAIR-code/deeplearnjs/commit/4a2a350ff2480a4fbf271b04fbc8ab5151142c34#diff-f1548fd08c9a2c852b537236e81c5cee

And make the same change to `build-demo` (so you can pass a target). And then from make-website we will pass the es5 target.

I made this change because debugging demos with es8 is *so much* simpler.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/432#-:-L0B6KIZH4RU2JVtkX9G:bhl68pd)*

ManrajGrover:
Yes, we should fix `build-demo` too. But looks like this error is caused while building Vue demos because of ES6+ code. I would suggest we have development and production configuration. That way we can separate targets.

I would need to check other ways on how to pass target to `poi` (one of them being diff of this PR).

nsthorat:
Yeah actually that's what I'm suggesting. I would also be okay with `watch-demo` passing `--target es2017` by default so that we get the es8 debugging for demo development.

Could you make that change?

Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/432#-:-L0GChv-dDhjwhlW20BI:b6s98cj)*

ManrajGrover:
@nsthorat @dsmilkov Merry Christmas! 😄

@nsthorat I've made changes that allow development and production configurations to exist for demos that make use of Vue. I'm working on adding changes to `build-demo` (as discussed above) for which I will raise a separate pull request. This PR can now be reviewed.

nsthorat:
Any reason to have two separate configs? Can we can just override target explicitly? If we add another divergence we can make a separate file.

Also, I think this doesn't cover watch-demo for non-vue demos. I think we just need to pass `--target es2017` from ./scripts/watch-demo.

Thanks for this awesome change, happy holidays :)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/432#-:-L1TByzcD-2mpsA8UZM8:blmh4ra)*

nsthorat:
Ah just saw the other PR. One question remains, why have 2 configs entirely?

ManrajGrover:
One main reason for following this approach was to pass the same development configuration to Vue builder `poi` as we are passing to polymer demos and since `poi` makes use of [ts-loader](https://github.com/TypeStrong/ts-loader), I could not find a better way to pass configuration to the library preset.

Another reason was to keep customizations out of the scripts and in one place together for all demos.

nsthorat:
:lgtm_strong:

Thanks for the explanation!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/432#-:-L1TJAKnDu07scPvg7oI:b7rm0uw)*

PR #434, opened 2017-12-12
2 standard deviations from the mean is (std * 2 +- mean).

nkreeger:
I accidentally used std squared which is throwing off randTruncatedNormal usages. I updated the code to reflect this. Due to this constraint - the precision of the rand truncated arrays are not as high and they won't pass the JB test. Simply ensure that the items fall within the bounds of truncation (std * 2 +- mean).

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/434)

### DIFF
--- a/src/math/matmul_test.ts
+++ b/src/math/matmul_test.ts
@@ -296,7 +296,7 @@ const commonTests: MathTests = it => {
 
 const gpuTests: MathTests = it => {
   it('Matrix times vector, large matrix', math => {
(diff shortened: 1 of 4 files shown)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/434#-:-L0BbsM8tyOwf3znPbfD:b-jbsf8z)*

PR #436, opened 2017-12-13
Add a Tape class and a gradientWrt method.

nsthorat:
* Add a `Tape`. The tape, which is owned by the `BackendEngine` records the kernel configurations on the forward pass. A tape is simply an array of `TapeNode` objects which `KernelNode` extends. A tape node itself can refer to another tape (not implemented yet).
* Add gradient to `Node`. This is a method that has the following signature:
`(dy: NDArray, y: NDArray) => {[inputName: string]: () => NDArray;}` The outer function is not executed until gradientWrt is called. Executing the outer function returns an object that maps each input to a function that returns its gradient. Since we prune the graph of any inputs that are not transitively a function of any of the xs, we don't want to compute gradients through those inputs, so we only call the gradient functions on the inputs where we need to backprop through. This also allows us to save variables to be shared between each of the inputs if we need to (instead of directly returning a map of inputName => gradient).

* Add `math.gradientWrt` which walks the tape backwards and computes the gradient of y with respect to all of the xs.

* Implements gradients for: matMul, relu, and sum

The tape walking algorithm is as follows:
1. Filter out the nodes that don't map xs[] => y. This is done by constructing a list of `NDArray`s that are a function of any of the input xs[] and constructing a list of `NDArray`s that lead to y. The nodes whose outputs are both from xs and to y are the nodes we care about. We also prune each node of any inputs that are not transitively a function of any of the xs so that when we execute backprop, we only backprop through inputs that will lead to the xs.
2. Walk the tape backwards, calling the gradient method on each of the nodes. This creates a set of gradients for each of the inputs. We accumulate these gradients in an activation NDArray => gradient NDArray map, computing the sum as we go. Here, we call directly into the backend method for sum so we don't add a configuration to the tape (we can revisit this).

What this PR does *not* do:
* Add the notion of a variable
* Dispose of `NDArray`s during a forward pass. If you use math.scope, this will error out for now.
* Create a sub-tape for the `gradientWrt` method. In a follow up CL, we will have a stack of tapes and an "activeTape" which is where new kernels will be pushed. When the `gradientWrt` method is called, the individual gradient functions which call top-level `NDArrayMath` operations will be recorded as a subtape. This sub-tape, which will be defined by a generic `TapeNode`, will have a gradient function which will call into its own sub-tapes `gradientWrt` method. This will allow us to compute second order derivatives.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/436)


nsthorat:
Review status: 0 of 29 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/math/backends/tape.ts, line 46 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0I7j9JcoqqOVoNHNsE:-L0I7j9KJ4T6Vhe0qzB2:b8do17c) ([raw file](https://github.com/pair-code/deeplearnjs/blob/70e8838a682e36f580148716adabba1d096016a3/src/math/backends/tape.ts#L46)):*
> ```XML
>     arrayAccumulatedGradientMap[y.id] = Scalar.new(1);
> 
>     // Walk the tape backwards and keep a map of NDArray to its gradient.
> ```

Extract this to a util method "backprop" that we can test directly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436)*

nsthorat:
Review status: 0 of 29 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/math/backends/tape.ts, line 46 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0I7j9JcoqqOVoNHNsE:-L0KdyGnNokzpmC-2cgL:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/70e8838a682e36f580148716adabba1d096016a3/src/math/backends/tape.ts#L46)):*

Extract this to a util method "backprop" that we can test directly
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436)*

dsmilkov:
:lgtm_strong: Awesome work!! So glad to see the first signs of eager mode! Left a dozen comments (it's a big CL), but nothing major

---

Reviewed 24 of 29 files at r1, 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/math/math.ts, line 2205 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LL61Y3NHs-d5th70e:-L0LL61Y3NHs-d5th70f:b-3z8s4z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math.ts#L2205)):*
> ```XML
>    * Warning: this is not fully implemented yet. Use with caution.
>    *
>    * Computes a gradient of y with respect to x.
> ```

I'm trying to remember why TF eager, `tf.eager.gradients_function` in particular, returns a function that needs to be called in order to get the gradients, instead of directly returning the gradients themselves.

Any thoughts? This is no blocker, we can revisit later.

---

*[src/math/math.ts, line 2213 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LKycKYFsjr2S4fXgA:-L0LKycKYFsjr2S4fXgB:btzz1rg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math.ts#L2213)):*
> ```XML
>    * method will return an object of the same shape.
>    */
>   gradientWrt<T extends NDArray|{[xName: string]: NDArray}>(y: Scalar, x: T):
> ```

Add an optional dy in the param list?

---

*[src/math/math_test.ts, line 284 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LFOU8RwDNaVa_gvKj:-L0LFOU9Ycd97JTlMq-s:bk430of) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math_test.ts#L284)):*
> grads

nit: slightly more readable to unpack directly:
`const {da, db} = math.gradientWrt(e, {a, b});`

---

*[src/math/matmul_test.ts, line 361 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LFkHdA-UJ6KlPKVh5:-L0LFkHdA-UJ6KlPKVh6:bkj5yy4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/matmul_test.ts#L361)):*
> ```XML
> 
> test_util.describeMathCPU('gradientWrt matMul', [commonTests, gradientTests]);
> test_util.describeMathGPU('gradientWrt matMul', [commonTests, gpuTests], [
> ```

forgot to add `gradientsTests` to `describeMathGPU`, or left out on purpose?

---

*[src/math/reduction_ops_test.ts, line 520 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LM1diEp5Hd9baOjbS:-L0LM1diEp5Hd9baOjbT:brz5phg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/reduction_ops_test.ts#L520)):*
> ```XML
> 
>       test_util.expectArraysClose(
>           grad.dataSync(), new Float32Array([1, 1, 1, 1, 1, 1]));
> ```

see previous comment about adding an optional dy, and then adding tests with dy that is not `dl.ones`

---

*[src/math/reduction_ops_test.ts, line 532 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LJ_uc5oDO_PfILIxY:-L0LJ_uc5oDO_PfILIxZ:bg6gmh2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/reduction_ops_test.ts#L532)):*
> ```XML
>       test_util.expectArraysClose(
>           grad.getValues(), new Float32Array([1, 1, 1, 1, 1, 1]));
>     });
> ```

can you add 2 more tests, sum 2d across axis=0 and sum 2d across axis=1, analogous to tests at line 467 and 481. I think it should work right away.

---

*[src/math/unaryop_test.ts, line 135 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LMgLNwkvpoaB6T29M:-L0LMgLNwkvpoaB6T29N:b1ye6qz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/unaryop_test.ts#L135)):*
> ```XML
> 
>     it('Relu gradient array', math => {
>       const a = Array2D.new([2, 2], [1, -1, 2, .1]);
> ```

add a zero as an entry to test at the boundary 

---

*[src/math/backends/backend_engine.ts, line 68 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LPCYVJO_ocw_AKzyW:-L0LPCYVJO_ocw_AKzyX:b-gddg05) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/backend_engine.ts#L68)):*
> ```XML
>     const evaluatedNode: KernelNode = {
>       name: `kernel: ${kernelName}`,
>       kernel: kernelName,
> ```

any reason why we'll need both `kernel` and `name`?

---

*[src/math/backends/tape.ts, line 29 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LRTe5ws_vUTccubwn:-L0LRTe5ws_vUTccubwo:b-puv49y) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape.ts#L29)):*
> KernelNode

can this be a `TapeNode`, or does it have to be a `KernelNode`?

---

*[src/math/backends/tape.ts, line 34 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LNpl60QAMRcLVTksO:-L0LNpl60QAMRcLVTksP:b-pa28zd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape.ts#L34)):*
> backend: MathBackend

Any strong reason not to pass backend into the Tape c-tor and let tape hold onto that backend? I know that the current approach reduces state, but given that our backend won't switch within executions, state shouldn't be an issue.

---

*[src/math/backends/tape_config.ts, line 1 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LRyg2Kt21AZqJDoYs:-L0LRyg2Kt21AZqJDoYt:bvjq6jt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape_config.ts#L1)):*
> ```XML
> /**
> ```

consider renaming to tape_types.ts or move to types/tape.ts since it's all types/interfaces.

---

*[src/math/backends/tape_config.ts, line 41 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LPVem2UpmfHJy8QG7:-L0LPVem2UpmfHJy8QG8:b-9ohc3z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape_config.ts#L41)):*
> ```XML
> 
> // Kernel nodes
> export interface KernelNode extends TapeNode {
> ```

will we ever have other types of `TapeNode`s , in addition to `KernelNode`s?

Maybe we should consolidate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-:-L0LS7CUQKvdyiu3RMW3:b-x4w9rq)*

nsthorat:
Review status: 3 of 29 files reviewed at latest revision, 13 unresolved discussions, some commit checks pending.

---

*[src/math/math.ts, line 2205 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LL61Y3NHs-d5th70e:-L0LzG6ZM4lNbku0alLD:bmgvbg3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math.ts#L2205)):*

I'm trying to remember why TF eager, `tf.eager.gradients_function` in particular, returns a function that needs to be called in order to get the gradients, instead of directly returning the gradients themselves.

Any thoughts? This is no blocker, we can revisit later.
</blockquote></details>

Yeah, I don't totally understand why they do that. I think this API is nicer to the user, but we can revisit this later!

---

*[src/math/math.ts, line 2213 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LKycKYFsjr2S4fXgA:-L0M51lhbiuzH0PR5fJS:bhg0r2a) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math.ts#L2213)):*

Add an optional dy in the param list?
</blockquote></details>

Spoke offline: this doesn't seem to be too useful right now. We can always add this later.

---

*[src/math/math_test.ts, line 284 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LFOU8RwDNaVa_gvKj:-L0LzmvdxRbiT5hlVQlT:bk1jp17) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/math_test.ts#L284)):*

> grads

nit: slightly more readable to unpack directly:
`const {da, db} = math.gradientWrt(e, {a, b});`
</blockquote></details>

Unfortunately you can't do do this destructing because the return type is the shape of {a, b} so you would have to destruct it as {a, b}.

---

*[src/math/matmul_test.ts, line 361 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LFkHdA-UJ6KlPKVh5:-L0M-3KT4I9ztvPdSiAb:b-73m6vw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/matmul_test.ts#L361)):*

forgot to add `gradientsTests` to `describeMathGPU`, or left out on purpose?
</blockquote></details>

Nice catch!

---

*[src/math/reduction_ops_test.ts, line 520 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LM1diEp5Hd9baOjbS:-L0M3iagrCt8_eveZcm4:b-d98ker) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/reduction_ops_test.ts#L520)):*

see previous comment about adding an optional dy, and then adding tests with dy that is not `dl.ones`
</blockquote></details>

Spoke offline: this doesn't seem to too useful, but we can add it if we find a use case for this

---

*[src/math/reduction_ops_test.ts, line 532 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LJ_uc5oDO_PfILIxY:-L0M50ki3yAagabjn8mD:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/reduction_ops_test.ts#L532)):*

can you add 2 more tests, sum 2d across axis=0 and sum 2d across axis=1, analogous to tests at line 467 and 481. I think it should work right away.
</blockquote></details>

Done.

---

*[src/math/unaryop_test.ts, line 135 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LMgLNwkvpoaB6T29M:-L0MSYO45wnW47LNMFv3:bd8xoyz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/unaryop_test.ts#L135)):*

add a zero as an entry to test at the boundary 
</blockquote></details>

Added -.001. Actually 0 fails because of precision ..... Ugh!

---

*[src/math/backends/backend_engine.ts, line 68 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LPCYVJO_ocw_AKzyW:-L0M-lmlhLQxd3y7Ei2L:bk3ggjj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/backend_engine.ts#L68)):*

any reason why we'll need both `kernel` and `name`?
</blockquote></details>

Name is generic (for gradients this will be a display now), kernel specifies the exact kernel so we can differentiate them.

---

*[src/math/backends/tape.ts, line 29 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LRTe5ws_vUTccubwn:-L0M02ytSjDljamNwA17:b-14wefg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape.ts#L29)):*

> KernelNode

can this be a `TapeNode`, or does it have to be a `KernelNode`?
</blockquote></details>

For now it has to be a kernel node, in a follow up PR this may become a TapeNode (driven by scopes, etc), but I haven't figured that out yet. For now I renamed it.

---

*[src/math/backends/tape.ts, line 34 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LNpl60QAMRcLVTksO:-L0M0o0MiOIDRx1BXGFh:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape.ts#L34)):*

> backend: MathBackend

Any strong reason not to pass backend into the Tape c-tor and let tape hold onto that backend? I know that the current approach reduces state, but given that our backend won't switch within executions, state shouldn't be an issue.
</blockquote></details>

Done.

---

*[src/math/backends/tape_config.ts, line 1 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LRyg2Kt21AZqJDoYs:-L0M0P5DWI_YnwU6SSQr:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape_config.ts#L1)):*

consider renaming to tape_types.ts or move to types/tape.ts since it's all types/interfaces.
</blockquote></details>

Done.

---

*[src/math/backends/tape_config.ts, line 41 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436#-L0LPVem2UpmfHJy8QG7:-L0M0PhSE0gXmnKBxi7W:b-l05jqd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/d854437f77431640811b9abb79ebd9af38a3ca32/src/math/backends/tape_config.ts#L41)):*

will we ever have other types of `TapeNode`s , in addition to `KernelNode`s?

Maybe we should consolidate.
</blockquote></details>

Yes, named sub-scopes, and custom gradient sub-tapes will be different type of TapeNodes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/436)*

PR #444, opened 2017-12-14
Improve backend registration and move it as part of the global ENV

dsmilkov:
- Backend registration is now moved to the global environment (`ENV`).
- Registration of backends is done with a factory method, which fails gracefully when the backend can't be constructed due to limitations in the environment (e.g. when WebGL is not supported on the device).
- Added tests for backend registration.
- Decouple the creation of `NDArrayMath` and setting it as a global math. We want to avoid implicit global changes. `NDArrayMathCPU` and `NDArrayMathGPU` set themselves as global math, for backwards compatibility, but those will be removed soon.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/444)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/environment.ts, line 262 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/444#-L0MS5qc6FrFenzibCVD:-L0MS5qc6FrFenzibCVE:bx0z8me) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1b0f4116663116caeb19d11216fe40673b8d4fd2/src/environment.ts#L262)):*
> ```XML
> 
>   // tslint:disable-next-line:no-any
>   private backendRegistry: {[id in Backends]: MathBackend} = {} as any;
> ```

do you need as any here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/444#-:-L0MSJaa9h2hqSEJfO6I:b-jbsf8z)*

dsmilkov:
Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/environment.ts, line 262 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/444#-L0MS5qc6FrFenzibCVD:-L0OhJOEOGP3FiELh3ls:b-393ggt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/1b0f4116663116caeb19d11216fe40673b8d4fd2/src/environment.ts#L262)):*
Unfortunately, yes. Otherwise you get the error:
> Type '{}' is not assignable to type '{ webgl: MathBackend; cpu: MathBackend; }'. Property 'webgl' is missing in type '{}'

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/444)*

PR #446, opened 2017-12-15
Add option to run tests under swift shader.

nkreeger:
I don't know if it makes sense to make this a default test for now - but maybe something on travis CI. It's not terribly fast for running the whole sweet of tests.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/446)


nsthorat:
I think we don't really want to run the tests under another environment, I was more curious just a relative benchmark comparison with screenshots. And mostly to make sure that if a blacklisted device uses deeplearn.js we *just work* without failing because of "failIfMajorPerformanceCaveat"

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/446#-:-L0QLM9QJKh7hHYMT-J7:b-6p5yz7)*

dsmilkov:
So tests just pass without any changes on your side? We were expecting the tests to fail since [this flag](https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/backends/webgl/gpgpu_util.ts#L31) is on.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/446#-:-L0lqbLbtwzlGm7p2rji:b-or2gr1)*

nkreeger:
Correct. Starting chrome with those flags disables webgl and switches to
swiftshader. I'll try and find cycles where I can run the benchmarks w/o
other tasks rolling on my workstation.

dsmilkov:
:lgtm_strong: Good to have this checked in either way. We can decide later if we should add it to BrowserStack (will slow down Travis PR testing)

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/446#-:-L1sXSYwDNJSAWgAY10Y:b6mnc73)*

PR #448, opened 2017-12-15
Consolidate Tape logic and scope logic.

nsthorat:
This PR:

* Moves all of the scope logic into the `backendEngine`. We do not break `math.scope()` users as we just push those requests into the `backendEngine`. This can be renamed in a follow up PR, but is not renamed here for ease of review.
* Removes the `Tape` class. Now, `Tape` is simply an array of `TapeNodes`. This is so that serialization is as easy as JSON-ifying the `Tape` (with some exceptions like the gradient function).
* Consolidates the stack logic for scope and tape. When a new scope is created, now we create a subtape.
* Calls scope in the `gradientWrt` method while backpropogating gradients. This scopes all of the backprop code into a subtape element. This subtape element will have its inputs automatically computed (so the subtape tape node operates as any other node on the tape, having inputs, outputs, and a gradient method).
* Adds multi-output functionality to `TapeNode` logic, including filtering and backpropagation. This is necessary because `gradientWrt` outputs multiple nodes. For now, we leave `KernelNode` to only have a single output, but we will easily be able to convert it to multi-output if we ever want to.
* Adds gradients for strict multiply, subtract, pow. We will flesh out broadcasted in a follow up PR.
* Add lots of unit tests for utility methods.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/448)


dsmilkov:
:lgtm_strong: Nice job on the tests! Especially tape_util_test which helped me understand what is going on :)

---

Reviewed 23 of 30 files at r1, 4 of 8 files at r2, 5 of 6 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/util.ts, line 25 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v5vjhbpBbzDHJOB8W:-L0v5vjiXdfxN9-jX_HC:b-4cwbiy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/util.ts#L25)):*
> NameArrayMap

slightly prefer `NamedArrayMap` more, but up to you

---

*[src/math/element_wise_arithmetic_test.ts, line 243 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0vA8wGeXKu0IS0CH1p:-L0vA8wGeXKu0IS0CH1q:bzrn6r) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/element_wise_arithmetic_test.ts#L243)):*
> ```XML
> 
>       expect(() => math.gradientWrt(result, a)).toThrowError();
>       expect(() => math.gradientWrt(result, b)).toThrowError();
> ```

a bit ambigious why this test will throw. Could be that result is not a Scalar, not necessarily because broadcast should throw. Maybe call result = math.sum(result)

Also, is the idea to delete this test in the next PR , once we have broadcasing support?

---

*[src/math/math.ts, line 953 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v7wG2TBtD6XOmeWZi:-L0v7wG2TBtD6XOmeWZj:bshl30) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/math.ts#L953)):*
> ```XML
>                      });
>                    },
>                    b: () => {
> ```

this is a bit hard to read. 

split into variables, here and for `math.sub` as well
```
const config = {inputs: {a, b}};
const derA = () => {...}
const derB = () => {...}
const der = (dy, y) => {a: derA, b:derB}

executeKernel('Pow', config, der);
```

---

*[src/math/math.ts, line 2286 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v9vLZJnSZfhJZtfFh:-L0v9vLZJnSZfhJZtfFi:b-rmmrym) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/math.ts#L2286)):*
> ```XML
>    */
>   gradientWrt<T extends NDArray|NameArrayMap>(y: Scalar, x: T): T {
>     const xs = util.flattenNameArrayMap(x);
> ```

missed from previous PR. Can you explicitly check that y.rank == 1 at run-time

---

*[src/math/matmul_test.ts, line 332 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v93HFVjxStkoTXZu0:-L0v93HFVjxStkoTXZu1:b-ifpin2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/matmul_test.ts#L332)):*
> onst expectedDa = math.matMul(
>         Array2D.ones([2, 2]), b, MatrixOrientation.REGULAR,

better to hard-code the values of `expectedDa` and `expectedDb` so this test is not linked to the impl of matmul

---

*[src/math/unaryop_test.ts, line 398 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0vAeHElHXpMSId6aBI:-L0vAeHElHXpMSId6aBJ:b-tic2k7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/unaryop_test.ts#L398)):*
> 1e-1

curious why this and following tests need such a loose bound `1e-1` and not the previous test?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-:-L0vEWUZIaw-3_z2Z5s8:bl9hqzv)*

nsthorat:
Thanks for the review!

---

Reviewed 28 of 30 files at r1, 6 of 8 files at r2, 5 of 6 files at r3.
Review status: 6 of 33 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/util.ts, line 25 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v5vjhbpBbzDHJOB8W:-L0vJkqk3gCSoKer6RaP:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/util.ts#L25)):*

> NameArrayMap

slightly prefer `NamedArrayMap` more, but up to you
</blockquote></details>

Done.

---

*[src/math/element_wise_arithmetic_test.ts, line 243 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0vA8wGeXKu0IS0CH1p:-L0vKRPwCZzbWcCf1jE-:bols01g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/element_wise_arithmetic_test.ts#L243)):*

a bit ambigious why this test will throw. Could be that result is not a Scalar, not necessarily because broadcast should throw. Maybe call result = math.sum(result)

Also, is the idea to delete this test in the next PR , once we have broadcasing support?
</blockquote></details>

Good point. I used sum, and yes that's the idea.

---

*[src/math/math.ts, line 953 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v7wG2TBtD6XOmeWZi:-L0vLnnMgN_PCcEVxP91:beegy5r) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/math.ts#L953)):*

this is a bit hard to read. 

split into variables, here and for `math.sub` as well
```
const config = {inputs: {a, b}};
const derA = () => {...}
const derB = () => {...}
const der = (dy, y) => {a: derA, b:derB}

executeKernel('Pow', config, der);
```
</blockquote></details>

Unfortunately I can't break it up like you're suggesting because of the dependency on dy, but I pulled "der" out to a separate variable.

---

*[src/math/math.ts, line 2286 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v9vLZJnSZfhJZtfFh:-L0vMZwvjJJfm8UyKwNz:b-il6iov) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/math.ts#L2286)):*

missed from previous PR. Can you explicitly check that y.rank == 1 at run-time
</blockquote></details>

Done. And added a unit test.

---

*[src/math/matmul_test.ts, line 332 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v93HFVjxStkoTXZu0:-L0vMorFFxixEXhuQGUi:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/matmul_test.ts#L332)):*

> onst expectedDa = math.matMul(
>         Array2D.ones([2, 2]), b, MatrixOrientation.REGULAR,

better to hard-code the values of `expectedDa` and `expectedDb` so this test is not linked to the impl of matmul
</blockquote></details>

Done.

---

*[src/math/unaryop_test.ts, line 398 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0vAeHElHXpMSId6aBI:-L0vOQd_XrefVTW3ya6O:bzi2wc4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/unaryop_test.ts#L398)):*

> 1e-1

curious why this and following tests need such a loose bound `1e-1` and not the previous test?
</blockquote></details>

Ah, great question. The problem is that the error accumulates because of the sum :/

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-:-L0vPj8Y7eZkZFnqzcEY:b-9qdoz3)*

nsthorat:
Review status: 6 of 34 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 953 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448#-L0v7wG2TBtD6XOmeWZi:-L0vQQHCGcR63A2KzPbk:bs6s3aa) ([raw file](https://github.com/pair-code/deeplearnjs/blob/47bc61d4fe483e9620c6b0a166a398bbf770f95a/src/math/math.ts#L953)):*

Unfortunately I can't break it up like you're suggesting because of the dependency on dy, but I pulled "der" out to a separate variable.
</blockquote></details>

Okay, pulled derA and derB out inside of the der function itself. Much better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/448)*

PR #450, opened 2017-12-16
Configure typedoc with typedocOptions in tsconfig.json

Lewuathe:
Since typedocOptions in `tsconfig.json` provides the way to configure typedoc, we can make `tsconfig-doc.json` obsolete. typedoc reads tsconfig.json as default. Command to generate API document becomes simpler.

Usage
```
$ typedoc --out ./apidoc src
```

- [x] : Make sure to exclude test code from api document
- [x] : Make sure to exclude private classes, method, variables from api document.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/450)


nsthorat:

---

*[tsconfig.json, line 24 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/450#-L0W4qH48T6vGokGTbdC:-L0W4qH48T6vGokGTbdD:bgreirw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b3f114867ce206a4925efe3eceae40b15b816c4b/tsconfig.json#L24)):*
> ```JSON
>     "src/"
>   ],
>   "typedocOptions": {
> ```

so the "include" list above is so that we can whitelist exactly what we export - tsconfig doesnt give us a way to start from somewhere (ie, index.ts) since we export methods from individual files but not at the module level.

Can you make sure the API documentation stays the same (with make-website) after this change and we're not showing extraneous documentation?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/450)*

Lewuathe:

---

*[tsconfig.json, line 24 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/450#-L0W4qH48T6vGokGTbdC:-L0mvC5auuWo5S27IOhj:btd81nv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b3f114867ce206a4925efe3eceae40b15b816c4b/tsconfig.json#L24)):*

so the "include" list above is so that we can whitelist exactly what we export - tsconfig doesnt give us a way to start from somewhere (ie, index.ts) since we export methods from individual files but not at the module level.

Can you make sure the API documentation stays the same (with make-website) after this change and we're not showing extraneous documentation?
</blockquote></details>

Thanks. Let me check with `make-website`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/450)*

Lewuathe:
@nsthorat Finally I found `typedocOptions` and existing options like `include` and `extends` didn't work as expected. Since `typedocOptions` did not provide `include` option for now, we may need to accept some incompatible change in documentation.

So I though it may be better to just fix outdated path in `tsconfig-doc.json` in this time.

dsmilkov:
:lgtm_strong: Thanks for updating the list of files!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/450#-:-L1sWMH5RgzenjXXduM6:b9yj58f)*

PR #466, opened 2017-12-25
Add delayed upload mode in MathBackendWebGL

dsmilkov:
- Add delayed upload mode in `MathBackendWebGL`. When delayed mode is `true`, newly constructed `NDArray`s won't get immediately uploaded to the GPU. Also when user reads back the values, the backing texture gets disposed to reduce probability of memory leaks. Delayed mode is `true` by default.
- Fix `math.getNumArrays()`
- Fix memory leaks in Squeezenet
- Fix bugs in `models/*/package.json` (`tsc` creates different dir structure under `dist/` once we added dependency on `../util.ts`)
- Fix imagenet demo flickering

After this PR is merged, I'll release new npm versions in this order:
1) deeplearn
2) deeplearn-squeezenet
3) deeplearn-knn-image-classifier

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/466)

### DIFF
--- a/demos/imagenet/imagenet.ts
+++ b/demos/imagenet/imagenet.ts
@@ -235,6 +235,8 @@ export class ImagenetDemo extends ImagenetDemoPolymer {
           this.backend.getTextureData(activationNDArray.id).texShape,
           activationNDArray.shape[0], activationNDArray.shape[2],
           this.inferenceCanvas.width, numRows);
+      // Unclear why, but unless we wait for the gpu to fully end, we get a
+      // flicker effect.
       await maxValues.data();
       await minValues.data();
     });
--- a/demos/imagenet/imagenet.ts
+++ b/demos/imagenet/imagenet.ts
@@ -75,6 +75,13 @@ export class ImagenetDemo extends ImagenetDemoPolymer {
     this.webcamVideoElement =
         this.querySelector('#webcamVideo') as HTMLVideoElement;
 
+    const gl = gpgpu_util.createWebGLContext(this.inferenceCanvas);
+    this.gpgpu = new GPGPUContext(gl);
+    this.backend = new MathBackendWebGL(this.gpgpu);
+    const safeMode = false;
+    this.math = new NDArrayMath(this.backend, safeMode);
+    ENV.setMath(this.math);
+
     this.layerNames = [
       'conv_1', 'maxpool_1', 'fire2', 'fire3', 'maxpool_2', 'fire4', 'fire5',
       'maxpool_3', 'fire6', 'fire7', 'fire8', 'fire9', 'conv10'
@@ -100,13 +107,6 @@ export class ImagenetDemo extends ImagenetDemoPolymer {
         });
       }
     });
-
-    const gl = gpgpu_util.createWebGLContext(this.inferenceCanvas);
-    this.gpgpu = new GPGPUContext(gl);
-    this.backend = new MathBackendWebGL(this.gpgpu);
-    const safeMode = false;
-    this.math = new NDArrayMath(this.backend, safeMode);
-    ENV.setMath(this.math);
 
     this.renderGrayscaleChannelsCollageShader =
         imagenet_util.getRenderGrayscaleChannelsCollageShader(this.gpgpu);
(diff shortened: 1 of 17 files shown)


nsthorat:
:lgtm_strong:

Nice work!

---

Reviewed 14 of 14 files at r1.
Review status: 13 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[demos/imagenet/imagenet.ts, line 238 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1G3ayq7VfQwzyHRyMv:-L1G3ayq7VfQwzyHRyMw:b-ts8svg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/demos/imagenet/imagenet.ts#L238)):*
> ```XML
>           activationNDArray.shape[0], activationNDArray.shape[2],
>           this.inferenceCanvas.width, numRows);
>       await maxValues.data();
> ```

can you put a comment here saying this helps with flickering

---

*[src/math/backends/backend_webgl.ts, line 72 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1IYGOulsa65LLZZ6eE:-L1IYGOulsa65LLZZ6eF:bxt58lg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/src/math/backends/backend_webgl.ts#L72)):*
> ```XML
>     let texture: WebGLTexture;
>     const texShape: [number, number] = [pixels.height, pixels.width];
>     if (id in this.texData &&
> ```

out of curiosity, when would this be true? in fromPixels, it seems like you make a new NDArray every time

---

*[src/math/backends/backend_webgl.ts, line 784 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1IZt1I7HcxKy2wuZxN:-L1IZt1I7HcxKy2wuZxO:bribmnl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/src/math/backends/backend_webgl.ts#L784)):*
> ```XML
>     this.texData[id].texture = newTexture;
>     this.texData[id].values = null;
>     if (values != null) {
> ```

when would values be null? should we throw?

---

*[src/math/backends/webgl/texture_manager.ts, line 25 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1I_hlZoDzdn-6sTLGn:-L1I_hlZoDzdn-6sTLGo:bf0eug2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/76742f07d27958cfc4af298e4685cbd1ac5aaae7/src/math/backends/webgl/texture_manager.ts#L25)):*
> ```XML
>   private freeTextures: {[shape: string]: WebGLTexture[]} = {};
>   private logEnabled = false;
>   private usedTextureCount: {[shape: string]: number} = {};
> ```

out of curiosity why are you removing this? Because we can move this up into the backend later? I added this for debugging purposes (so we know what shapes are leaking)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-:-L1Ia2mgyMEnhvsfuYCQ:b-rtmsva)*

dsmilkov:
Thanks!

---

Review status: 11 of 14 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[demos/imagenet/imagenet.ts, line 238 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1G3ayq7VfQwzyHRyMv:-L1J0EtXIyGpF8polId0:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/demos/imagenet/imagenet.ts#L238)):*

can you put a comment here saying this helps with flickering
</blockquote></details>

Done.

---

*[src/math/backends/backend_webgl.ts, line 72 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1IYGOulsa65LLZZ6eE:-L1J0KxAeKQmveJeaovp:bwz7lxp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/src/math/backends/backend_webgl.ts#L72)):*

out of curiosity, when would this be true? in fromPixels, it seems like you make a new NDArray every time
</blockquote></details>

Assume immediate storage mode. Inside `NDArray` c-tor (regardless of whether it's backed by pixels or values), we call `math.write`, which allocates a texture of shape given by `webgl_util.getTexShapeFromLogical`. We need this when we make output array used by gpgpu programs (`this.makeOutputArray`). But then, since we have pixels, we actually want a texShape of `[height, width]` (channels are not part of the shape), and this is achieved by calling `math.writePixels` from `NDArray.fromPixels`,  so we have to dispose the old texture with the incorrect shape.

---

*[src/math/backends/backend_webgl.ts, line 784 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1IZt1I7HcxKy2wuZxN:-L1J1T8CLYTU-oG1x5Lv:b8tqzcn) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4249b0e0c75ebc1a3c8483a3f24b30713b1ddfbc/src/math/backends/backend_webgl.ts#L784)):*

when would values be null? should we throw?
</blockquote></details>

Values will be null whenever we need to allocate a new array, backed by a new texture, but not populate it with values. We need this when we create an output array for each math operation. See `this.makeOutputArray()`.

---

*[src/math/backends/webgl/texture_manager.ts, line 25 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-L1I_hlZoDzdn-6sTLGn:-L1J3mAI6iyWdzKgYfLj:b-6rt9cz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/76742f07d27958cfc4af298e4685cbd1ac5aaae7/src/math/backends/webgl/texture_manager.ts#L25)):*

out of curiosity why are you removing this? Because we can move this up into the backend later? I added this for debugging purposes (so we know what shapes are leaking)
</blockquote></details>

got it. I removed it because it was private and never used to affect logic internally nor exposed externally. Didn't know you are using it in debug mode. Reverting the removal.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/466#-:-L1J4We7SxkBK4am7bZk:b-ucfknh)*

PR #478, opened 2017-12-27
Add rank as generic type and introduce Variable (updateble NDArray)

dsmilkov:
- Add rank as generic. Now we have `NDArray<D, R>` where D is dtype and R is rank. Rank specific classes constrain the rank, that is: `Scalar<D> extends NDArray<D, '0'>`. `Array1D<D> extends NDArray<D, '1'>`, etc.
- Update a few files due to more explicit and powerful inference. Methods that want `Scalar` can't take `NDArray` anymore, since `NDArray` is not assignable to `Scalar`, but `Scalar` is assignable to `NDArray`.
- Add initial implementation of `Variable`, which is basically a globally tracked and update-able `NDArray`. It extends `NDArray` so math ops can take it without requiring code changes.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/478)


nsthorat:
:lgtm_strong:

Really great work again. Variables look really solid, and the new rank generics are sweet. Nice work making it backwards compatible.

---

Reviewed 14 of 19 files at r1, 2 of 4 files at r3, 6 of 6 files at r4.

---

*[src/math/math.ts, line 56 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1SuE4-mO7oAj47NZQo:-L1SuE4-mO7oAj47NZQp:b-8ilume) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/math.ts#L56)):*
> ```XML
> 
>   // Public since optimizers will use it.
>   registeredVariables = new Map<string, Variable>();
> ```

any reason to use map type? we've been using the normal {[string: varName]: Variable} everywhere else

---

*[src/math/ndarray.ts, line 53 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1Svp8qxseRO0TUmy_u:-L1Svp8qxseRO0TUmy_v:b-sw988f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/ndarray.ts#L53)):*
> ```XML
> }
> 
> export interface Rank {
> ```

does this need to be an interface?

I think you can just export a type:

`export type Rank = '0'|'1'|'2'|'3'|'higher';`

And then `R extends keyof Rank` becomes `R extends Rank`

---

*[src/math/variable.ts, line 34 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1Sqgaav5T1tnDPSbuP:-L1Sqgaav5T1tnDPSbuQ:b2lcsj9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/variable.ts#L34)):*
> ```XML
>   private constructor(
>       initialValue: NDArray<D, R>, public trainable = true, name?: string) {
>     super(initialValue.shape, initialValue.dtype, null, initialValue.id);
> ```

could you add an inline comment next to null:

, null /* values */, 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-:-L1SxusTnmvKGI1PSUe9:b4t5k0k)*

dsmilkov:
Thanks!

---

Reviewed 16 of 19 files at r1, 2 of 3 files at r2, 4 of 4 files at r3.

---

*[src/math/math.ts, line 56 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1SuE4-mO7oAj47NZQo:-L1SyR5LFUR0uwFqg-Ou:bi56ajg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/math.ts#L56)):*

any reason to use map type? we've been using the normal {[string: varName]: Variable} everywhere else
</blockquote></details>

`Map`s are built-in data structures and potentially more optimized than {}, because they have more constraints.

---

*[src/math/ndarray.ts, line 53 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1Svp8qxseRO0TUmy_u:-L1T-6t_A2OjiDoNqilM:b-83sjoh) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/ndarray.ts#L53)):*

does this need to be an interface?

I think you can just export a type:

`export type Rank = '0'|'1'|'2'|'3'|'higher';`

And then `R extends keyof Rank` becomes `R extends Rank`
</blockquote></details>

Nice!

---

*[src/math/variable.ts, line 34 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-L1Sqgaav5T1tnDPSbuP:-L1T-Hiqn0jnKlD7Ic5M:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ab8e15458c31fd2d32c68fc9cb703e72cf30e05f/src/math/variable.ts#L34)):*

could you add an inline comment next to null:

, null /* values */, 
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/478#-:-L1SyNy7HaPQqhHRGs5S:b-ucfknh)*

PR #486, opened 2017-12-28
Add math.vjp, custom gradients, softmax gradients, softmaxCrossEntropyWithLogits + gradients.

nsthorat:
This PR:
* Adds `math.vjp()`, which is like `gradients()`, but allows f() to return any NDArray, and takes a dy to backpropagate. Returns the vector-jacobian product as an NDArray.
* Adds `math.customGradient` which takes a function that is executed immediately, as well as a gradient function to be used on the tape.
* Add softmaxCrossEntropyWithLogits op.
* Add custom gradient for softmax and softmaxCrossEntropyWithLogits. softmaxCrossEntropyWithLogits calls softmax - the custom gradient for softmaxCrossEntropyWithLogits overrides the custom gradient for softmax.
* Converts unit tests to use vjp. This allows us to test functions with any dy.
* Fix gradients where we didn't multiply by dy, and add unit tests.
* Adds unit tests for gradientsScope to test memory cleanup.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/486)

### DIFF
--- a/src/math/backends/backend_engine.ts
+++ b/src/math/backends/backend_engine.ts
@@ -100,11 +100,11 @@ export class BackendEngine {
   }
 
   customGradient<G extends DataType, T extends NDArray<G>>(
-      name: string, f: () => {
+      f: () => {
         value: T,
         gradients: (dy: T, y: T) => TapeNodeInputGradientArrays
       },
-      inputs: NamedArrayMap): T {
+      inputs: NamedArrayMap, name: string): T {
     this.customGradientDepth++;
 
     let gradientsFunc: (dy: T, y: T) => TapeNodeInputGradientArrays;
@@ -118,14 +118,11 @@ export class BackendEngine {
     this.customGradientDepth--;
 
     if (this.activeTape != null && this.customGradientDepth === 0) {
-      const inputAndArgs =
-          tape_util.stripUndefinedInputsFromInputConfig({inputs});
-
       const evaluatedNode: TapeNode<T> = {
         id: this.nextTapeNodeId++,
         type: 'customGradient',
         name,
-        inputAndArgs,
+        inputAndArgs: {inputs},
         output: result,
         gradient: gradientsFunc
       };
--- a/src/math/backends/backend_engine.ts
+++ b/src/math/backends/backend_engine.ts
@@ -109,6 +109,8 @@ export class BackendEngine {
       return f();
     }, gradientsMode);
 
+    this.customGradientDepth--;
+
     if (this.activeTape != null && this.customGradientDepth === 0) {
       const inputAndArgs =
           tape_util.stripUndefinedInputsFromInputConfig({inputs});
@@ -121,10 +123,10 @@ export class BackendEngine {
         output: result,
         gradient
       };
+
       this.activeTape.push(evaluatedNode);
     }
 
-    this.customGradientDepth--;
     return result;
   }
 
--- a/src/math/backends/backend_engine.ts
+++ b/src/math/backends/backend_engine.ts
@@ -16,11 +16,14 @@
  */
 
 import * as util from '../../util';
+import {NamedArrayMap} from '../../util';
 import {DataType, NDArray, Scalar} from '../ndarray';
+
 import {MathBackend} from './backend';
 import * as kernel_registry from './kernel_registry';
 import {KernelConfigRegistry} from './kernel_registry';
-import {KernelNode, Tape} from './tape_types';
+// tslint:disable-next-line:max-line-length
+import {KernelNode, Tape, TapeNode, TapeNodeInputGradientArrays} from './tape_types';
 import * as tape_util from './tape_util';
 import {ScopeResult, ScopeResultImmediate} from './tape_util';
 
@@ -34,6 +37,8 @@ export class BackendEngine {
 
   private activeTape: Tape;
   private gradientScopeCount = 0;
+
+  private customGradientDepth = 0;
 
   // Keep NDArrays that parallel the tapes.
   private activeScope: ScopeState;
@@ -76,7 +81,7 @@ export class BackendEngine {
       util.checkForNaN(vals, result.dtype, name);
     }
 
-    if (this.activeTape != null) {
+    if (this.activeTape != null && this.customGradientDepth === 0) {
       config = tape_util.stripUndefinedInputsFromInputConfig(config) as C;
 
       const evaluatedNode: KernelNode = {
@@ -91,6 +96,35 @@ export class BackendEngine {
       this.activeTape.push(evaluatedNode);
     }
 
+    return result;
+  }
+
+  customGradient<T extends NDArray>(
+      name: string, f: () => T, inputs: NamedArrayMap,
+      gradient: (dy: T, y: T) => TapeNodeInputGradientArrays): T {
+    this.customGradientDepth++;
+
+    const gradientsMode = false;
+    const result = this.scope('customGradient', () => {
+      return f();
+    }, gradientsMode);
+
+    if (this.activeTape != null && this.customGradientDepth === 0) {
+      const inputAndArgs =
+          tape_util.stripUndefinedInputsFromInputConfig({inputs});
+
+      const evaluatedNode: TapeNode<T> = {
+        id: this.nextTapeNodeId++,
+        type: 'customGradient',
+        name,
+        inputAndArgs,
+        output: result,
+        gradient
+      };
+      this.activeTape.push(evaluatedNode);
+    }
+
+    this.customGradientDepth--;
     return result;
   }
 
@@ -119,16 +153,34 @@ export class BackendEngine {
     }
   }
 
-  private gradientWrt(y: Scalar, xs: NDArray[]): NDArray[] {
+  vjp<T extends NDArray>(f: () => T, xs: NDArray[], dy: T): NDArray[] {
+    const gradientsMode = true;
+    return this.scope('vjp', () => {
+      const y = f();
+      if (!util.arraysEqual(y.shape, dy.shape)) {
+        throw new Error(
+            `Cannot compute vector jacobian product, ` +
+            `y shape (${y.shape}) does not match dy shape (${dy.shape}).`);
+      }
+      if (y.dtype !== dy.dtype) {
+        throw new Error(
+            `Cannot compute vector jacobian product, ` +
+            `y dtype (${y.dtype}) does not match dy dtype (${dy.dtype}).`);
+      }
+      return this.gradientWrt(y, xs, dy);
+    }, gradientsMode);
+  }
+
+  private gradientWrt<T extends NDArray>(y: T, xs: NDArray[], dy?: T):
+      NDArray[] {
     // Filter out the nodes that don't connect x => y.
     const filteredTape = tape_util.getFilteredNodesXToY(this.activeTape, xs, y);
     if (filteredTape.length === 0) {
       throw new Error(`Cannot compute gradient: y is not a function of xs.`);
     }
 
-    // Seed the gradient of dy to be 1.
     const arrayAccumulatedGradientMap: {[ndarrayId: number]: NDArray} = {};
-    arrayAccumulatedGradientMap[y.id] = Scalar.new(1);
+    arrayAccumulatedGradientMap[y.id] = dy == null ? Scalar.new(1) : dy;
 
     // Backprop gradients through the filtered nodes.
     tape_util.backpropagateGradients(arrayAccumulatedGradientMap, filteredTape);
@@ -208,7 +260,7 @@ export class BackendEngine {
         continue;
       }
 
-      if (this.activeTape != null) {
+      if (this.activeTape != null && this.customGradientDepth === 0) {
         arraysToTrackInParent.push(ndarray);
       } else {
         ndarray.dispose();
--- a/src/math/backends/backend_engine.ts
+++ b/src/math/backends/backend_engine.ts
@@ -99,14 +99,20 @@ export class BackendEngine {
     return result;
   }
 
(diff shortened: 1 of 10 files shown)


dsmilkov:
:lgtm_strong: Amazing work!

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/math/math.ts, line 840 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cJtmJtGztuz64xxH4:-L1cJtmJtGztuz64xxH5:bel1qox) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L840)):*
> ```XML
>    *
>    * logits and labels must have the same shape, e.g. [batch_size, num_classes]
>    * and the same dtype.
> ```

missing `@param labels ...`

---

*[src/math/math.ts, line 859 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cN-yn-GhAfIr-Bie1:-L1cN-yn-GhAfIr-Bie2:b-tp5g4j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L859)):*
> ```XML
>     }
> 
>     // Use a custom gradient for numerical stability.
> ```

wrap in `this.executeOp('softmaxCrossEntropyWithLogits')` for hierarchical timing/debugging. OR have customGradient call `executeOp`, in which case remove the `executeOp('softmax', ...)` in `math.softmax()`

---

*[src/math/math.ts, line 863 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cKXwcDZNb22PMuN27:-L1cKXwcDZNb22PMuN28:br0ym87) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L863)):*
> Scalar.new(1e-5)

I wonder if we should extract these Scalars throughout Math as private members so we don't re-upload at every step in the training process.

A much better solution, since this re-upload inefficiency is an implementation detail of the webgl backend (other backends might not care), might be:
Optimize our shader compiler to use uniforms instead of sampler2d whenever the input lives in cpu and is scalar or array1d with size <= 4 :) This is not highest priority obviously. WDYT? I can file an issue if its worth pursing later.

---

*[src/math/math_test.ts, line 295 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cOSYDsdsUABzGFtQz:-L1cOSYDsdsUABzGFtR-:b-wl67tl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math_test.ts#L295)):*
> ```XML
>       {'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 2},
>       {'WEBGL_FLOAT_TEXTURE_ENABLED': false, 'WEBGL_VERSION': 1}
>     ]);
> ```

if not too much trouble, add second order vjp test, of `math.gradient` wrapped in a `math.vjp`. Polynomial function might be good (y = x^3, element-wise power, so y is not scalar)

---

*[src/math/math_test.ts, line 346 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cO4auAjPgIKesuv92:-L1cO4auAjPgIKesuv93:b-gksgat) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math_test.ts#L346)):*
> 1e-1

this used to not need 1e-1. What changed now? Here and the tests below

---

*[src/math/matmul_test.ts, line 248 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cQqTBzaQH_F7ADIjk:-L1cQqTBzaQH_F7ADIjl:b-fjuj9v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4b157c7537725ae4596d1cd39f4e21dff2aa798f/src/math/matmul_test.ts#L248)):*
> ```XML
> };
> 
> // TODO(nsthorat): fix the precision for backprop.
> ```

optional: maybe keep the TODO

---

*[src/math/backends/backend_engine.ts, line 84 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cTe_XHXE8WprvXrDB:-L1cTe_XHXE8WprvXrDC:b-sutt3v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L84)):*
> this.customGradientDepth === 0

Thoughts for future - we can cross this bridge later: I wonder how customGradient can interoperate with higher-order gradients?

---

*[src/math/backends/backend_engine.ts, line 102 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cHpHOkQ78eiopIM8q:-L1cHpHOkQ78eiopIM8r:b-1fvzrb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L102)):*
> ```XML
>   }
> 
>   customGradient<G extends DataType, T extends NDArray<G>>(
> ```

No need to be done in this PR, but we should expose customGradient to the user, via `NDArrayMath`.

Either way, can you add a jsdoc to this method. We can move that jsdoc to math when we expose it there.

---

*[src/math/backends/backend_engine.ts, line 103 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cSiTbeJpc8cX8EDS3:-L1cSiTbeJpc8cX8EDS4:brbyikt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L103)):*
> name: string,

how important is it to have a name? wondering if name can be the last argument, and optional.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-:-L1cTvueZQtvgNaS3_7W:b-tbz4hy)*

dsmilkov:
Review status: all files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

*[src/math/math_test.ts, line 295 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cOSYDsdsUABzGFtQz:-L1cUUIsLy4dVpLffm_S:b-92qqiv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math_test.ts#L295)):*

if not too much trouble, add second order vjp test, of `math.gradient` wrapped in a `math.vjp`. Polynomial function might be good (y = x^3, element-wise power, so y is not scalar)
</blockquote></details>

Or vjp wrapped in a vjp

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486)*

nsthorat:
Review status: 4 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 840 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cJtmJtGztuz64xxH4:-L1cWC77oaALNx6dWWOy:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L840)):*

missing `@param labels ...`
</blockquote></details>

Done.

---

*[src/math/math.ts, line 859 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cN-yn-GhAfIr-Bie1:-L1cWhn9Jz1VPVTmon4x:b5k6scg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L859)):*

wrap in `this.executeOp('softmaxCrossEntropyWithLogits')` for hierarchical timing/debugging. OR have customGradient call `executeOp`, in which case remove the `executeOp('softmax', ...)` in `math.softmax()`
</blockquote></details>

Let's keep it separate for now. Done.

---

*[src/math/math.ts, line 863 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cKXwcDZNb22PMuN27:-L1cWmAMQnVUjjwoQ6fy:b-b58a4t) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math.ts#L863)):*

> Scalar.new(1e-5)

I wonder if we should extract these Scalars throughout Math as private members so we don't re-upload at every step in the training process.

A much better solution, since this re-upload inefficiency is an implementation detail of the webgl backend (other backends might not care), might be:
Optimize our shader compiler to use uniforms instead of sampler2d whenever the input lives in cpu and is scalar or array1d with size <= 4 :) This is not highest priority obviously. WDYT? I can file an issue if its worth pursing later.
</blockquote></details>

Yeah, I was thinking about that too. Another idea would be to allow common operations like this to take a number|NDArray. In the case of number, we either 1) splice it into the shader program or 2) upload it as a uniform directly. This allows us to be pretty explicit about what is happening and when.

Yeah agreed that this can wait a little :) Here is the issue I filed: https://github.com/PAIR-code/deeplearnjs/issues/494

---

*[src/math/math_test.ts, line 295 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cOSYDsdsUABzGFtQz:-L1cdnaSNrUuSKPQ_01_:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math_test.ts#L295)):*

Or vjp wrapped in a vjp
</blockquote></details>

Done.

---

*[src/math/math_test.ts, line 346 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cO4auAjPgIKesuv92:-L1cYjj3sjEyXZ_mu1wY:b-yuwoqi) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/math_test.ts#L346)):*

> 1e-1

this used to not need 1e-1. What changed now? Here and the tests below
</blockquote></details>

Because we weren't multiplying by dy before :(. Now that we multiply by dy, the error seems to accumulate even more.

I think we need to really revisit precision on byte packing with dynamic bounds quantization (or using a log-scale quantization)

---

*[src/math/matmul_test.ts, line 248 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cQqTBzaQH_F7ADIjk:-L1cdloVhM5yvqeUMpy3:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4b157c7537725ae4596d1cd39f4e21dff2aa798f/src/math/matmul_test.ts#L248)):*

optional: maybe keep the TODO
</blockquote></details>

Done.

---

*[src/math/backends/backend_engine.ts, line 84 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cTe_XHXE8WprvXrDB:-L1cd3v8VChmYTFx5z0L:b-6pwqsc) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L84)):*

> this.customGradientDepth === 0

Thoughts for future - we can cross this bridge later: I wonder how customGradient can interoperate with higher-order gradients?
</blockquote></details>

Added some tests for second order gradient through a custom gradient. Spoke offline, but the only issue now is if we reuse computation from forward mode in the custom gradient during a second order gradient. For now, we'll ignore this (it throws an error), but we can revisit. 

---

*[src/math/backends/backend_engine.ts, line 102 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cHpHOkQ78eiopIM8q:-L1c_IBoVi4vL9K8h7tj:bq5ew8q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L102)):*

No need to be done in this PR, but we should expose customGradient to the user, via `NDArrayMath`.

Either way, can you add a jsdoc to this method. We can move that jsdoc to math when we expose it there.
</blockquote></details>

Added this to `NDArrayMath` and added a jsdoc.

---

*[src/math/backends/backend_engine.ts, line 103 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-L1cSiTbeJpc8cX8EDS3:-L1c_3-a8Iah4Dez-vkf:b-6ecgln) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fd434ca4761af2b0d0d495850ff2600933a9d610/src/math/backends/backend_engine.ts#L103)):*

> name: string,

how important is it to have a name? wondering if name can be the last argument, and optional.
</blockquote></details>

It's now the last name and optional on math, not optional on the backend engine.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486)*

nsthorat:
Thanks for the awesome review!

---

Review status: 4 of 10 files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/486#-:-L1djhrbLGh36siVvJup:bx5yqhg)*

PR #489, opened 2017-12-29
squeezenet image resize to [227, 227]

MariasStory:
The squeezenet model needed the following dimensions of the image [227, 227]. As discussed in https://github.com/PAIR-code/deeplearnjs/issues/469#issuecomment-354444559, I used resizeBilinear3D to resize the image 3D array.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/489)

### DIFF
--- a/models/squeezenet/demo.html
+++ b/models/squeezenet/demo.html
@@ -21,7 +21,8 @@
 
     resultElement.innerText = 'Predicting...';
 
-    const pixels = dl.Array3D.fromPixels(cat);
+    //make a 3D array from the image and resize it to standart dimentions
+    const pixels = math.resizeBilinear3D(dl.Array3D.fromPixels(cat),[227, 227]);
 
     const result = squeezeNet.predict(pixels);
 


nsthorat:

---

*[models/squeezenet/demo.html, line 25 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/489#-L1eB4RMiHpDyZuDcnwb:-L1eB4RMiHpDyZuDcnwc:b-rrzdm9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/cfc70287ef5eb66e53f1afd49caa2c5e7e3e8f56/models/squeezenet/demo.html#L25)):*
> ```HTML
> 
>     //make a 3D array from the image and resize it to standart dimentions
>     const pixels = math.resizeBilinear3D(dl.Array3D.fromPixels(cat),[227, 227]);
> ```

it'd be nice to put this in the model itself, instead of just in the demo.

First, check whether the dimensions aren't 227 before calling bilinear resize.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/489)*

MariasStory:
There is a problem with this approach:
Nonstandard image dimensions are deformed and cannot be recognized.
A somewhat better approach is to cut the square aria out of the image and then resize.

nsthorat:
Sorry for the delay, thinking about this more I think this was the right decision to do it in demo.html (another person sent a PR doing this). I merged that PR, so closing this. Thank you!

PR #492, opened 2017-12-30
Force matching dtypes for ops that take 2 inputs, to be compatible with TensorFlow

dsmilkov:
Force matching dtypes both at run-time and compile-time for ops that take 2 inputs, to be fully compatible with TensorFlow. Ops that require both inputs to be of same dtype:
- `maximum`
- `minimum`
- `add`
- `subtract`
- `equal`
- `multiply`

Also cleanup: always use the letter `D` for `D extends DataType`

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/492)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,8 @@
     "**/bower_components": true,
     "coverage/": true,
     "dist/": true,
-    "**/bundle.js": true
+    "**/bundle.js": true,
+    "**/yarn.lock": true
   },
   "tslint.enable": true,
   "tslint.run": "onType",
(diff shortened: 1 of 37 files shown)


nsthorat:
:lgtm_strong:

Nice cleanup. It might be nice to add quick unit tests for the 6 ops that now throw if the dtypes dont match at runtime.

Unfortunately my PR added non-D DTypes, apologies. Only two places you'll have to update: math.customGradient & backendEngine.customGradient

---

Reviewed 10 of 10 files at r1.
Review status: 7 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/492#-:-L1eE0eqFtV0aXNqsYab:bdvea79)*

PR #499, opened 2018-01-01
Make NDArray's id unique. Move reshape() and asType() to Math.

### LINKED ISSUE #484: Move reshape(), asType() to Math.
When computing gradients, we need to be able to backprop through reshape(), and asType().

This will likely require:
* Allowing multiple NDArrays to point to the same data (each with different logical shapes).
* Doing data reference counting for disposal of data.
* Adding a shader program for asType that does a GPU-cast (floor, etc)

dsmilkov:
Now each `NDArray` has a unique `id`, but can share its `dataId` with other arrays.

- `const b = a.reshape(newShape)`
  - `b` has a different `id`, but the same `dataId` as `a`.
- `const b = a.asType(newType)`
   - `b` has a different `id`, and
   - same `dataId` if casting to a wider type (no loss of precision)
   - different `dataId` if casting to a narrower type, in which case we need to trim the data (e.g. flooring when doing to `int32`)

Also added:
 - `math.notEqual()`, analogous to `tf.not_equal()`.
 - `math.cast()` analogous to `tf.cast()`
 - `backend.int()` since it's a non-trivial operation: positive numbers are floored, negative numbers are ceiled. This is not exposed to the user. `math.cast()` uses it internally.

Fixes #484 

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/499)

### DIFF
--- a/src/math/backends/backend.ts
+++ b/src/math/backends/backend.ts
@@ -18,7 +18,7 @@
 
 import {Conv2DInfo} from '../conv_util';
 // tslint:disable-next-line:max-line-length
-import {Array1D, Array2D, Array3D, Array4D, DataType, DataTypeMap, NDArray} from '../ndarray';
+import {Array1D, Array2D, Array3D, Array4D, DataType, DataTypeMap, NDArray, Rank} from '../ndarray';
 import {SumTypes} from '../types';
 import {MatrixOrientation} from './types/matmul';
 
@@ -75,7 +75,7 @@ export interface MathBackend extends NDArrayStorage {
   argMax(x: NDArray, axes: number[]): NDArray<'int32'>;
 
   equal(a: NDArray, b: NDArray): NDArray<'bool'>;
-
+  notEqual(a: NDArray, b: NDArray): NDArray<'bool'>;
   topKValues<D extends DataType, T extends NDArray<D>>(x: T, k: number):
       Array1D<D>;
   topKIndices(x: NDArray, k: number): Array1D<'int32'>;
@@ -103,6 +103,7 @@ export interface MathBackend extends NDArrayStorage {
   leakyRelu<T extends NDArray>(x: T, alpha: number): T;
   prelu<T extends NDArray>(x: T, alpha: T): T;
   preluDer<T extends NDArray>(x: T, alpha: T): T;
+  int<R extends Rank>(x: NDArray<DataType, R>): NDArray<'int32', R>;
 
   clip<T extends NDArray>(x: T, min: number, max: number): T;
 
--- a/src/math/backends/backend.ts
+++ b/src/math/backends/backend.ts
@@ -23,13 +23,13 @@ import {SumTypes} from '../types';
 import {MatrixOrientation} from './types/matmul';
 
 export interface NDArrayStorage {
-  read<D extends DataType>(id: number): Promise<DataTypeMap[D]>;
-  readSync<D extends DataType>(id: number): DataTypeMap[D];
-  disposeData(id: number): void;
+  read<D extends DataType>(dataId: number): Promise<DataTypeMap[D]>;
+  readSync<D extends DataType>(dataId: number): DataTypeMap[D];
+  disposeData(dataId: number): void;
   write<D extends DataType>(
-      id: number, values: DataTypeMap[D], dtype: D, shape: number[]): void;
+      dataId: number, values: DataTypeMap[D], dtype: D, shape: number[]): void;
   writePixels(
-      id: number,
+      dataId: number,
       pixels: ImageData|HTMLImageElement|HTMLCanvasElement|HTMLVideoElement,
       numChannels: number): void;
   time(query: () => NDArray): Promise<number>;
--- a/src/math/backends/backend_cpu.ts
+++ b/src/math/backends/backend_cpu.ts
@@ -24,7 +24,7 @@ import * as concat_util from '../concat_util';
 import {Conv2DInfo} from '../conv_util';
 import {NDArrayMath} from '../math';
 // tslint:disable-next-line:max-line-length
-import {Array1D, Array2D, Array3D, Array4D, DataType, DataTypeMap, NDArray, Scalar} from '../ndarray';
+import {Array1D, Array2D, Array3D, Array4D, DataType, DataTypeMap, NDArray, Rank, Scalar} from '../ndarray';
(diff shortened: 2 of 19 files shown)


nsthorat:
:lgtm_strong:

Nice work! Few comments inline, mostly minor.

---

Reviewed 1 of 11 files at r1, 8 of 16 files at r2.
Review status: 9 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/math.ts, line 338 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yYMBGT5djDOVrh3qZ:-L1yYMBGT5djDOVrh3q_:biwluyh) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L338)):*
> ```XML
>       x: NDArray<D>, newShape: number[]): T {
>     newShape = util.inferFromImplicitShape(newShape, x.size);
>     if (util.arraysEqual(x.shape, newShape)) {
> ```

Do we want to do this? We might have problems here:

`const c = math.scope(() => {
  const a = Scalar.new(1);
  return a.reshape([]);
})

// c is disposed`

maybe we want to always do the reshape.

---

*[src/math/math.ts, line 359 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZOV4t-5WUCF56QrW:-L1yZOV4t-5WUCF56QrX:bkbrt64) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L359)):*
> ```XML
>   cast<D extends DataType, R extends Rank>(
>       x: NDArray<DataType, R>, newDType: D): RankMap<D>[R] {
>     if (x.dtype === newDType) {
> ```

see comment above here as well.

---

*[src/math/math.ts, line 364 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZQtNVySvZCHvIwqQ:-L1yZQtNVySvZCHvIwqR:b8bzp61) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L364)):*
> ```XML
>     }
>     const grad = (dy: NDArray, y: NDArray) => {
>       return {x: () => dy};
> ```

here too, I think we need to use a separate NDArray (make all gradients functional)

---

*[src/math/math.ts, line 1234 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZamFR16WOoe_yYKz:-L1yZamFR16WOoe_yYL-:b-inzk57) ([raw file](https://github.com/pair-code/deeplearnjs/blob/9d6765d2a88739c598e4d74b760c3cd000ba458b/src/math/math.ts#L1234)):*
> ```XML
>   /**
>    * Computes floor of input NDArray element-wise. y = floor(x)
>    * TODO(nsthorat): Make this return an int32 when we add rank as a generic.
> ```

shouldnt this now return 'int32'?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-:-L1z7mT0ktrDyU0G3fdH:b-kqdwoy)*

nsthorat:
One more high level question, have you seen this?

https://www.tensorflow.org/versions/r0.12/api_docs/python/array_ops/casting

Might be nice to have "toFloat" and "toInt32" helper methods.

---

Review status: 9 of 19 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-:-L1z9VnVKY3cE0caB3h9:bt3hgb2)*

dsmilkov:
Review status: 8 of 19 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[src/math/math.ts, line 338 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yYMBGT5djDOVrh3qZ:-L21AroRfu-XMICZovcZ:b-dra9h8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L338)):*

Do we want to do this? We might have problems here:

`const c = math.scope(() => {
  const a = Scalar.new(1);
  return a.reshape([]);
})

// c is disposed`

maybe we want to always do the reshape.
</blockquote></details>

Done. Reshape is functional now

---

*[src/math/math.ts, line 359 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZOV4t-5WUCF56QrW:-L21Axm9nkAp5BTgmdzA:b-169wuk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L359)):*

see comment above here as well.
</blockquote></details>

Done. asType is functional now

---

*[src/math/math.ts, line 364 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZQtNVySvZCHvIwqQ:-L21AtwAUuPmx5SRI0eh:b-b6i8tg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/253eb4e3e8614cfe4e071c08d9ff5c4423f09a9a/src/math/math.ts#L364)):*

here too, I think we need to use a separate NDArray (make all gradients functional)
</blockquote></details>

Done. functional now.

---

*[src/math/math.ts, line 1234 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499#-L1yZamFR16WOoe_yYKz:-L20jNKHgSfAdbrFax1b:b-lxksqd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/9d6765d2a88739c598e4d74b760c3cd000ba458b/src/math/math.ts#L1234)):*

shouldnt this now return 'int32'?
</blockquote></details>

[`tf.floor`](https://www.tensorflow.org/api_docs/python/tf/floor) returns the same dtype as the input `x`. So should we and we already do this, so I removed the TODO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/499)*

PR #505, opened 2018-01-03
Fix a bug in webgl storage and fix demos

dsmilkov:
- Fix a bug in webgl where the user can't directly interact with the backend like in `imagenet` and `cppn` demos (e.g. calling `backend.getTexture(id)` right after calling `NDArray.make()`)
- Make the backend storage mehanism clearer (separate registration from writing).
- Small changes of hyper-params in the complementary colors demo, to converge faster
- To be more compatible with TF, `add/sub/multiply/divide` now require both inputs to be of the same dtype. Update outdated demos and squeezenet model.

After this PR is in, I will:
- release a new `deeplearn` package
- release a new `deeplearn-squeezenet` package
- update `demos/package.json` to depend on the newest libraries

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/505)

### DIFF
--- a/demos/complementary-color-prediction/complementary-color-prediction.ts
+++ b/demos/complementary-color-prediction/complementary-color-prediction.ts
@@ -16,7 +16,7 @@
  */
 
 // tslint:disable-next-line:max-line-length
-import {Array1D, CostReduction, ENV, FeedEntry, Graph, InCPUMemoryShuffledInputProviderBuilder, NDArrayMath, Session, SGDOptimizer, Tensor} from 'deeplearn';
+import {Array1D, CostReduction, ENV, FeedEntry, Graph, InCPUMemoryShuffledInputProviderBuilder, Session, SGDOptimizer, Tensor} from 'deeplearn';
 
 class ComplementaryColorModel {
   // Runs training.
@@ -126,7 +126,7 @@ class ComplementaryColorModel {
 
   predict(rgbColor: number[]): number[] {
     let complementColor: number[] = [];
-    this.math.scope((keep, track) => {
+    this.math.scope(() => {
       const mapping = [{
         tensor: this.inputTensor,
         data: Array1D.new(this.normalizeColor(rgbColor)),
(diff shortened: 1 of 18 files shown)


nsthorat:
:lgtm_strong:

Nice work!

---

Reviewed 4 of 17 files at r1.
Review status: 4 of 17 files reviewed at latest revision, 3 unresolved discussions.

---

*[demos/imagenet/imagenet.ts, line 80 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/505#-L1zAq9zdLCgA8v1SUMj:-L1zAq9zdLCgA8v1SUMk:b-t24gwj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/37f308a79a2db9ca2f00afc253f459f000730c0f/demos/imagenet/imagenet.ts#L80)):*
> ```XML
>     this.webcamVideoElement =
>         this.querySelector('#webcamVideo') as HTMLVideoElement;
>     this.videoCanvas = this.querySelector('#video-canvas') as HTMLCanvasElement;
> ```

any reason this needs to be in the DOM?

---

*[demos/imagenet/imagenet.ts, line 194 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/505#-L1zAcw97D8HJV1hLOxl:-L1zAcwAcT095LLXQNvx:b-mxup3t) ([raw file](https://github.com/pair-code/deeplearnjs/blob/37f308a79a2db9ca2f00afc253f459f000730c0f/demos/imagenet/imagenet.ts#L194)):*
> ```XML
>       const element =
>           isWebcam ? this.webcamVideoElement : this.staticImgElement;
>       this.videoCtx.drawImage(element, 0, 0, element.width, element.height);
> ```

quick question - this is fine for now, but *everybody* will run into this. What would you think about us doing this for the user if given a video tag?

We can hold onto a canvas that's not attached to the dom

---

*[src/math/backends/backend_webgl.ts, line 220 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/505#-L1zC3sJgi5Ai_nGYlqe:-L1zC3sJgi5Ai_nGYlqf:buksj7k) ([raw file](https://github.com/pair-code/deeplearnjs/blob/37f308a79a2db9ca2f00afc253f459f000730c0f/src/math/backends/backend_webgl.ts#L220)):*
> ```XML
>   clone<D extends DataType, T extends NDArray<D>>(x: T): T {
>     this.throwIfNoData(x.id);
>     this.uploadToGPU(x.id);
> ```

can you explain this to me since this is non obvious (or maybe I'm just an idiot)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/505#-:-L1zCK2fqEKQRNUfDaD6:b-rtmsva)*

PR #518, opened 2018-01-06
add Jasmine fit and xit support to math tests

chaosmail:
Hi guys,

Using `TestMode.FOCUS` as a third argument on custom `it` functions, we can get now the same behavior of `fit` and `xit` in the tests (docs here: https://jasmine.github.io/2.1/focused_specs.html).

I found this super useful while implementing `LRN`.

Best,
Christoph

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/518)


chaosmail:
Ooh, seems that `Calls to 'fit' and 'xit' are not allowed.` - which makes sense for tests.

dsmilkov:
:lgtm_strong: Pretty useful. FYI, what we've been doing is:
`yarn test --browsers=Chrome --grep='substring that matches the test desc'`

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/518#-:-L2GV9ZmGB4s2_HjREDn:bp2o0wa)*

dsmilkov:
Hey @chaosmail,
 
> Ooh, seems that Calls to 'fit' and 'xit' are not allowed. - which makes sense for tests.

Can you clarify what you meant by that? Is this PR still functional? Thanks!

chaosmail:
it is functional; I use this code a lot (usually cherry pick the commit, and then drop it when i push the changes).

However, TSLint throws an error that `fit` and `xit` should not be used (in a test when committed) - which makes sense when you don't want to oversee a `fit` call somewhere in the tests - which would only execute this test in CI.

```
$ tslint -p . -t verbose
ERROR: (ban) /home/travis/build/PAIR-code/deeplearnjs/src/test_util.ts[330, 7]: Calls to 'xit' are not allowed.
ERROR: (ban) /home/travis/build/PAIR-code/deeplearnjs/src/test_util.ts[333, 7]: Calls to 'fit' are not allowed.
```

dsmilkov:
Ah, it's a tslint feature, I see! Nice!

dsmilkov:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[src/test_util.ts, line 330 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/518#-L2GtOkTs-gWDfi1wmWD:-L2GtOkUPITH8DwNT2xg:bjdxql9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2bebb8321eac1232a810397ffd6b81e201ff4b13/src/test_util.ts#L330)):*
> ```XML
>   switch (testMode) {
>     case TestMode.EXCLUDE:
>       xit(name, cb);
> ```

add `// tslint:disable-next-line:ban` above this line. do the same with fit. This will overwrite tslint

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/518)*

chaosmail:
Sure, this will make the commit pass Travis; I pushed the cahnge. However, now we loose the safety of someone accidentally leaving TestMode.FOCUS in the code in the PR.

We could add a simple script to Travis to avoid focused tests to be checked in, something like:
```sh
FOCUSED_TESTS=$(grep -rnw src -e 'TestMode.FOCUS' | grep -v 'src/test_util.ts')
if $FOCUSED_TESTS; then
  echo "Error. Found focused tests in PR."
  echo $FOCUSED_TESTS
  exit 1
fi
```

Let me know what you think! Running only a single test/a couple of tests greatly speeds up development.

dsmilkov:
Yeah, hopefully we will catch that with code reviews. Alternatively we can pass fit to the user, just like we pass it. From the user point of view:
```
const gpuTests: MathTests = (it, fit) => {
    fit('debug mode does not error when no nans', math => {
       ....
    });

This way:
 - we don't depart away from jasmine's API - good for people that are already familiar with it
 - lint will still fail

What do you think?

**Comment by chaosmail** [comment]
love the idea! will modify the code accordingly

**Comment by chaosmail** [comment]
Hi @dsmilkov, I implemented it as we discussed it and it is really much nicer.

Now you can go from

```ts
// math.someTest
{
  const tests: MathTests = it => {
    it('do something', math => {
      // custom test
    });

  test_util.describeMathCPU('someTest', [tests]);
  test_util.describeMathGPU('someTest', [tests], [
    {'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 1},
    {'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 2},
    {'WEBGL_FLOAT_TEXTURE_ENABLED': false, 'WEBGL_VERSION': 1}
  ]);
}
```

to

```ts
// math.someTest
{
  const tests: MathTests = (it, fit, xit) => {
   // run only this test, TSLint will complain and CI will fail
   fit('do something', math => {
      // custom test
    });

   // exclude this test, TSLint will complain and CI will fail
   xit('do something', math => {
      // custom test
    });

  test_util.describeMathCPU('someTest', [tests]);
  test_util.describeMathGPU('someTest', [tests], [
    {'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 1},
    {'WEBGL_FLOAT_TEXTURE_ENABLED': true, 'WEBGL_VERSION': 2},
    {'WEBGL_FLOAT_TEXTURE_ENABLED': false, 'WEBGL_VERSION': 1}
  ]);
}
```
even with keeping the npm lint errors. This is speeding up development quite a lot with tests running in the background.

I thought about documenting this feature somewhere. Shall I create a new page `docs/development.md` to start some development guidelines? Or should I put this into a new PR?

Best,
Christoph

dsmilkov:
We already have a [Development section](https://github.com/PAIR-code/deeplearnjs#development) in our `README.md`, but I like the idea to move it in a separate doc and update it with the karma updates you made. I looked at a few libs and they have `CONTRIBUTING.md` file at the root of the project.

If you write that doc, would be great to point to it from our `README.md`.

dsmilkov:
:lgtm_strong: Very nice! No remarks. Will merge now

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/518#-:-L2mHC8Sis4cwEqvdb3W:b-kvgkdj)*

PR #519, opened 2018-01-07
Fix bug in environment.reset() - we didn't dispose custom backends

dsmilkov:
- `ENV.reset()` wasn't disposing the custom backends that were registered. Also now we guarantee that there is one un-disposable registry of backends, the one added during library init. So if the user calls `ENV.reset()` several times, it will only dispose custom backends, not the initial ones.
- Fix memory leak with Variables. `variable(a: NDArray)` now will point to `a`'s data, and dispose `a`.
- Fix small bug with `SGDOptimizer.setLearningRate()`

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/519)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/519#-:-L2Dn9cFe77fV33uwqcV:b-jbsf8z)*

PR #532, opened 2018-01-09
add Squeeze op for NDArray

pyu10055:
same as the tf api squeeze
https://www.tensorflow.org/api_docs/python/tf/squeeze

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/532)

### DIFF
--- a/src/math/ndarray.ts
+++ b/src/math/ndarray.ts
@@ -219,8 +219,8 @@ export class NDArray<D extends DataType = DataType, R extends Rank = Rank> {
    * squeezes the dimensions listed. The dimension index starts at 0. It is an
    * error to squeeze a dimension that is not 1.
    */
-  squeeze(axis?: number[]): NDArray {
-    return this.reshape(util.squeezeShape(this.shape, axis).newShape);
+  squeeze<T extends NDArray<D>>(axis?: number[]): T {
+    return this.reshape(util.squeezeShape(this.shape, axis).newShape) as T;
   }
 
   /** Flatten a NDArray to a 1D array. */
--- a/src/math/ndarray.ts
+++ b/src/math/ndarray.ts
@@ -214,6 +214,15 @@ export class NDArray<D extends DataType = DataType, R extends Rank = Rank> {
     return this.math.reshape(this, newShape);
   }
 
+  /**
+   * @param axis An optional list of number. If specified, only
+   * squeezes the dimensions listed. The dimension index starts at 0. It is an
+   * error to squeeze a dimension that is not 1.
+   */
+  squeeze(axis?: number[]): NDArray {
+    return this.reshape(util.squeezeShape(this.shape, axis).newShape);
+  }
+
   /** Flatten a NDArray to a 1D array. */
   flatten(): Array1D<D> {
     this.throwIfDisposed();
--- a/src/math/ndarray_test.ts
+++ b/src/math/ndarray_test.ts
@@ -1449,23 +1449,25 @@ const testsAsType: MathTests = it => {
     test_util.expectArraysClose(a, [2.4]);
   });
 };
+
 const testSqueeze: MathTests = it => {
   it('squeeze no axis', () => {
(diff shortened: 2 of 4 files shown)


dsmilkov:
:lgtm_strong: This is great. Couple of comments, but nothing major. Feel free to submit after the comments!

---

Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/util.ts, line 316 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532#-L2VcMOfWw09IEsw2xjL:-L2VcMOfWw09IEsw2xjM:b8dq961) ([raw file](https://github.com/pair-code/deeplearnjs/blob/305ce55a793388541c9c63af3d095debfdad2ddc/src/util.ts#L316)):*
> ```XML
>   let j = 0;
>   for (let i = 0; i < shape.length; ++i) {
>     if (axis) {
> ```

we practice `s/if (axis)/if (axis != null)` throughout the code base, to be safe, since occasionally we miss on `0`

---

*[src/math/ndarray.ts, line 222 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532#-L2VevmaF5hlcOyhQCql:-L2VevmaF5hlcOyhQCqm:b8e8td7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/305ce55a793388541c9c63af3d095debfdad2ddc/src/math/ndarray.ts#L222)):*
> ```XML
>    * error to squeeze a dimension that is not 1.
>    */
>   squeeze(axis?: number[]): NDArray {
> ```

we can improve the typing inference a bit:
```
squeeze<T extends NDArray<D>>(axis?: number[]): T {
  return this.reshape(....) as T; 
}
```

this will keep the dtype of the original array and moreover try to infer the rank of the output, if the output is used in another method. For example:

if user calls `matmul(a.squeeze(), b)`, the output of `a.squeeze()` will be inferred as `Array2D` since matmul wants `Array2D`. This reduces the amount of extraneous typing the user has to make, though if the rank of the output is not 2, at run-time, we will throw an Error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532#-:-L2Vg00DwqCvscTxaXGu:bli10x9)*

pyu10055:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[src/util.ts, line 316 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532#-L2VcMOfWw09IEsw2xjL:-L2X-MsN8SrgPco5njuA:b-n915p1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/305ce55a793388541c9c63af3d095debfdad2ddc/src/util.ts#L316)):*

we practice `s/if (axis)/if (axis != null)` throughout the code base, to be safe, since occasionally we miss on `0`
</blockquote></details>

good call, I checked against undefined instead of null.

---

*[src/math/ndarray.ts, line 222 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532#-L2VevmaF5hlcOyhQCql:-L2X-U_UgjarYVmizLcV:bd5l6mr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/305ce55a793388541c9c63af3d095debfdad2ddc/src/math/ndarray.ts#L222)):*

we can improve the typing inference a bit:
```
squeeze<T extends NDArray<D>>(axis?: number[]): T {
  return this.reshape(....) as T; 
}
```

this will keep the dtype of the original array and moreover try to infer the rank of the output, if the output is used in another method. For example:

if user calls `matmul(a.squeeze(), b)`, the output of `a.squeeze()` will be inferred as `Array2D` since matmul wants `Array2D`. This reduces the amount of extraneous typing the user has to make, though if the rank of the output is not 2, at run-time, we will throw an Error.
</blockquote></details>

thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/532)*

PR #536, opened 2018-01-10
import Tf model proto file and run inference with deeplearn

pyu10055:
This is a prototype for loading tf frozen model in proto format and
using deeplearn.js as backend for inference.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/536)


dsmilkov:
Hi,

Just to make sure we are on the same page, we've decided to not merge this to the dljs repo, but develop it in a separate repo, right? Having the PR open for now is good so we can look at the code.

Does that sounds right?

pyu10055:
Thanks Daniel, and yes, I think that is the decision. But we need to figure out which organization to associate this repo with.

nkreeger:
Super-high level pass. Some really good stuff that we can iron out - nice work!

---

Review status: 0 of 31 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[models/tensorflow/src/node.ts, line 64 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/536#-L33b2f-vIwa9bfaj10F:-L33b2f-vIwa9bfaj10G:brmeb15) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2df0d6653967f2ff5dcac5a6753a252f5fc9acd1/models/tensorflow/src/node.ts#L64)):*
> export function performMathOp
Does it make sense to have an actual Node class? Might make the design a little more structured going forward (not as many in-params, etc).

---

*[models/tensorflow/src/node.ts, line 73 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/536#-L33apYJ8IqeGSl3aTGB:-L33apYJ8IqeGSl3aTGC:be9wyrf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2df0d6653967f2ff5dcac5a6753a252f5fc9acd1/models/tensorflow/src/node.ts#L73)):*
> if (inputs[0].dtype !== inputs[1].dtype) {
>         inputs[0] = inputs[0].asType('float32');
>         inputs[1] = inputs[1].asType('float32');
>       }

I can imagine this class getting pretty long... I see this pattern repeated a few times. Maybe this is a candidate for a utility method?

---

*[models/tensorflow/src/node.ts, line 153 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/536#-L33bLwbmQPAbFjuRWkc:-L33bLwbmQPAbFjuRWkd:b-jhel8p) ([raw file](https://github.com/pair-code/deeplearnjs/blob/2df0d6653967f2ff5dcac5a6753a252f5fc9acd1/models/tensorflow/src/node.ts#L153)):*
>  case 'MaxPool': {
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>
>       const poolingParam = node.attr;
>       const stride =
>           getNumericArrayParam(poolingParam, 'strides', [1, 1], 1, 3);
>       const pad =
>           getStringParam(poolingParam, 'padding', 'valid').toLowerCase();
>       const kernelSize =
>           getNumericArrayParam(poolingParam, 'ksize', [1, 1], 1, 3);
</details>
>  
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>
>       return math.maxPool(
>           input as Array3D, kernelSize as [number, number],
>           stride as [number, number], pad as 'valid' | 'same');
>     }
</details>
I'm actually wondering if it makes sense to provide an interface that automatically provides these node constants in some structure. This would be implemented in DLJS and we would subclass in the TF repo. This would allow others to provide model loading and inference through the API.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/536#-:-L33bh_HusciEajCDpFF:b5n4zvl)*

nsthorat:
Hey Ping,

We should probably wait for the design doc to be reviewed before going forward with this, is that right?

Issue #569, opened 2018-01-20
invalid link: /deeplearnjs/blob/master/yolo_mobilenet/demo.html 


jaassoon:
in deeplearnjs/master/models/yolo_mobilenet/README.md
`Check out [demo.html](https://github.com/PAIR-code/deeplearnjs/blob/master/yolo_mobilenet/demo.html)`

PR #573, opened 2018-01-21
invalid link: /deeplearnjs/blob/master/yolo_mobilenet/demo.html (#569)

jaassoon:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/573)


nsthorat:
Thanks for the contribution!

PR #577, opened 2018-01-22
Add @operation decorator and migrate more math ops

dsmilkov:
Add `@operation` decorator that wraps the method in a math.scope(methodName, () => {}). This helps annotate all ops, track them (in debug mode) and significantly reduce chances of memory leaks.

Also migrate more math ops to global namespace:
 - compare (less, greater, equal etc)
 - min/max/minimum/maximum/argMin/argMax/argMaxEquals
 - transpose

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/577)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -364,18 +364,4 @@ function getOrMakeEnvironment(): Environment {
   return ns.ENV;
 }
 
-/**
- * Decorator for wrapping functions that perform math operations on
- * NDArrays. The function will be wrapped in a named scope that cleans all
- * memory usage after the function is done.
- */
-export function operation(
-    target: {}, name: string, descriptor: PropertyDescriptor) {
-  const fn = descriptor.value;
-  // tslint:disable-next-line:no-any
-  descriptor.value = (...args: any[]) =>
-      ENV.math.scope(name, () => fn(...args));
-  return descriptor;
-}
-
 export let ENV = getOrMakeEnvironment();
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -364,4 +364,18 @@ function getOrMakeEnvironment(): Environment {
   return ns.ENV;
 }
 
+/**
+ * Decorator for wrapping functions that perform math operations on
+ * NDArrays. The function will be wrapped in a named scope that cleans all
+ * memory usage after the function is done.
+ */
+export function operation(
+    target: {}, name: string, descriptor: PropertyDescriptor) {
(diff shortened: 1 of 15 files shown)


nsthorat:
Awesome change, lgtm. Spoke offline, but for posterity:
1) Can you check what the generated code looks like for es5? Just curious if it does anything funky to the stack trace / if there's anything unexpected. Obviously not blocking, but would be good to understand.
2) We should convert unit tests to call top level `dl.xyz()`ops since that's the API we're going to be supporting. Maybe that can be done in a follow up, but let's keep it on the radar.

:lgtm_strong:

---

Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/environment.ts, line 372 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/577#-L3QsKqGf6Thucs4Kn33:-L3QsKqGf6Thucs4Kn34:b-pu77pa) ([raw file](https://github.com/pair-code/deeplearnjs/blob/40f686cc3e9246e3c039feb89b012bf197032bea/src/environment.ts#L372)):*
> ```XML
>  * memory usage after the function is done.
>  */
> export function operation(
> ```

this feels a little weird in this file, maybe this could live on its own?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/577#-:-L3QuQKCYF0SlspsMzGp:bxp5waa)*

dsmilkov:
Review status: 0 of 14 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/environment.ts, line 372 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/577#-L3QsKqGf6Thucs4Kn33:-L3R89MKgOoyWn7c5ZyO:b-ev5ken) ([raw file](https://github.com/pair-code/deeplearnjs/blob/40f686cc3e9246e3c039feb89b012bf197032bea/src/environment.ts#L372)):*

this feels a little weird in this file, maybe this could live on its own?
</blockquote></details>

Done. Moved to `math/decorators.ts`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/577)*

dsmilkov:
Stack trace looks good. Only 1 more call got added to the stack (`description.value`). Attaching the trace of a `math.matMul()` call from a standalone `<script>` tag:

stack-trace

PR #579, opened 2018-01-22
Replace undefined Error object

skwbc:
In `scripts/dump_checkpoints/dump_checkpoint_vars.py` `Error` object is raised, but it doesn't exist in Python. So, I replaced `Error` with [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError). 
I also found invalid use of `format` method. As you know, `%s` is not treated by `format` method. So, I replaced `%s` with `{}`.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/579)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/579#-:-L3iRvwMsrBUa012CMO9:b-jbsf8z)*

PR #591, opened 2018-01-25
Add cifar 10 demo

tafsiri:
Here is a branch with a CIFAR10 inference demo. There is no UI yet just output in the console.

Not really looking to merge at this point but for some initial review of the code. You should be able to pull down this branch/repo and take a look. This port is primarily based on https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10.py

My next steps are to look into how training works and then think through how I can better compare this model with the python one it was ported form.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/591)


nsthorat:
Nice work! Sent some thoughts inline...

---

Review status: 0 of 36 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[demos/cifar10/cifar10.ts, line 2 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMEDZYBCBD1pMfata:-L3iMED_i_ZFD5CzBYgb:b-wk9r7h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L2)):*
> ```XML
> // tslint:disable-next-line:max-line-length
> import {Array3D} from 'deeplearn/dist/math/ndarray';
> ```

dont import like this, import from 'deeplearn' only.

the way we're going to do demos from here on is "import * as dl from 'deeplearn'" and use dl prefixes

---

*[demos/cifar10/cifar10.ts, line 10 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iIPmjG43CCW26avDs:-L3iIPmjG43CCW26avDt:bs0fe50) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L10)):*
> ```XML
>   console.log('Start');
> 
>   Promise.all([loadModelVariables('weights/'), loadCifarData()])
> ```

one thing you could do here is make start async and:

`await Promise.all([loadModelVariables(), loadCifarData()])`

instead of using .then (we started removing then where we can)

---

*[demos/cifar10/cifar10.ts, line 21 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMLymOID99CvX_bgk:-L3iMLymOID99CvX_bgl:bmcs89g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L21)):*
> ```XML
>         for (let start = 0; start < testData.images.length;
>              start += batchSize) {
>           const batch =
> ```

maybe this compiling of the batch should happen in data?

---

*[demos/cifar10/cifar10.ts, line 23 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMWqixwqCl1kr4Zes:-L3iMWqixwqCl1kr4Zet:b-eiy2z7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L23)):*
> ```XML
>           const batch =
>               testData.images.slice(start, start + batchSize) as Array3D[];
>           const batched = toArray4d(batch);
> ```

maybe call this "inputBatch"

---

*[demos/cifar10/cifar10.ts, line 28 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMcn89dedxb48NDEH:-L3iMcn9A9ocTjJHT3Ne:bg1b81d) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L28)):*
> ```XML
>           const res = inference(modelVars, batched);
> 
>           for (let i = 0; i < res.predictionLabel.length; i++) {
> ```

Instead of doing it this way, you should vectorize it by using "argmax", "equals", and "sum"

---

*[demos/cifar10/data.ts, line 39 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLjCbZDkIAbG29kxv:-L3iLjCbZDkIAbG29kxw:bb1nmc0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L39)):*
> ```XML
>       'ship', 'truck'
>     ],
>     modelConfigs: undefined,
> ```

use null over undefined (we don't assign undefined as part of standard practices)

---

*[demos/cifar10/data.ts, line 44 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLc4owqeEHaL5CA2g:-L3iLc4owqeEHaL5CA2h:boask35) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L44)):*
> ```XML
>   const dataset = new XhrDataset(datasetConfig);
> 
>   return dataset.fetchData().then(() => {
> ```

use "await" everywhere you use .then, we've been converting everything to async/await (when we compile ts has a es5 polyfill)

---

*[demos/cifar10/data.ts, line 52 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLIfzZqPy6Lj5l0ow:-L3iLIfzZqPy6Lj5l0ox:b-89nrvl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L52)):*
> ```XML
>     // Resize the images
>     images = images.map((i) => {
>       return ENV.math.resizeBilinear3D(i as Array3D, [24, 24]);
> ```

images.map((i) => ENV.math.resizeBilinear3D(i as Array3D, [24, 24]);

---

*[demos/cifar10/data.ts, line 60 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLDXVSPwuAgkNetdi:-L3iLDXVSPwuAgkNetdj:bq92rk8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L60)):*
> ```XML
>       return {
>         label,
>         labelString,
> ```

remove trailing comma

---

*[demos/cifar10/data.ts, line 66 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLCadrOq-k8YmmShO:-L3iLCadrOq-k8YmmShP:bq92rk8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L66)):*
> ```XML
>     return {
>       images,
>       labelled,
> ```

remove trailing comma

---

*[demos/cifar10/data.ts, line 82 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLT_V6WpEnhBqm5GQ:-L3iLT_V6WpEnhBqm5GR:b-n4ngm4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L82)):*
> ```XML
>   });
> 
>   const all = asArray4d.reduce((memo, curr) => {
> ```

asArray4d.reduce((memo, curr) => ENV.math.concat4D(memo, curr, 0));

---

*[demos/cifar10/model.ts, line 65 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iKProQo19iA42fOZS:-L3iKProQo19iA42fOZT:b-2cngfd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L65)):*
> ```XML
>           'same'  // Could be nice as a constant.
>       );
>       const relu = math.relu(conv);
> ```

for simplicity you could do `return math.relu(...)` here and everywhere else

---

*[demos/cifar10/model.ts, line 81 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iKjrABODDZhV3TCJE:-L3iKjrBYzh72AE7rmMM:bisuczr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L81)):*
> ```XML
>           conv2Biases,
>           [1, 1],
>           'same',
> ```

kill the trailing comma and auto format will be kind

---

*[demos/cifar10/model.ts, line 108 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iJobkigFVE4Ax4pXp:-L3iJobkigFVE4Ax4pXq:bmfnh8l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L108)):*
> ```XML
>     // Local4
>     const local4Res = math.scope((keep) => {
>       const mul = math.matMul(local3Res, local4Weights);
> ```

Soon youll be able to do:

local3Res.matMul(softmaxWeights).add(softmaxBiases).relu()!

---

*[demos/cifar10/model.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iJMzbaXd5Yz-anFTS:-L3iJMzbaXd5Yz-anFTT:b-of5alu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L132)):*
> ```XML
>   const prediction = math.argMax(finalRes, 1) as Array1D<'int32'>;
>   const predictionLabel =
>       Array.from(prediction.getValues()).map(i => labelStrings[i]);
> ```

getValues is deprecated, use dataSync

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-:-L3iNAxv8FgEf2hoAuE5:b9jw2zl)*

tafsiri:
Thanks for the review. Left a few comments and will be working on changes.

---

Review status: 0 of 36 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[demos/cifar10/cifar10.ts, line 2 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMEDZYBCBD1pMfata:-L3ioCsrdWTp_MNviG0D:bilg754) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L2)):*

dont import like this, import from 'deeplearn' only.

the way we're going to do demos from here on is "import * as dl from 'deeplearn'" and use dl prefixes
</blockquote></details>

Yep! I think this was an auto-import that i didn't notice.

---

*[demos/cifar10/cifar10.ts, line 21 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMLymOID99CvX_bgk:-L3iosryD_D9mrXVuTR2:b-jdxka9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L21)):*
> Yep! I think this was an auto-import that i didn't notice.
I mostly think of data as data loading. Batch size and other things being determined by the consumer of the data. 

Separately, one possible thing we could do in future if we have a data loading api/utility for things would be to support iterators that produce batches. 

---

*[demos/cifar10/cifar10.ts, line 28 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMcn89dedxb48NDEH:-L3ipRT48QXaLGqvQSuU:b-o7ha5f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L28)):*

Instead of doing it this way, you should vectorize it by using "argmax", "equals", and "sum"
</blockquote></details>

will look into this. I think its this way because i converted to strings too early.

---

*[demos/cifar10/data.ts, line 44 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLc4owqeEHaL5CA2g:-L3inCjYAU3NnJQ8p9Ee:buf7l7u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L44)):*

use "await" everywhere you use .then, we've been converting everything to async/await (when we compile ts has a es5 polyfill)
</blockquote></details>

Cool. so this particular one is one i had feedback on from an api point of view. I found it strange that the promise didn't seem to resolve to any value (no params passed to the 'then' call back). Am I missing something. I basically have to keep a reference to dataset and thus couldn't just return the promise. 

---

*[demos/cifar10/data.ts, line 60 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iLDXVSPwuAgkNetdi:-L3ipYJyD4t-0zmr6-5g:b-47ezbz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/data.ts#L60)):*

remove trailing comma
</blockquote></details>

:( okay... (jk - i too used to be against trailing commas)

---

*[demos/cifar10/model.ts, line 81 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iKjrABODDZhV3TCJE:-L3ipvvam6e0KbNcuWST:b-w1m49w) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L81)):*

kill the trailing comma and auto format will be kind
</blockquote></details>

I wonder if clang-format can do that automatically

---

*[demos/cifar10/model.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iJMzbaXd5Yz-anFTS:-L3iqTNOCuWqL9yyRgwT:b-gt37z5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L132)):*

getValues is deprecated, use dataSync
</blockquote></details>

huh interesting? are they just aliases or do they mean different things?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-:-L3iqcKgmIe7T3C-wP3p:b-wdxitl)*

tafsiri:
*[demos/cifar10/cifar10.ts, line 2 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMEDZYBCBD1pMfata:-L3irO2GkTOaEhzWkYLf:bvuihw1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L2)):*

Yep! I think this was an auto-import that i didn't notice.
</blockquote></details>

Is the preference to use the * import style even when just using one or two exported identifiers? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591)*

tafsiri:
I think i've addressed all the comments and left an additional comment.

---

Review status: 0 of 34 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[demos/cifar10/cifar10.ts, line 21 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iMLymOID99CvX_bgk:-L3jYfonoLQMQtrZ27ut:blv84eb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/cifar10.ts#L21)):*

> Yep! I think this was an auto-import that i didn't notice.
I mostly think of data as data loading. Batch size and other things being determined by the consumer of the data. 

Separately, one possible thing we could do in future if we have a data loading api/utility for things would be to support iterators that produce batches. 
</blockquote></details>

To add an additional comment i think its nice to show batching as an explicit author action as opposed to something 'inherent' in how data comes in.

---

*[demos/cifar10/model.ts, line 91 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3jYzIEkEPb-LTiQhaX:-L3jYzIEkEPb-LTiQhaY:brnrwod) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b63e95db28c1d7e0affeb5dfa03107775aa4e70c/demos/cifar10/model.ts#L91)):*
> ```XML
>     // implementation has a comment to that effect even though it is called
>     // softmax_linear.
>     const softmaxRes = math.scope((keep) => {
> ```

Do you think we should call this softMax? I did that because its that way in tensorflow, but the comment in the tensorflow code is that this does not perform a softMax. This is the line/comment in question https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10.py#L260 which i didn't fully understand.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-:-L3jYpeljhyZzupCGLeT:batzjct)*

nsthorat:
Review status: 0 of 34 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[demos/cifar10/model.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3iJMzbaXd5Yz-anFTS:-L3jyCuhhuVwaLkFhlkL:b-fjokmm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ee69af74facc8a16b394b8379c1ef771afe095f2/demos/cifar10/model.ts#L132)):*

huh interesting? are they just aliases or do they mean different things?
</blockquote></details>

Spoke offline: this is an alias, we wanted users to use the short .data() in general, which is a promise, and dataSync is the synchronous download.

---

*[demos/cifar10/model.ts, line 91 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591#-L3jYzIEkEPb-LTiQhaX:-L3jxvrYmNTZMfz_OeKu:b-d3z78l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b63e95db28c1d7e0affeb5dfa03107775aa4e70c/demos/cifar10/model.ts#L91)):*

Do you think we should call this softMax? I did that because its that way in tensorflow, but the comment in the tensorflow code is that this does not perform a softMax. This is the line/comment in question https://github.com/tensorflow/models/blob/master/tutorials/image/cifar10/cifar10.py#L260 which i didn't fully understand.
</blockquote></details>

Yeah, you shouldn't call it softmax here, the TF model is wrong.

The softmax of the logits happen in the loss function defined in their model. If you were interested in printing out class probabilities instead of just the class index of the best prediction you could call math.softmax on the result, but yes, you shouldn't call it softmax here for clarity :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/591)*

nsthorat:
Can we close this PR?

tafsiri:
This is working but we are not merging for now. Closing the PR, though i will leave the branch up .

PR #606, opened 2018-01-28
Unary Ops Grad: Adds gradients for math.ceil and math.floor

ManrajGrover:
This PR adds gradients for `math.ceil` and `math.floor`

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/606)


nsthorat:
Hm.. I think the gradients for ceil and floor should both be 0. Think about what the graph of ceil(x) and floor(x) are.

ManrajGrover:
@nsthorat I feel that since the graph has discontinuity when x is an integer and constant when x < ceil(x), the derivative of ceil(x) is 0 when x < ceil(x) else undefined. Please correct me if I'm wrong.

dsmilkov:
The gradient of tf.ceil and tf.floor is zero (other than exactly at the integer spot, but that makes things complicated). TF chose 0: https://github.com/tensorflow/tensorflow/issues/897. But TF also does something special where they return `None` instead of bunch of zeros, so they don't waste computation and memory. Let's revert this PR. I can then go add the infrastructure code necessarily to deal with null as gradients of ops (make sure the backprop works will null). Then,  we can come back to ceil and floor and return null as gradients. Does that sound good?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606#-:-L4WnucSQgaV1MktdOR-:b-8iox1i)*

ManrajGrover:
@dsmilkov Yeps. I'll make it return a `NDArray.zero` for now. Or, we can keep this PR open till we can make backprop work with nulls.

I feel we will face similar issues in `clip`, `step` and other similar functions. I'm assuming we'll return zeros for them as well.

dsmilkov:
Talked to Nikhil. Since optimization is non-trivial, and shouldn't block this PR, let's merge this in. e can optimize later. Don't revert please! :)

ManrajGrover:
@dsmilkov Should I make it return `NDArray.zeros` or let it remain the same?

dsmilkov:
Return NDArray.zeros() always. Don't detect integers.

ManrajGrover:
@dsmilkov Made the required change

nsthorat:
Awesome! Will merge after tiny comment.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[src/math/unary_ops.ts, line 45 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606#-L4XeBGEF52z0iePzM7Z:-L4XeBGEF52z0iePzM7_:bxbrsgo) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8cee3c69eaeb73d5c945e2174ebe5175b2ea90fc/src/math/unary_ops.ts#L45)):*
> ```XML
>   static ceil<T extends NDArray>(x: T): T {
>     const gradient = (dy: T, y: T) => {
>       return {x: () => NDArray.zeros(y.shape)};
> ```

we're going to deprecate the NDArray methods soon, can you call "Ops.zeros" after importing array_ops?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606#-:-L4XeRA6Gk7MHOwqzW6v:b-52xe32)*

nsthorat:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[src/math/unary_ops.ts, line 45 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606#-L4XeBGEF52z0iePzM7Z:-L4Xec3j655K-e4nJFlF:b2r7liq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8cee3c69eaeb73d5c945e2174ebe5175b2ea90fc/src/math/unary_ops.ts#L45)):*

we're going to deprecate the NDArray methods soon, can you call "Ops.zeros" after importing array_ops?
</blockquote></details>

It looks like "ops" is already imported, I think you can call "ops.zeros" as is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606)*

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/606#-:-L4aC_XZuFdk_cPhlbIX:b-jbsf8z)*

Issue #613, opened 2018-01-29
Add dl.linspace and dl.arange.


nsthorat:
We should add helper linspace and arange functions to generate numbers within a range.

tafsiri:
@nsthorat Do you think there is any value in naming `arange` as `range` instead? I imagine the arange name in python is to make it distinct from the built in range function. My guess being arange -> 'array range', which doesn't have the same meaning/connotation in js (where arrays are the standard datastructure as opposed to lists).

dsmilkov:
+1 to calling it just `range`. As you said, `arange` is a side-effect of `range` being already built-in in python :)

PR #624, opened 2018-01-31
613 add dl range

tafsiri:
Partially addresses #613

Adds an NDArray.range function to generate new 1d arrays with spanning a given range of integers.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/624)


tafsiri:
Note that unlike the python version `start` is not optional. Should it be?

src/math/ndarray.ts
```diff
@@ -161,6 +161,12 @@ export class NDArray<R extends Rank = Rank> {
     return ops.randUniform(shape, a, b, dtype);
   }
 
+  static range(
```
tafsiri: It might be that this isn't in the right place as a number of other array creation methods right above this are indicated as deprecated. Lets chat about where this should be. I guess this would also affect how the tests are called.

nsthorat:
Nice work!!

---

Review status: 0 of 4 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/math/array_ops.ts, line 304 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CvTbVHfZkySonm7SM:-L4CvTbVHfZkySonm7SN:bp8smpo) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L304)):*
> ```XML
> 
>   /**
>    * Creates a new Array1D filled with the numbers in the range
> ```

Can you use terminology similar to numpy here, specifically using "half-open interval [start, stop)"

---

*[src/math/array_ops.ts, line 310 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4Cv3bYhhO6v03lUWmW:-L4Cv3bYhhO6v03lUWmX:b3ippul) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L310)):*
> ```XML
>    * negative step values are supported.
>    *
>    * @param start an integer start value
> ```

Capitalize "An" and period at the end (we should be doing this throughout, where we don't it's a mistake)

---

*[src/math/array_ops.ts, line 315 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CtW6Ft-RO9QQvYXgQ:-L4CtW6GLopPC5LD4LxH:b5doylq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L315)):*
> ```XML
>    * @param dtype an optional dtype
>    */
>   static range(
> ```

tag this with @operation. The "scope" won't do anything here, but we're going to use @operation decorators to help with logging

---

*[src/math/array_ops.ts, line 317 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4Ct3DrtQbceKkEdL03:-L4Ct3DrtQbceKkEdL04:bhpeeav) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L317)):*
> ```XML
>   static range(
>       start: number, stop: number, step = 1,
>       dtype: DType.float32|DType.int32 = DType.float32): Array1D {
> ```

use the string literals for dtype here for consistency

dtype: 'float32'|'int32' = 'float32'

---

*[src/math/array_ops.ts, line 328 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CtEP4DGX06XGbnSq0:-L4CtEP4DGX06XGbnSq1:b-tzczw0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L328)):*
> ```XML
>     if (sameStartStop || increasingRangeNegativeStep ||
>         decreasingRangePositiveStep) {
>       return Array1D.new(makeZerosTypedArray(0, dtype), dtype);
> ```

Can you just call our "zeros" method here (it lives in this file)

---

*[src/math/array_ops_test.ts, line 1191 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4Cu7_Ejj5wDYqCjjVS:-L4Cu7_Ejj5wDYqCjjVT:b-7na0vs) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops_test.ts#L1191)):*
> ```XML
> };
> 
> const testsRange: MathTests = it => {
> ```

can you also test the shape / rank in all of these method?

---

*[src/math/array_ops_test.ts, line 1193 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CtjrmBZ-LcsVMmXrW:-L4CtjrmBZ-LcsVMmXrX:b72bvl5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops_test.ts#L1193)):*
> ```XML
> const testsRange: MathTests = it => {
>   it('range start stop', () => {
>     const a = NDArray.range(0, 3);
> ```

You should `import * as dl from 'deeplearn'` here and use dl.range. We're going to change the other tests to do that in a follow up PR, but in the meantime let's do it here.

---

*[src/math/array_ops_test.ts, line 1267 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CuHhPfeUY5oB380jU:-L4CuHhPfeUY5oB380jV:b-vf5z7v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops_test.ts#L1267)):*
> ```XML
>   it('range start stop incompatible step', () => {
>     const a = NDArray.range(3, 10, -2);
>     test_util.expectArraysEqual(a, new Float32Array(0));
> ```

Why can't you just use "[]"?

---

*[src/math/array_ops_test.ts, line 1285 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4Ctz5GAkDTOV1xF-OD:-L4Ctz5GAkDTOV1xF-OE:b81bgly) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops_test.ts#L1285)):*
> ```XML
> 
>   it('range float32 dtype', () => {
>     const a = NDArray.range(1, 4, undefined, DType.float32);
> ```

use 'float32' for consistency

---

*[src/math/ndarray.ts, line 164 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4ChdoZbhEI29QNlwYD-r2-164:-L4Cug7bn7PEaJ3rixtM:bkzeqyy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/ndarray.ts#L164)):*

It might be that this isn't in the right place as a number of other array creation methods right above this are indicated as deprecated. Lets chat about where this should be. I guess this would also affect how the tests are called.
</blockquote></details>

Yeah, your intuition is right. This should only live on top level "dl" by exporting it in ops.ts like you do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-:-L4CthIdMFzgtC_aj--K:b2ky7kf)*

tafsiri:
Review status: 0 of 4 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/math/array_ops.ts, line 304 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CvTbVHfZkySonm7SM:-L4D3yCdMrbBb6NmyLP0:b-jt9h44) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L304)):*

Can you use terminology similar to numpy here, specifically using "half-open interval [start, stop)"
</blockquote></details>

What do you think of using half open terminology in the second sentence but not the first? Its a bit of a red flag to me that the numpy docs have to explain the term right after they use it.  So i was trying to see if we could present something clearer and friendlier at first.

---

*[src/math/array_ops_test.ts, line 1267 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CuHhPfeUY5oB380jU:-L4D6LTw22xo4LEqsOMc:b5ri30f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops_test.ts#L1267)):*

Why can't you just use "[]"?
</blockquote></details>

As discussed I suspect this is a bug in the test_util. I think i'd prefer to try fix it in a separate PR just to not risk accidentally making some other things pass that shouldn't. Thoughts?

I could also make a PR into this branch to work on the test util.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624)*

tafsiri:
Updated the code to address most of the review comments.

src/math/array_ops_test.ts
```diff
@@ -1188,17 +1188,142 @@ const testsClone: MathTests = it => {
...
+    expect(a.shape).toEqual([0]);
+
+    const b = dl.range(40, 3, 2);
+    test_util.expectArraysEqual(b, new Float32Array(0));
+    expect(b.shape).toEqual([0]);
+  });
+
+  it('range zero step', () => {
+    expect(() => dl.range(2, 10, 0))
+        .toThrow(new Error('Cannot have a step of zero'));
```
nsthorat: Here let's not assert the exact text, let's just say assertThrows

nsthorat:
Reviewable is down right now, but both of the last comments you sent are fine with me (follow up PR & your terminology for the jsdoc). Feel free to squash and merge once you update the branch.

nsthorat:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/math/array_ops.ts, line 304 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-L4CvTbVHfZkySonm7SM:-L4DBLn41OrF40qo8j2T:bff5zk8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b665b9b3a991e6c43f6c3be950c9597d5a9f2379/src/math/array_ops.ts#L304)):*

What do you think of using half open terminology in the second sentence but not the first? Its a bit of a red flag to me that the numpy docs have to explain the term right after they use it.  So i was trying to see if we could present something clearer and friendlier at first.
</blockquote></details>

This looks good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624)*

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/624#-:-L4NRmsjBRIMK-j2YvZC:b-jbsf8z)*

tafsiri:
Note that this was merged even though travis is failing as these tests do not overlap with those ones.

PR #629, opened 2018-02-01
refact_kernel_registry

jaassoon:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/629)


nsthorat:
Hi @jaassoon, I'd like to keep this as is for now. Thanks for the change though :)

PR #631, opened 2018-02-01
Add dl.linspace

### LINKED ISSUE #613: Add dl.linspace and dl.arange.

tafsiri:
Closes #613

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/631)


tafsiri:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/math/array_ops.ts, line 325 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631#-L4aMvWS2SFq6QjOjDEQ:-L4bisk_44qjyOLe8W5P:b-o82n9d) ([raw file](https://github.com/pair-code/deeplearnjs/blob/17e38571c03c84be3e0f0facd8c7e470b7b3dd13/src/math/array_ops.ts#L325)):*

looks like we duplicate with range, can we consolidate and use the same logic (one calls the other)?
</blockquote></details>

I don't think we can. Each one calculates something in a different way (either step, or the number of elements).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/math/array_ops.ts, line 444 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/631#-L4cA_6P4WMcErkqOp_e:-L4cA_6QTMaCYfblraCd:b-i1143p) ([raw file](https://github.com/pair-code/deeplearnjs/blob/852c9cf5b5af84ccf36fddd0ce326a92475f22ab/src/math/array_ops.ts#L444)):*
> ```XML
>   @operation
>   @doc({heading: 'Tensors', subheading: 'Creation'})
>   static linspace(start: number, stop: number, num: number, endpoint = true) {
> ```

let's get rid of endpoint to be exactly TF

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/631#-:-L4cAf84TNz_p54ULaZ7:b-jbsf8z)*

tafsiri:
*[src/math/array_ops.ts, line 444 at r2](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631#-L4cA_6P4WMcErkqOp_e:-L4cFz253fCSUF6C8tsi:b6icn65) ([raw file](https://github.com/pair-code/deeplearnjs/blob/852c9cf5b5af84ccf36fddd0ce326a92475f22ab/src/math/array_ops.ts#L444)):*

let's get rid of endpoint to be exactly TF
</blockquote></details>

Okay. I'll keep in mind for future reference that we are targeting TF compat and not numpy.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631)*

tafsiri:
*[src/math/array_ops.ts, line 444 at r2](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631#-L4cA_6P4WMcErkqOp_e:-L4cGlc58MnDklzWv5rn:bfdzr2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/852c9cf5b5af84ccf36fddd0ce326a92475f22ab/src/math/array_ops.ts#L444)):*

Okay. I'll keep in mind for future reference that we are targeting TF compat and not numpy.
</blockquote></details>

Though i do use numpy to generate test cases... It should be exactly the same this case though.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/631)*

Issue #630, opened 2018-02-01
Tests failing on travis


tafsiri:
We seem to have some tests failing on travis. https://travis-ci.org/PAIR-code/deeplearnjs/builds/335849897 I think they were introduced in #627. @dsmilkov would you be able to take a look?

nsthorat:
https://github.com/PAIR-code/deeplearnjs/pull/633

nsthorat:
Ah, apologies. You are right, they were introduced here:

https://github.com/PAIR-code/deeplearnjs/pull/627

Specifically "clone" interacting with "set" now breaks:
https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/ndarray_test.ts#L74-L79

@dsmilkov can you send a fix?

dsmilkov:
Yup, will remove `set` and `add` in NDArray. There is no point in keeping backwards compatibility with these 2 methods when we are making `NDArray`s immutable.

dsmilkov:
Fixed in #625 (I did it in that PR since the fix was tiny, but I wanted to make sure Travis passes)

PR #633, opened 2018-02-02
Turn off localResponseNormalization#D test for integer-packed textures.

nsthorat:
I disabled the wrong test in the previous PR. This will remove flake, but we should investigate why this is flaky.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/633)


nsthorat:
@chaosmail any idea why this test would be flaky?

chaosmail:
hmm very strange, all tests passed on travis for LRN right? Could you give me the exact test settings to reproduce the bug so I could take a look?

nsthorat:
It seems to be flaky, so it's unclear. Here is an example: https://api.travis-ci.org/v3/job/336158839/log.txt

dsmilkov:
:lgtm_strong: Add a todo in the test

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/633#-:-L4N5TirNYKuUe5k8WpX:b8s1hkj)*

tafsiri:
@nsthorat Just FYI doubt its related to the change you made but i'm seeing other failures in the log. e.g.  dl.reverse as well. We can continue to track flakiness in #630

PR #644, opened 2018-02-05
Add gradient for clip

caisq:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/644)


src/math/unary_ops.ts
```diff
@@ -121,8 +122,14 @@ export class Ops {
         (min <= max),
         `Error in clip: min (${min}) must be` +
             `less than or equal to max (${max}).`);
-    return ENV.engine.executeKernel('Clip', {inputs: {x}, args: {min, max}}) as
-        T;
+    return ENV.engine.executeKernel(
+        'Clip', {inputs: {x}, args: {min, max}}, (dy: T, y: T) => {
+      return {
+          x: () => dy.where(
```
ManrajGrover: Gradient for `clip` is not defined for x = min and x = max. We should add a TODO here to fix this once it is defined. Reference: #606

src/math/unary_ops.ts
```diff
@@ -121,8 +122,14 @@ export class Ops {
         (min <= max),
         `Error in clip: min (${min}) must be` +
             `less than or equal to max (${max}).`);
-    return ENV.engine.executeKernel('Clip', {inputs: {x}, args: {min, max}}) as
-        T;
+    return ENV.engine.executeKernel(
```
ManrajGrover: `dl.where` currently does not support `Scalar` values. We should throw an error for this.

src/math/unary_ops_test.ts
```diff
@@ -1309,6 +1309,18 @@ import {Array1D, Array2D, Scalar} from './ndarray';
       };
       expect(f).toThrowError();
     });
+
```
ManrajGrover: Let's add tests for `Scalars` to check considering it is not supported yet.

caisq:
oops, I need to pull and rebase.

nsthorat:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/math/unary_ops.ts, line 125 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/644#-L4asXvnfc1509SSs16T-r1-125:-L4foGJpzeu09uWqRFB0:bxfz60) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ede5139e436885fcb61a77c0af2aeb1edc71acf2/src/math/unary_ops.ts#L125)):*
<details><summary><i>Previously, ManrajGrover (Manraj Singh) wrote…</i></summary><blockquote>

`dl.where` currently does not support `Scalar` values. We should throw an error for this.
</blockquote></details>

Any idea why this is? It should be easy to fix right?

https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/logical_ops.ts#L100

---

*[src/math/unary_ops.ts, line 129 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/644#-L4fnvQruF8tL5IyefIO:-L4fnvQruF8tL5IyefIP:b-xcz9c8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ede5139e436885fcb61a77c0af2aeb1edc71acf2/src/math/unary_ops.ts#L129)):*
> ```XML
>       return {
>           x: () => dy.where(
>               x.greater(Scalar.new(min)).logicalAnd(x.less(Scalar.new(max))),
> ```

import "scalar" from "ops" and call that function here instead of "Scalar.new", "Scalar.new" is deprecated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/644)*

ManrajGrover:
@nsthorat 

> Any idea why this is? It should be easy to fix right?
> 
> https://github.com/PAIR-code/deeplearnjs/blob/master/src/math/logical_ops.ts#L100

This happens because `dl.where` internally calls `getCoordsDataType()` which does not support rank 0. We should make it use `int` for both rank 0 and rank 1.

ManrajGrover:
Have opened an issue for this #665 and fix is in #666

src/math/unary_ops.ts
caisq: Done.

src/math/unary_ops.ts
caisq: Done.

src/math/unary_ops_test.ts
caisq: Since you are implementing it soon, I decided to not add an explicit unit test for that for now.

nsthorat:
@caisq can you merge with master? @ManrajGrover fixed the scalar issue.

caisq:
@nsthorat Done.

caisq:
PTAL @nsthorat @ManrajGrover ?

dsmilkov:
:lgtm_strong:

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/644#-:-L4krtFcoXSRBtyLL0h0:b-jbsf8z)*

dsmilkov:
Thanks @caisq . All comments were addressed, so I merged!

PR #646, opened 2018-02-05
Introduce dl.tensor/scalar/tensor1d/.... and migrate old API

dsmilkov:
Also
- rename `concat1D` --> `concat1d`. Similarly for other ops that have suffix `[0-9]D`
- Make SymbolicTensor (in Graph) assignable to Tensor, so we don't break Graph users.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/646)


nsthorat:
:lgtm_strong:

High level question, since we're calling it "tensor1d", lower case "d", what do you think of making this consistent for some of the rank-specific methods, aka "concat4D" => "concat4d"

---

Reviewed 51 of 52 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/util_test.ts, line 15 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-L4bM4eJqVwpCQJSC53l:-L4bM4eJqVwpCQJSC53m:b-83tlrg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df47edc3b379ad81d7f51ee91b0a8d0e9e75b53b/src/util_test.ts#L15)):*
> ```XML
>  * See the License for the specific language governing permissions and
>  * limitations under the License.
> ```

unit tests for TypedArray passed

---

*[src/util_test.ts, line 343 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-L4bM8mjuRKO3PSpZqM0:-L4bM8mjuRKO3PSpZqM1:b-ywzuid) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df47edc3b379ad81d7f51ee91b0a8d0e9e75b53b/src/util_test.ts#L343)):*
> ```XML
>       const a = scalar(1);
>       const b = scalar(3);
>       const c = dl.tensor1d([1, 2, 3]);
> ```

let's be consistent here? dl.scalar(1) everywhere instead of scalar(1)

---

*[src/math/array_ops.ts, line 31 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-L4bL6qtPuhScvBI7EsQ:-L4bL6qtPuhScvBI7EsR:b9pvfk7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df47edc3b379ad81d7f51ee91b0a8d0e9e75b53b/src/math/array_ops.ts#L31)):*
> ```XML
> 
> export class Ops {
>   /** Creates a tensor with the provided values, shape and dtype. */
> ```

Flesh this out, specifically @params, here and below (see how it's done on other array_ops)

---

*[src/math/array_ops.ts, line 34 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-L4bLhgtqqyWtEBO3Zh7:-L4bLhgtqqyWtEBO3Zh8:b1jy7aw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df47edc3b379ad81d7f51ee91b0a8d0e9e75b53b/src/math/array_ops.ts#L34)):*
> ```XML
>   @doc({heading: 'Tensors', subheading: 'Creation'})
>   static tensor<R extends Rank>(
>       values: AllowedValues, shape?: ShapeMap[R], dtype: DataType = 'float32'):
> ```

maybe just inline the AllowedValues type since it's used once (this also makes it easy to see the type in the docs if we want to show the type).

---

*[src/math/binary_ops.ts, line 152 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-L4bMnTHKJHAcFhPhFrP:-L4bMnTHKJHAcFhPhFrQ:b-bx6ebu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/df47edc3b379ad81d7f51ee91b0a8d0e9e75b53b/src/math/binary_ops.ts#L152)):*
> ```XML
>       const derBase = () => {
>         const dx = exp.toFloat().mul(
>             base.pow(exp.sub(ops.scalar(1, 'int32'))).toFloat());
> ```

In library code, I think it makes sense to import {scalar} and use it directly

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/646#-:-L4bNn9l4uKHviwUrQnv:b-ybfkjx)*

dsmilkov:
This is part of #656

PR #670, opened 2018-02-07
Add dl.memory() and use WeakMap to avoid mem leaks when data is stored on cpu

### LINKED ISSUE #662: Add dl.memory()
Expose `dl.memory()` which will return a rich object:
```ts
{
  numDataBuffers: number;
  numBytes: number;
}
```
`dataBuffers <= tensors`, since `a.reshape(newShape)` makes a new Tensor that shares the same data buffer with `a`.

dsmilkov:
Fixes #662

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/670)


tafsiri:
Review status: 0 of 15 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/math/math_test.ts, line 30 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4loc-vJ7dsycf9-ait:-L4loc-vJ7dsycf9-aiu:bt2gyzj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L30)):*
> ```XML
> {
>   const gpuTests: MathTests = it => {
>     it('scope returns Tensor', async math => {
> ```

Should probably be `tidy` instead of `scope`.  We could remove scope/tidy from here and add it to the describe call. 

---

*[src/math/math_test.ts, line 37 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lmP2zr5JGQ8jjtzCP:-L4lmP2zr5JGQ8jjtzCQ:b-aldsdn) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L37)):*
> ```XML
>         expect(dl.memory().numDataBuffers).toBe(2);
>         expect(dl.memory().numBytes).toBe(2 * 3 * 4);
>         await dl.tidy(async () => {
> ```

This is not part of this change, but i'm wondering if this callback needs to be `async`? It doesn't `await` anything in the body of the callback and appears to return immediately.

I also noticed this in multiple test cases below, so there may be something i am missing.

---

*[src/math/math_test.ts, line 52 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lnWt15bkQ88hTngPg:-L4lnWt15bkQ88hTngPh:b6xyu1s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L52)):*
> ```XML
> 
>         // a, b are still here, result should be disposed.
>         expect(dl.memory().numDataBuffers).toBe(2);
> ```

Should we also check that the bytes are back to 2 * 3 * 4?

---

*[src/math/math_test.ts, line 71 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lpHH0u3kxG4tkhwTX:-L4lpHH0u3kxG4tkhwTY:bpk2iw8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L71)):*
> ```XML
>     });
> 
>     it('scope returns Tensor[]', async math => {
> ```

scope -> tidy

---

*[src/math/math_test.ts, line 113 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lpdn45ay3Qv368mBI:-L4lpdn45ay3Qv368mBJ:bwkft1v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L113)):*
> ```XML
>     });
> 
>     it('scope returns Promise<Tensor>', async math => {
> ```

not related to this change, but I don't understand what this test wants to test, it doesn't explicitly test for a promise return and wrapping it in an async function will always make it return a promise regardless. 

---

*[src/math/tensor.ts, line 90 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ltD_QICcoc86G_kd9:-L4ltD_QICcoc86G_kdA:b-vixfc0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/tensor.ts#L90)):*
> ```XML
> }
> 
> /** We wrap data id since we use weak map to avoid memory leaks. */
> ```

I think this comment could use some elaboration. Its not clear how this wrapping aids your use of weakmaps or where you will be using weakmaps. 

On the typescript side, will this empty interface force a DataId to be a non-primitive object?

---

*[src/math/tensor.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lyEkGI9zxg4iqNYvG:-L4lyEkGI9zxg4iqNYvH:bf7aamb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/tensor.ts#L132)):*
> ```XML
>     this.dtype = dtype || 'float32';
>     this.strides = computeStrides(shape);
>     this.dataId = dataId != null ? dataId : {};
> ```

Would it ever help debug-ability to have an auto-incrementing integer inside this object?

---

*[src/math/backends/backend_engine.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lwbxIp0OJcf74PMGc:-L4lwbxIp0OJcf74PMGd:bc217em) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L132)):*
> ```XML
> 
>   registerVariable(v: Variable) {
>     if (this.registeredVariables[v.name] != null) {
> ```

Just checking that you intended to check against both `null` and `undefined` by using `!=`

---

*[src/math/backends/backend_engine.ts, line 139 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lx9IKBRUXSV9TnUE3:-L4lx9IKBRUXSV9TnUE4:b-h9innm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L139)):*
> ```XML
> 
>   disposeTensor(a: Tensor): void {
>     if (!this.refCounter.has(a.dataId)) {
> ```

Just checking my understanding. If `a` has previously been disposed will it still have a `dataId`?

---

*[src/math/backends/backend_engine.ts, line 157 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lxcwq-f1Fpx5z8gg4:-L4lxcwq-f1Fpx5z8gg5:b-avwr8y) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L157)):*
> ```XML
>   }
> 
>   memory(): {numDataBuffers: number; numBytes: number;} {
> ```

should this have a JSDoc comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670)*

dsmilkov:
Review status: 0 of 15 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/math/math_test.ts, line 37 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lmP2zr5JGQ8jjtzCP:-L4m1GO7BPGqd8eCJ7ae:b-766xpy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L37)):*

This is not part of this change, but i'm wondering if this callback needs to be `async`? It doesn't `await` anything in the body of the callback and appears to return immediately.

I also noticed this in multiple test cases below, so there may be something i am missing.
</blockquote></details>

nice catch. no need for await/async. this is probably copy/paste code that spilled over from other tests

---

*[src/math/math_test.ts, line 52 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lnWt15bkQ88hTngPg:-L4m37fKQ4Qw61e24lux:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L52)):*

Should we also check that the bytes are back to 2 * 3 * 4?
</blockquote></details>

Done.

---

*[src/math/math_test.ts, line 71 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lpHH0u3kxG4tkhwTX:-L4m3GNb-y-VYs93LXUG:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L71)):*

scope -> tidy
</blockquote></details>

Done.

---

*[src/math/math_test.ts, line 113 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lpdn45ay3Qv368mBI:-L4m3VWMzgXvtkbUIg1a:b7pbydk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L113)):*

not related to this change, but I don't understand what this test wants to test, it doesn't explicitly test for a promise return and wrapping it in an async function will always make it return a promise regardless. 
</blockquote></details>

Nice catch! Not sure how this test got here. Again, looks like a copy paste from another test :) `dl.tidy(f)` returns what `f` returns, whether that's promise or not is irrelevant.

---

*[src/math/backends/backend_engine.ts, line 157 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lxcwq-f1Fpx5z8gg4:-L4m4Pzdd2fXrs6rfrfS:b-5o9s7z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L157)):*

should this have a JSDoc comment?
</blockquote></details>

We have 1 jsdoc in Environment.ts where we wrap the static method in a @doc. This method is internal

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670)*

dsmilkov:
Review status: 0 of 15 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/math/math_test.ts, line 30 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4loc-vJ7dsycf9-ait:-L4m52CoTi7dWSdPgIpJ:bhh8qyp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L30)):*

Should probably be `tidy` instead of `scope`.  We could remove scope/tidy from here and add it to the describe call. 
</blockquote></details>

Done. Removed tidy prefix as it's already in the describe

---

*[src/math/tensor.ts, line 90 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ltD_QICcoc86G_kd9:-L4m8B7LPW3xquOYC79O:b-mh6dx5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/tensor.ts#L90)):*

I think this comment could use some elaboration. Its not clear how this wrapping aids your use of weakmaps or where you will be using weakmaps. 

On the typescript side, will this empty interface force a DataId to be a non-primitive object?
</blockquote></details>

Added comment. Re second point, making it empty interface allows primitives :) I changed it to `object` (lower-case, upper-case `Object` didn't work), to constraint it to non-primitives at compile-time. Note that users won't create dataid objects themselves, this is internal, but still useful for our own bugs.

---

*[src/math/tensor.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lyEkGI9zxg4iqNYvG:-L4m5K0D56_CClRKI462:b-lhal5d) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/tensor.ts#L132)):*

Would it ever help debug-ability to have an auto-incrementing integer inside this object?
</blockquote></details>

Ha, so I had the same thinking. So far, I haven't seen benefit from that auto-inc integer and couldn't think of one. If we see a benefit/need, we should bring it back (otherwise we save some memory not having it)

---

*[src/math/backends/backend_engine.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lwbxIp0OJcf74PMGc:-L4m5l-aQGdoFcEOaX4E:b-fthupe) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L132)):*

Just checking that you intended to check against both `null` and `undefined` by using `!=`
</blockquote></details>

Yup, we have a convention to never use `undefined` anywhere in our code base. We treat undefined and null as equivalent and always use `!=null`, `==null` for testing if things are there, and `===` , `!==` for anything primitive (string, numbers, etc). Our linter enforces that.

---

*[src/math/backends/backend_engine.ts, line 139 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lx9IKBRUXSV9TnUE3:-L4m6VrUp0P7t2new-Bd:b-i2ldqf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L139)):*

Just checking my understanding. If `a` has previously been disposed will it still have a `dataId`?
</blockquote></details>

yes. it will. disposing a tensor won't nullify its members. The whole `Tensor` object gets gc'ed once tensor as an object gets dereferenced (out of scope).

Do you think it helps the GC to nullify the members of Tensor when someone calls `tensor.dispose()`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670)*

dsmilkov:
FYI, per offline discussion, I delegated the memory stuff to be backend-specific, and the return signature of memory() is
{ numBytes, numDataBuffers, unreliable?} where cpu returns `unreliable: true`. The jsdoc in Environment.memory explains it.

nsthorat:
:lgtm_strong:

Nice work, this is super useful.

High level thoughts, non-blocking, is numBytes and numDataBuffers of only "allocated tensors" actually useful? What I actually want is "seriously, how much GPU memory am I actually using" not "what's the subset of memory that happens to be in use right now"? We're kind of lying :) 

---

Reviewed 4 of 15 files at r1, 7 of 8 files at r3.
Review status: 11 of 16 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/environment.ts, line 249 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ncNvEYgXbsRXv-2ei:-L4ncNvEYgXbsRXv-2ej:b-ebueml) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/environment.ts#L249)):*
> ```XML
>    * - `unreliable` optional, present if true. If the current backend can't
>    *    measure reliably (e.g. the cpu backend is unreliable due to automatic
>    *    garbage collection).
> ```

maybe add that this means that the actual numBytes and numDataBuffers would be strictly less than what we present.

---

*[src/math/math_test.ts, line 27 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4na_V0T0KidD0I4fnY:-L4na_V1yvmJV7G12x-Z:b-7h6fjp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/math_test.ts#L27)):*
> ```XML
> const gradientsScope = Gradients.gradientsScope;
> 
> // math.tidy
> ```

dl.tidy

---

*[src/math/math_test.ts, line 36 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ncpiVY-XsB0A74DPQ:-L4ncpiVY-XsB0A74DPR:b-nz1lw7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/math_test.ts#L36)):*
> ```XML
> 
>         expect(dl.memory().numDataBuffers).toBe(2);
>         expect(dl.memory().numBytes).toBe(2 * 3 * 4);
> ```

I wouldn't test numBytes here as this doesn't seem useful, will spill over everywhere in the codebase as people copy and paste, and will be brittle to changes in the way we manage memory (integer textures, for example)... and everywhere below.

I would leave the test for numBytes only for the dl.memory() unit tests.

---

*[src/math/tensor.ts, line 132 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lyEkGI9zxg4iqNYvG:-L4ndayhHiHG8WfLO5g7:b1kkgpu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/tensor.ts#L132)):*

Ha, so I had the same thinking. So far, I haven't seen benefit from that auto-inc integer and couldn't think of one. If we see a benefit/need, we should bring it back (otherwise we save some memory not having it)
</blockquote></details>

+1, not useful

---

*[src/math/tensor.ts, line 98 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nb8HI7xKPYPm7YC2h:-L4nb8HI7xKPYPm7YC2i:b-ksorqm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/tensor.ts#L98)):*
> ```XML
>  * Global_Objects/WeakMap
>  */
> export type DataId = object;
> ```

type this explicitly as {}

---

*[src/math/backends/backend_cpu.ts, line 56 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ne039QFS2AMvF4Qhi:-L4ne039QFS2AMvF4Qhj:b-z8cez6) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_cpu.ts#L56)):*
> ```XML
>   register(dataId: DataId, shape: number[], dtype: DataType): void {
>     if (this.data.has(dataId)) {
>       throw new Error(`data id ${dataId} already registered`);
> ```

this is just going to give you an opaque object in the error message, I would simply throw an error without ${dataId}

---

*[src/math/backends/backend_engine.ts, line 139 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4lx9IKBRUXSV9TnUE3:-L4nbo5NQltFWPU8w-uK:b-ymgn2g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/backends/backend_engine.ts#L139)):*

yes. it will. disposing a tensor won't nullify its members. The whole `Tensor` object gets gc'ed once tensor as an object gets dereferenced (out of scope).

Do you think it helps the GC to nullify the members of Tensor when someone calls `tensor.dispose()`?
</blockquote></details>

It helps the GC only in the sense that it will immediately clean up those values, but I wouldn't worry about that, and in fact it's quite nice to keep things like "shape" around as it helps identify the Tensor.

---

*[src/math/backends/backend_webgl.ts, line 85 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nehjGEBElFysmGzzp:-L4nehjGEBElFysmGzzq:b-78houu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_webgl.ts#L85)):*
> ```XML
>   register(dataId: DataId, shape: number[], dtype: DataType): void {
>     if (this.texData.has(dataId)) {
>       throw new Error(`data id ${dataId} already registered`);
> ```

same thing here, don't log ${dataId}

---

*[src/math/backends/backend_webgl.ts, line 88 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4neemZR1Zbv6nxWJYB:-L4neemZR1Zbv6nxWJYC:b9tdn8l) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_webgl.ts#L88)):*
> ```XML
>     }
>     this.numDataBuffers++;
>     this.numBytes += util.sizeFromShape(shape) * util.bytesPerElement(dtype);
> ```

can't we compute this only when you ask for the profile?

---

*[src/math/optimizers/momentum_optimizer_test.ts, line 42 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nfcnlyY-QNOSqTWFM:-L4nfcnlyY-QNOSqTWFN:bfvfrer) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/optimizers/momentum_optimizer_test.ts#L42)):*
> ```XML
>         w.reshape([1, 2]).matMul(x.reshape([2, 1])).add(b).sum() as Scalar;
> 
>     let numTensors = dl.memory().numDataBuffers;
> ```

what do you think about exposing "numTensors" on memory as well? We're kind of fast and loose with swapping them, but for unit tests I think it's more semantically correct to test the number of tensors, numDataBuffers here feels a little bit too into the guts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-:-L4ngddpvEJfLhfkUmzm:b-dipq93)*

tafsiri:
LGTM

To Nikhil's I thought numBytes would be more or less accurate because all the allocations and de-allocations on the GPU go through this (possibly with some constant overhead to set the thing up). Are there situations where we allocate GPU mem that don't go through the ref counter? 

---

Review status: 11 of 16 files reviewed at latest revision, 12 unresolved discussions, all commit checks successful.

---

*[src/environment.ts, line 249 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ncNvEYgXbsRXv-2ei:-L4pw_KKC4cM5yDaJ3Q6:b4ie53x) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/environment.ts#L249)):*

maybe add that this means that the actual numBytes and numDataBuffers would be strictly less than what we present.
</blockquote></details>

I thought we were going with reporting  numBytes and numBuffers from the CPU backend as -1 to discourage interpreting them (though i do like the addition of the unreliable flag). If not I wonder if we should document this function on the backend so that people know in what ways it is unreliable (we have a comment on the backend that it is cumulative but that wont be visible). Alternatively we could explicitly spell out when we expect unreliable to be true and what the numbers might mean here. 

---

*[src/math/math_test.ts, line 30 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4loc-vJ7dsycf9-ait:-L4mGWfnvAIS5fP4wbts:b9cqyd4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L30)):*

Done. Removed tidy prefix as it's already in the describe
</blockquote></details>

Non blocking comment but tidy is still here. I don't think its a big deal though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-:-L4pyYq3DwH1VYKoovGQ:bjypp3h)*

dsmilkov:
We allocate a bit more memory but that's deep in the internals of our texture manager, which recycles textures instead of disposing them, since creation of texture is slow. The user has no control over that, so it's unclear what the immediate value is in reporting it. If you want to really know how much GPU memory is used, you can use chrome profiler. The goal of dl.memory() is to provide something different than the chrome profiler - something the user is in direct control of, regardless of the environment/the backend.

---

Review status: 11 of 16 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

*[src/environment.ts, line 249 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ncNvEYgXbsRXv-2ei:-L4q9AjRtBMxnf0U1OR-:bmsga7r) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/environment.ts#L249)):*

I thought we were going with reporting  numBytes and numBuffers from the CPU backend as -1 to discourage interpreting them (though i do like the addition of the unreliable flag). If not I wonder if we should document this function on the backend so that people know in what ways it is unreliable (we have a comment on the backend that it is cumulative but that wont be visible). Alternatively we could explicitly spell out when we expect unreliable to be true and what the numbers might mean here. 
</blockquote></details>

I went with not reporting `-1` since we have a bunch of tests failing, since same test runs on both cpu and webgl backend and tests that high-level functions are disposing memory properly.

---

*[src/math/math_test.ts, line 30 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4loc-vJ7dsycf9-ait:-L4q9gL8bZ_gD-kMkGQ1:b1zn9xb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e1f72faa20468d85a2ce376f30cf97a12c602e66/src/math/math_test.ts#L30)):*

Non blocking comment but tidy is still here. I don't think its a big deal though.
</blockquote></details>

I forgot to push. tidy is gone.

---

*[src/math/math_test.ts, line 27 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4na_V0T0KidD0I4fnY:-L4q9dVOktpuoojltlSp:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/math_test.ts#L27)):*

dl.tidy
</blockquote></details>

Done.

---

*[src/math/math_test.ts, line 36 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ncpiVY-XsB0A74DPQ:-L4q9tGDn7c0QX9DapYQ:bszklsa) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/math_test.ts#L36)):*

I wouldn't test numBytes here as this doesn't seem useful, will spill over everywhere in the codebase as people copy and paste, and will be brittle to changes in the way we manage memory (integer textures, for example)... and everywhere below.

I would leave the test for numBytes only for the dl.memory() unit tests.
</blockquote></details>

Done. Great point.

---

*[src/math/tensor.ts, line 98 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nb8HI7xKPYPm7YC2h:-L4q9ufrgB0hJJqwJ7MQ:b-6gbuo4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/tensor.ts#L98)):*

type this explicitly as {}
</blockquote></details>

`object` is on purpose. If I do `{}`, that allows const a: DataId = 4; (primitive), while `object` forces DataId to be non-primitive. 

---

*[src/math/backends/backend_cpu.ts, line 56 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4ne039QFS2AMvF4Qhi:-L4qAHtV7CPnqIaNVU3-:b-4nowv7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_cpu.ts#L56)):*

this is just going to give you an opaque object in the error message, I would simply throw an error without ${dataId}
</blockquote></details>

Nice catch. Done.

---

*[src/math/backends/backend_webgl.ts, line 85 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nehjGEBElFysmGzzp:-L4qAmfjvBhqHcCUpM5Z:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_webgl.ts#L85)):*

same thing here, don't log ${dataId}
</blockquote></details>

Done.

---

*[src/math/backends/backend_webgl.ts, line 88 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4neemZR1Zbv6nxWJYB:-L4qB_mZ43QuuDpBBMpV:bbclyax) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/backends/backend_webgl.ts#L88)):*

can't we compute this only when you ask for the profile?
</blockquote></details>

can't iterate over the keys of the weakmap. if we kept the keys, it defeats auto gc.

---

*[src/math/optimizers/momentum_optimizer_test.ts, line 42 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-L4nfcnlyY-QNOSqTWFM:-L4qAx7KgjG57mwj2rhE:bv6bbjr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/097e8d492d16ea6f33855a7abd1e15ef72acd086/src/math/optimizers/momentum_optimizer_test.ts#L42)):*

what do you think about exposing "numTensors" on memory as well? We're kind of fast and loose with swapping them, but for unit tests I think it's more semantically correct to test the number of tensors, numDataBuffers here feels a little bit too into the guts.
</blockquote></details>

I left it out for now, following the logic of "lets prove that it is useful and then we'll add it". data buffers is useful since it's directly connected to memory usage. you can have 100 scalar tensors, each being a reshape of the other, and just 1 buffer of memory allocated.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-:-L4q8Dy5TJztoihTxLx6:b-16p9ei)*

nsthorat:
Just to respond to this (again, no action) - the user has complete control over it, tighter uses of tidy() will reduce the size of real allocation of memory. The chrome profiler will give the *entire applications* GPU memory usage, whereas this will tell you exactly how much GPU memory is used by textures.

---

Review status: 8 of 16 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/670#-:-L4qBrhMZrelC35JMvjQ:b-4v9brf)*

PR #683, opened 2018-02-09
Remove kernel_registry and call backends directly.

dsmilkov:
- Remove kernel_registry and call backends directly
- Allow for forward function to save tensors that will be needed by the backend function. We pass `dy` and the `saved` tensors to the backprop function.
- Remove logic in tape to handle multiple outputs. We can bring it back if needed.
- consolidate `tape_types.ts` and `tape_util.ts` into a single file `tape.ts`
- in debug mode, log the name of the `activeScope`, which should point to the last operation which called the kernel. The aggregation of these and showing hierarchical debug view is still a TODO.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/683)


nsthorat:
:lgtm_strong:

Awesome work, glad you're killing the tape types :)

Only major comment is in tensor.ts

---

Reviewed 49 of 79 files at r3.
Review status: 49 of 58 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 37 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e1llsZ0DcldtHXWIZ:-L5e1lltT7kBf0Yu1Yfj:b6fcdf7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L37)):*
> ```XML
> 
> export type ForwardFunc<T extends Tensor> =
>     (backend: KernelBackend, save?: <S extends Tensor>(tensor: S) => S) => T;
> ```

it'd be nice to document "save" somewhere, maybe here. Just say that these are tensors computed in the forward pass that we need in the backwards pass.

---

*[src/tape.ts, line 121 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5dxMcrCruy0veaGYhq:-L5dxMcrCruy0veaGYhr:b7lhnsl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/tape.ts#L121)):*
> ```XML
>       const prunedNode = Object.assign({}, node) as TapeNode;
>       prunedNode.inputs = prunedInputs;
>       prunedNode.output = prunedOutputs;
> ```

you can consolidate this block to:

`prunedNode.output = node.output`

and completely remove "prunedOutputs"

---

*[src/tensor.ts, line 598 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5dy8XLNK-qEnMCP7XT:-L5dy8XLNK-qEnMCP7XU:b-ofppty) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/tensor.ts#L598)):*
> ```XML
>     return ops.subStrict(this, x);
>   }
>   pow<T extends Tensor>(this: T, exp: Tensor): T {
> ```

this change seems a little weird to me - why are the chained ops now taking this as the first argument, and why not all of them?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-:-L5e1tRaqnLoA2QDVm09:bedg75s)*

nsthorat:
Review status: 49 of 58 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 102 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e2JIJV0dggwyAu54h:-L5e2JIJV0dggwyAu54i:b-9ubdek) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L102)):*
> ```XML
>       return x;
>     };
>     const kernelName = this.activeScope.name;
> ```

this is kind of funny, won't the active scope name be the operation name because of @operation and the scope around the operation, and not the actual kernelName?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683)*

nsthorat:
Review status: 49 of 58 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 102 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e2JIJV0dggwyAu54h:-L5e2sLfsaCe-OSLYNJf:b-8t37fh) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L102)):*

this is kind of funny, won't the active scope name be the operation name because of @operation and the scope around the operation, and not the actual kernelName?
</blockquote></details>

One idea is to keep passing a string for the kernel name just for logging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683)*

dsmilkov:
Review status: 48 of 58 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 37 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e1llsZ0DcldtHXWIZ:-L5eaobJYOC2mV-Qfzc2:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L37)):*

it'd be nice to document "save" somewhere, maybe here. Just say that these are tensors computed in the forward pass that we need in the backwards pass.
</blockquote></details>

Done.

---

*[src/engine.ts, line 102 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e2JIJV0dggwyAu54h:-L5ebC3cm_CCocdZxbRI:b-1p0jnj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L102)):*

One idea is to keep passing a string for the kernel name just for logging.
</blockquote></details>

Great observation. So I was thinking about this a little.

There will be a skew between kernels and user-ops and I think for now, before we have aggregation done, it's better to report user-ops since that is what the user controls (we report the inner-most named scope, with ability to aggregate).

Couple reasons for this argument:
- When we backpropagate, if we see that a tapeNode is missing gradFunc, (meaning we haven't implemented gradient yet), it's more useful to report an error that this `opName` is missing gradients, so we can't backpropagate, as opposed to this `kernelName` is missing grad, since from the user perspective, the culprit is `dl.opName()`, not `dl.kernelName()`.
- When reporting timing in debug mode, kernelNames might be confusing, e.g. imagine we reuse a conv2d kernel for higher-order convs, reporting the logical conv is more helpful than reporting conv2d (from the user perspective).

Once we have aggregation done, we can come back and decide maybe to report both kernel and op, at which point, we can add a required `name` argument to engine.executeKernel().

WDYT?

---

*[src/tape.ts, line 121 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5dxMcrCruy0veaGYhq:-L5e_vO9sBj5ZJ3KV83K:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/tape.ts#L121)):*

you can consolidate this block to:

`prunedNode.output = node.output`

and completely remove "prunedOutputs"
</blockquote></details>

Done.

---

*[src/tensor.ts, line 598 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5dy8XLNK-qEnMCP7XT:-L5e_xOjW3pbQpd7lnbn:b2vgivd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/tensor.ts#L598)):*

this change seems a little weird to me - why are the chained ops now taking this as the first argument, and why not all of them?
</blockquote></details>

`this: T` is not a real argument, it's a special signal to the TS compiler that if `this` is of type `T`, when `this.pow()` will return the same type. Previously we had no constraint that `a.pow()` will return the same rank as `a`. Now we do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683)*

nsthorat:
Review status: 46 of 58 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[src/engine.ts, line 102 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e2JIJV0dggwyAu54h:-L5fiopZBM9ng3W2yLHU:bkcvm1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L102)):*

Great observation. So I was thinking about this a little.

There will be a skew between kernels and user-ops and I think for now, before we have aggregation done, it's better to report user-ops since that is what the user controls (we report the inner-most named scope, with ability to aggregate).

Couple reasons for this argument:
- When we backpropagate, if we see that a tapeNode is missing gradFunc, (meaning we haven't implemented gradient yet), it's more useful to report an error that this `opName` is missing gradients, so we can't backpropagate, as opposed to this `kernelName` is missing grad, since from the user perspective, the culprit is `dl.opName()`, not `dl.kernelName()`.
- When reporting timing in debug mode, kernelNames might be confusing, e.g. imagine we reuse a conv2d kernel for higher-order convs, reporting the logical conv is more helpful than reporting conv2d (from the user perspective).

Once we have aggregation done, we can come back and decide maybe to report both kernel and op, at which point, we can add a required `name` argument to engine.executeKernel().

WDYT?
</blockquote></details>

That sounds reasonable to me. Maybe let's name this field to that effect: s/kernelName/scopeName or something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683)*

dsmilkov:
Review status: 46 of 58 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[src/engine.ts, line 102 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683#-L5e2JIJV0dggwyAu54h:-L5fk0HPy_GFtX24Jfq7:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/680f4f1dc42a5fc2157c5bf9c20a6a2ad69bb540/src/engine.ts#L102)):*

That sounds reasonable to me. Maybe let's name this field to that effect: s/kernelName/scopeName or something.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/683)*

PR #699, opened 2018-02-13
Fix small bug in mnist_eager demo

davidsoergel:
I think each iteration of the training loop should obtain one batch and then optimize over that, where currently the optimizer repeatedly sees new batches within its minimize() loop.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/699)

### DIFF
--- a/demos/mnist_eager/model.ts
+++ b/demos/mnist_eager/model.ts
@@ -25,9 +25,8 @@ const loss = (labels: dl.Tensor2D, ys: dl.Tensor2D) =>
 export async function train(data: MnistData, log: (message: string) => void) {
   const returnCost = true;
   for (let i = 0; i < TRAIN_STEPS; i++) {
+    const batch = data.nextTrainBatch(BATCH_SIZE);
     const cost = optimizer.minimize(() => {
-      const batch = data.nextTrainBatch(BATCH_SIZE);
-
       return loss(batch.labels, model(batch.xs));
     }, returnCost);
 


dsmilkov:
I think this change introduces a memory leak. `minimize(f)` calls `f` only once.  Moreover minimize wraps `f` in a `dl.tidy()` which cleans up memory after `f` finishes. This means that you want to have `data.nextTrainBatch()` inside `f` so its return (a batch of data) gets disposed at every step.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/699#-:-L5CF4AI9OGBEWtJlUuQ:b-tocp0l)*

davidsoergel:
Aha, good points.  I noticed this in the context of my Dataset variant of this demo, where data.nextTrainBatch() is async; that makes it very tricky/impossible to include it within the function passed to minimize().  In that context maybe I can wrap in another layer of tidy().  Here, I agree things are fine as is.  Thanks!

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/699#-:-L5CJBupROYTNMSg-9Hq:b57l117)*

dsmilkov:
You make a good point. We should allow async functions inside `minimize` as `nextBatch()` will be async - not sure how easy/tricky this is, but I think we can do it.

dsmilkov:
So if the user can't go backwards in the stream, only forward, then it might make sense to have something like:

```js
for (let i = 0; i < numSteps; i++) {
  batch = dataset.pointer();
  optimizer.minimize(() => model(batch));
  batch.advance(); // Cleans up old data, points to new data.
}
```

nsthorat:
That seems reasonable to me, @davidsoergel does this make sense to you, or is this diverging too much?

What if we had an "NDArrayStream" which was a decorator of sorts that applied to anything that generates `NDArray`s, any time you call `getNext()` the previous element is cleaned up.

PR #714, opened 2018-02-14
 Compute min and max of numerical columns over a Dataset.

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/714)


dsmilkov:
Yay! The ball is rolling. Left a few comments, some of them for my own understanding

---

Reviewed 1 of 5 files at r1.
Review status: 1 of 5 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/dataset.ts, line 50 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Ll_Se5jV-xzZDy1DV:-L5Ll_Se5jV-xzZDy1DW:bkdqp8z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/dataset.ts#L50)):*
> ```XML
>   abstract async getStream(): Promise<DataStream<DatasetElement>>;
> 
>   // TODO(soergel): plumb this through, to distinguish e.g. webcam from files
> ```

you can elaborate a bit what this means? Since this is user facing, let's move the elaborate jsdocs from internal methods to here.

---

*[src/contrib/data/statistics.ts, line 24 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LkDcBKLG_d4HqWwvK:-L5LkDcBKLG_d4HqWwvL:bqlo4p8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L24)):*
> ```XML
> import {ElementArray} from './types';
> 
> // TODO(soergel) This whole file is the barest stopgap.
> ```

for my own understanding, can you clarify what this means?

---

*[src/contrib/data/statistics.ts, line 26 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LkZlRCUjHZnpQWUv9:-L5LkZlRCUjHZnpQWUvA:b-upto1s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L26)):*
> ```XML
> // TODO(soergel) This whole file is the barest stopgap.
> 
> export class NumericColumnStatistics {
> ```

no need for this to be a class, classes have overhead. Make it an interface, of simply a type:
```ts
export type NumericColumnStatistics = {
  min: number;
  max: number;
}
```
do the pos_inf/neg_inf initialization inside the method that computes min/max

---

*[src/contrib/data/statistics.ts, line 40 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lqni_mRBG5CZAqfva:-L5Lqni_mRBG5CZAqfvb:b-3vc1zb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L40)):*
> ```XML
>  *
>  * @param min the lower bound of the inputs, which should be mapped to 0.
>  * @param min the upper bound of the inputs, which should be mapped to 1
> ```

typo: s/@param min/@param max/

---

*[src/contrib/data/statistics.ts, line 52 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlKb_i269xBGk9L6h:-L5LlKb_i269xBGk9L6i:b47szyh) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L52)):*
> ```XML
>     } else {
>       if (value instanceof Tensor) {
>         const result = value.sub(minTensor).div(maxTensor);
> ```

I think it should be `(x - min)/(max - min)` (that is, divide by the range=max-min). Here and below. Also update tests to reflect this, with cases where `min != 0`

---

*[src/contrib/data/statistics.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LquwSEkbLyuFzjYTB:-L5LquwSEkbLyuFzjYTC:bfa0hsf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L54)):*
> ```XML
>         const result = value.sub(minTensor).div(maxTensor);
>         return result;
>       } else if (value instanceof Array) {
> ```

for my own understanding, in which case (what part of the pipeline) would a value be `instanceof Array` and not an `ElementArray`

---

*[src/contrib/data/statistics.ts, line 66 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LpIVBo--f3YOhILIl:-L5LpIVBo--f3YOhILIm:bott8lv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L66)):*
> ```XML
>  * Gathers statistics from a Dataset (or optionally from a sample).
>  *
>  * Currently we gather only the minimum and maximum value from each numeric
> ```

let's remove this sentence and minimize things we need to keep in sync as code progresses

---

*[src/contrib/data/statistics.ts, line 76 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LqADe56Jk5z19Rbwh:-L5LqADf7X_ss1yn9BDw:b6oc0u3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L76)):*
> ```XML
>  * some way.
>  *
>  * A truly random shuffle of the stream would of course solve this
> ```

Let's move all of this jsdoc to the user-facing public API, which is dataset.getStats() i believe

---

*[src/contrib/data/statistics.ts, line 88 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LrLmvUOZdnzJGAmd5:-L5LrLmvUOZdnzJGAmd6:b-152gru) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L88)):*
> ```XML
>  * @param shuffleSize The size of the shuffle window to use, if any.  (Not
>  *   recommended, as described above).
>  * @param sampleSize The number of examples to take from the (possibly shuffled)
> ```

switch @param shuffleSize and sampleSize to align with the order in the method

---

*[src/contrib/data/statistics.ts, line 93 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LrWxYIagjEZEOQ0be:-L5LrWxYIagjEZEOQ0bf:bg7b7gm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L93)):*
> ```XML
> export async function computeStatistics(
>     dataset: Dataset, sampleSize?: number,
>     shuffleSize?: number): Promise<DatasetStatistics> {
> ```

how do you feel renaming this to shuffleWindowSize? or shuffleBufferSize?

---

*[src/contrib/data/statistics.ts, line 120 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LuqzqUTX9sprJxhcL:-L5LuqzqUTX9sprJxhcM:bm4pnbo) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L120)):*
> ```XML
>           recordMax = value.reduce((a, b) => Math.max(a, b));
>           recordMin = value.reduce((a, b) => Math.min(a, b));
>         } else {
> ```

let's make sure this is actually a numeric value. Check for number `!isNaN(value) && isFinite(value)` otherwise throw an error in the last `else` for unsupported data type

---

*[src/contrib/data/statistics_test.ts, line 17 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlVHWMUyz8k3dbc-X:-L5LlVHWMUyz8k3dbc-Y:b-d9o3vp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics_test.ts#L17)):*
> ```XML
>  */
> /*
> ```

why are tests commented out?

---

*[src/contrib/data/streams/data_stream.ts, line 122 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lw1dmOh6XYmnfwrh_:-L5Lw1dmOh6XYmnfwrha:b7imov7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L122)):*
> ```XML
>   async resolveFully(): Promise<void> {
>     let x = await this.next();
>     while (x !== undefined) {
> ```

do `!= null` here and elsewhere (captures both `null` and `undefined`)  - simplifies the world. we do this everywhere in our codebase and it would be good to be consistent.

---

*[src/contrib/data/streams/data_stream.ts, line 159 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LsgznXeaqMNdUu_qN:-L5LsgznXeaqMNdUu_qO:b-wu8zl6) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L159)):*
> ```XML
>    */
>   async forEach(f: (value: T) => void|Promise<void>): Promise<void> {
>     return this.map(f).resolveFully();
> ```

This uses MapStream, which seems to push the result of `f()` into an outputQueue, which means we'll have an array with N`undefined`values in it, is there a more efficient thing we can do than map. 

---

*[src/contrib/data/streams/data_stream.ts, line 411 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LyQru7-gktwWTzIWL:-L5LyQru7-gktwWTzIWM:bkhl7e6) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L411)):*
> ```XML
>     const item = await this.upstream.next();
>     if (item == null) {
>       return false;
> ```

Not in this PR. It's unnatural to skip null elements when mapping an array. We should be consistent with JS's map and Python's map behavior and pass nulls to transform as well - maybe the user wants to map null values to some reasonable defaults, or count the number of null values.

---

*[src/contrib/data/streams/data_stream.ts, line 417 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lu4RB8qiB5VABuL-r:-L5Lu4RB8qiB5VABuL-s:bniycvp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L417)):*
> ```XML
>       mapped = await mapped;
>     }
>     this.outputQueue.push(mapped);
> ```

Sorry if this is a silly question: why do `MapStream` or `FilterStream` need a queue? Not for this PR, but I think we should simplify the world and remove `QueueStream` and move that logic to BatchStream. The logic is so simple that is not worth distributing in a separate class, and makes it hard to follow. E.g. I thought outputQueue never gets cleaned, until I realized that this class extends QueueStream whose next() method does the unshift on outputQueue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-:-L5Lz8UqSK1EhaYiDOMw:bdwc55g)*

davidsoergel:
Yep, now that the basic framework is in place, there are lots of features to add!

Thanks for the review!  And sorry some parts were premature before. :S

---

Review status: 0 of 5 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/dataset.ts, line 50 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Ll_Se5jV-xzZDy1DV:-L5MInvKubiEXtst-0Y7:b-wioh1k) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/dataset.ts#L50)):*

you can elaborate a bit what this means? Since this is user facing, let's move the elaborate jsdocs from internal methods to here.
</blockquote></details>

Done-- elaborated this comment, and also moved the (unrelated) jsdocs on the following method.

---

*[src/contrib/data/statistics.ts, line 24 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LkDcBKLG_d4HqWwvK:-L5MPESfL3i96IbcgXVz:b35wl8g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L24)):*

for my own understanding, can you clarify what this means?
</blockquote></details>

Expanded the comment.

---

*[src/contrib/data/statistics.ts, line 26 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LkZlRCUjHZnpQWUv9:-L5MJjBH80NZPqoESnS8:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L26)):*

no need for this to be a class, classes have overhead. Make it an interface, of simply a type:
```ts
export type NumericColumnStatistics = {
  min: number;
  max: number;
}
```
do the pos_inf/neg_inf initialization inside the method that computes min/max
</blockquote></details>

Done.

---

*[src/contrib/data/statistics.ts, line 40 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lqni_mRBG5CZAqfva:-L5MJGUYNMKkMHwHdOSf:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L40)):*

typo: s/@param min/@param max/
</blockquote></details>

Done.

---

*[src/contrib/data/statistics.ts, line 52 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlKb_i269xBGk9L6h:-L5MnQ24vlckdrkxWksM:bdb10th) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L52)):*

I think it should be `(x - min)/(max - min)` (that is, divide by the range=max-min). Here and below. Also update tests to reflect this, with cases where `min != 0`
</blockquote></details>

Oh wow, that is so embarrassing.  I must have been really tired when I wrote that, and then the tests assumed min==0 to boot.  Gah!  All fixed now.

---

*[src/contrib/data/statistics.ts, line 54 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LquwSEkbLyuFzjYTB:-L5MPWX7cCiWHwkYqCeR:ba76exo) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L54)):*

for my own understanding, in which case (what part of the pipeline) would a value be `instanceof Array` and not an `ElementArray`
</blockquote></details>

ElementArray is a union type: number|number[]|Tensor|string, so this covers the number[] case.

---

*[src/contrib/data/statistics.ts, line 66 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LpIVBo--f3YOhILIl:-L5MJxGvRAKD9DiJ8MwB:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L66)):*

let's remove this sentence and minimize things we need to keep in sync as code progresses
</blockquote></details>

Done.

---

*[src/contrib/data/statistics.ts, line 76 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LqADe56Jk5z19Rbwh:-L5MJxfUzTu99kWUqlSi:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L76)):*

Let's move all of this jsdoc to the user-facing public API, which is dataset.getStats() i believe
</blockquote></details>

Done.

---

*[src/contrib/data/statistics.ts, line 88 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LrLmvUOZdnzJGAmd5:-L5MKZvy0SFwvWxsswzV:b-gwt350) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L88)):*

switch @param shuffleSize and sampleSize to align with the order in the method
</blockquote></details>

Oops, i went back and forth on this and left it inconsistent; it was actually the method argument order that was switched.  My logic was that people might want sampleSize often, and shuffleBufferSize rarely, so this order allows getStats(1000) as opposed to getStats(undefined, 1000) etc.  Fixed throughout.

---

*[src/contrib/data/statistics.ts, line 93 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LrWxYIagjEZEOQ0be:-L5MLNsWf5HWQRU07V8h:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L93)):*

how do you feel renaming this to shuffleWindowSize? or shuffleBufferSize?
</blockquote></details>

Done.

---

*[src/contrib/data/statistics.ts, line 120 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LuqzqUTX9sprJxhcL:-L5MQxQsgSF0wK_2IHbR:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics.ts#L120)):*

let's make sure this is actually a numeric value. Check for number `!isNaN(value) && isFinite(value)` otherwise throw an error in the last `else` for unsupported data type
</blockquote></details>

Done.

---

*[src/contrib/data/statistics_test.ts, line 17 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlVHWMUyz8k3dbc-X:-L5MLuVi1EJ9SvXS2Z9r:bbqvqx) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics_test.ts#L17)):*

why are tests commented out?
</blockquote></details>

Oops.  Git branch juggling / debugging artifact, sorry!

The tests pass--mostly.  I consistently get numerical errors on these tests and others, but only under certain WEBGL_ENVS.  I assume this is due to bugs in my (~4 year old) GPU (& drivers etc.).  Let's see if the Travis test passes.

---

*[src/contrib/data/streams/data_stream.ts, line 122 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lw1dmOh6XYmnfwrh_:-L5MMVUKKfYd-CQy60JK:b-t6nzle) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L122)):*

do `!= null` here and elsewhere (captures both `null` and `undefined`)  - simplifies the world. we do this everywhere in our codebase and it would be good to be consistent.
</blockquote></details>

Done.  There are a few more cases in unrelated code to fix in a separate CL.

---

*[src/contrib/data/streams/data_stream.ts, line 159 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LsgznXeaqMNdUu_qN:-L5MqfyWkgLQZo2RLYgU:b-gxx0e6) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L159)):*

This uses MapStream, which seems to push the result of `f()` into an outputQueue, which means we'll have an array with N`undefined`values in it, is there a more efficient thing we can do than map. 
</blockquote></details>

If I don't use a queue for MapStream, then this issue evaporates.  So, let's punt this question to another PR where I'll try that.

---

*[src/contrib/data/streams/data_stream.ts, line 411 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LyQru7-gktwWTzIWL:-L5MMjiYIiyKprsHCcV9:bdgrjog) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L411)):*

Not in this PR. It's unnatural to skip null elements when mapping an array. We should be consistent with JS's map and Python's map behavior and pass nulls to transform as well - maybe the user wants to map null values to some reasonable defaults, or count the number of null values.
</blockquote></details>

The DataStream API represents the end of the stream as 'undefined'.  This mechanism avoids requiring a separate call for hasNext() (or isEmpty(), etc.), which would produce race conditions.  Since we don't want to distinguish between null and undefined, it follows that a stream cannot contain null values, so they cannot be transformed or counted.  Happy to discuss!

Anyway, given that, here we're saying that if the upstream stream is exhausted, then nothing can be pumped.

---

*[src/contrib/data/streams/data_stream.ts, line 417 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5Lu4RB8qiB5VABuL-r:-L5MqGtbFGb3WsmK79_C:bwq3mc8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L417)):*

Sorry if this is a silly question: why do `MapStream` or `FilterStream` need a queue? Not for this PR, but I think we should simplify the world and remove `QueueStream` and move that logic to BatchStream. The logic is so simple that is not worth distributing in a separate class, and makes it hard to follow. E.g. I thought outputQueue never gets cleaned, until I realized that this class extends QueueStream whose next() method does the unshift on outputQueue.
</blockquote></details>

Good point-- I think map and filter probably don't need queues.  That approach was left over from my thinking about 1-to-many mappers, which would need a queue, but which I haven't implemented (map() is 1-to-1 so far).  I'll revisit this question in another PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-:-L5MoyQ31BPoHqlbzCmC:bo1ucd2)*

davidsoergel:
Review status: 0 of 5 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/statistics_test.ts, line 17 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlVHWMUyz8k3dbc-X:-L5OnjWNl0TcQtBXdXdK:b-qr9cif) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics_test.ts#L17)):*

Oops.  Git branch juggling / debugging artifact, sorry!

The tests pass--mostly.  I consistently get numerical errors on these tests and others, but only under certain WEBGL_ENVS.  I assume this is due to bugs in my (~4 year old) GPU (& drivers etc.).  Let's see if the Travis test passes.
</blockquote></details>

Hm, the tests still didn't run.  How can I cause $BROWSERSTACK_KEY to be set during the Travis build?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714)*

dsmilkov:
:lgtm_strong: Thanks for adding the new envs! Left a few responses, but no action needed for this PR. Feel free to merge. 

---

Reviewed 1 of 4 files at r2.
Review status: 1 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/statistics_test.ts, line 17 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LlVHWMUyz8k3dbc-X:-L5OucGYWTqnehpzjj-A:blcu5yu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/statistics_test.ts#L17)):*

Hm, the tests still didn't run.  How can I cause $BROWSERSTACK_KEY to be set during the Travis build?
</blockquote></details>

Talked offline. Future PRs should originate from branches in the dljs repo. As for failing tests, use webgl_float_envs for all tests in tf.contrib.data (we can change that later when we fix float packing)

---

*[src/contrib/data/streams/data_stream.ts, line 411 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LyQru7-gktwWTzIWL:-L5Ov-SFROauFCswxMdI:b-w6hzuw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L411)):*

The DataStream API represents the end of the stream as 'undefined'.  This mechanism avoids requiring a separate call for hasNext() (or isEmpty(), etc.), which would produce race conditions.  Since we don't want to distinguish between null and undefined, it follows that a stream cannot contain null values, so they cannot be transformed or counted.  Happy to discuss!

Anyway, given that, here we're saying that if the upstream stream is exhausted, then nothing can be pumped.
</blockquote></details>

Not for this PR. While js does have asynchronous behavior, it is single threaded, so we won't have race conditions. I feel strongly we should align with python/js behavior when it comes to .map and .filter. This means we should add `hasNext()` and avoid encoding `null/undefined` as end of stream. We want for users to be able to count missing values early in the pipeline when things are still primitive values.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-:-L5OwLTnZqh4eX7sPst6:bjzt2bf)*

davidsoergel:
Thanks!

---

Review status: 1 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/contrib/data/streams/data_stream.ts, line 411 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-L5LyQru7-gktwWTzIWL:-L5PVn6zXSHYnO3fIHNW:b-w78xdd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/49c1b5af079c76ea6a3cae083cd9ee249defe924/src/contrib/data/streams/data_stream.ts#L411)):*

Not for this PR. While js does have asynchronous behavior, it is single threaded, so we won't have race conditions. I feel strongly we should align with python/js behavior when it comes to .map and .filter. This means we should add `hasNext()` and avoid encoding `null/undefined` as end of stream. We want for users to be able to count missing values early in the pipeline when things are still primitive values.
</blockquote></details>

I did have both next() and hasNext() originally, but had to switch to this approach to avoid resulting bugs.  The problem is that we need next() (and hasNext()) to be async, so there's no guarantee that the stream state would not change between the hasNext() resolution and the next() resolution.  There's also no guarantee that the next() promises resolve in the order that they were created (especially given nested Promises).  The consequences that I observed were things like a) the order of elements in the stream could get scrambled, b) the stream could terminate prematurely, c) hasNext() might have no way of knowing what to return until all prior next() promises were resolved, etc.  So basically I think the standard JS map() behavior doesn't apply cleanly in the async case.  Happy to discuss other solutions (e.g., maybe taking the ChainedStream approach throughout would fix this).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/714#-:-L5PYh204FO5ECfU1e3m:b-ucfknh)*

PR #716, opened 2018-02-14
Zip example streams together.

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/716)


dsmilkov:
Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/contrib/data/dataset.ts, line 253 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UWVkJCPFIk53o4LlE:-L5UWVkK0zTmNvjOyvHk:bppox9h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/dataset.ts#L253)):*
> ```XML
> }
> 
> export function datasetFromZipped(datasets: Dataset[]) {
> ```

Looks like this method is not used in the PR? Is the idea to expose this to users (seems internal)? Missing dataset.zip() ?

---

*[src/contrib/data/streams/data_stream.ts, line 85 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UWk_vE7rLc2T0Hf3f:-L5UWk_vE7rLc2T0Hf3g:b-fbfxdp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L85)):*
> ```XML
>  * provided by streams later in the input list override those from earlier ones.
>  */
> export async function streamFromZipped(
> ```

Small suggestion: Might be better to name the function based on what it does, instead of what it returns (we can see from the typing that it returns a stream).

How about naming this `zipStreams()`. Similarly for other methods

---

*[src/contrib/data/streams/data_stream.ts, line 87 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UYPV83R1jfhIlWPya:-L5UYPV83R1jfhIlWPyb:b-7ry2u9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L87)):*
> ```XML
> export async function streamFromZipped(
>     streams: Array<DataStream<DatasetElement>>):
>     Promise<DataStream<DatasetElement>> {
> ```

What's the reasoning for returning a promise? We should push promises deeper to the lowest level possible (e.g. getNext() makes sense to return a promise), but not zip(). Likewise for other methods in contrib/data.

If we return a promise here, then we have to write:
```const result = (await (await streamFromZipped([a, b])).collectRemaining());```

instead, we should write:
```const result = await streamFromZipped([a, b]).collectRemaining();```

We followed this principle in dljs: you can chain operations as if they were synchronous, and only data() returns a promise:
```const meanSqLoss = await a.sub(b).square().mean().data();```
```


---

*[src/contrib/data/streams/data_stream.ts, line 452 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UXc6TCCDRml9DSCQx:-L5UXc6TCCDRml9DSCQy:bgnv1o5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L452)):*
> ```XML
> 
> class ZipperState {
>   constructor(public readonly item: DatasetElement) {}
> ```

this is super strange. What's the reason to have a special class that wraps a single element? You can just type the promise later
instead of `Promise<ZipperState>`, do Promise<DatasetElement> and

---

*[src/contrib/data/streams/data_stream.ts, line 461 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5U_W6LDTJh-v7xFeHy:-L5U_W6M1VZF5uRfPZfR:b-q78485) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L461)):*
> ```XML
>  *
>  * The result of the zip operation is simply to merge the underlying
>  * `DatasetElement`s, so their keys should be disjoint.  If they are not, values
> ```

regarding merging keys, does this match the behavior of tf.dataset.zip?

---

*[src/contrib/data/streams/data_stream.ts, line 462 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5U_oxt2Y584GrAGyAD:-L5U_oxt2Y584GrAGyAE:b-gq7z8f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L462)):*
> ```XML
>  * The result of the zip operation is simply to merge the underlying
>  * `DatasetElement`s, so their keys should be disjoint.  If they are not, values
>  * provided by streams later in the input list override those from earlier ones.
> ```

Took some time to digest. Maybe make it more explicit, smth like:
when calling dataset.zip((a, b)), b's values will override a's values for the matching key, since b comes later.

---

*[src/contrib/data/streams/data_stream.ts, line 488 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5Ubl-a637CPkKSIPR1:-L5Ubl-a637CPkKSIPR2:bfxviw7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L488)):*
> ```XML
>       if (elements.every(x => x == null)) {
>         // The streams ended simultaneously, as expected
>         return new ZipperState(undefined);
> ```

use null instead.

---

*[src/contrib/data/streams/data_stream.ts, line 498 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5Ud-n53wB6cwH1P_RD:-L5Ud-n53wB6cwH1P_RE:b-61d7em) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L498)):*
> ```XML
>       // TODO(soergel): error on colliding keys.  Is checking worth the cost?
>       // As written here, later values for the same key overwrite earlier ones.
>       Object.assign(mergedElement, e);
> ```

what happens if I want to zip primitives (early in the pipeline):
that is zip([1, 2, 3], [4, 5, 6])  returns `[(1, 4), (2, 5), (3, 6)]`. I think here we'll have a problem since both objects have the key 0, which does collision. We should align with [tf.data](https://www.tensorflow.org/api_docs/python/tf/contrib/data/Dataset#zip) and python , which seems to support cases like this

---

*[src/contrib/data/streams/data_stream_test.ts, line 241 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UY0OZ8Zm4Oecc76XX:-L5UY0OZ8Zm4Oecc76XY:btvpyq3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream_test.ts#L241)):*
> ```XML
>     const a = new TestIntegerStream().map(x=>({'a':x}));
>     const b = new TestIntegerStream().map(x=>({'b':x}));
>     const readStreamPromise = streamFromZipped([a, b]);
> ```

take advantage of await? will improve readability:
const readStream = await streamFromZipped([a, b]);

Likewise for all other tests in here.

---

*[src/contrib/data/streams/data_stream_test.ts, line 251 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716#-L5UcrwE0p1xdmqlbjLB:-L5UcrwE0p1xdmqlbjLC:b-dqwzju) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream_test.ts#L251)):*
> ```XML
>         .then(done)
>         .catch(done.fail);
>   });
> ```

add a test that demonstrates that
zip([1, 2, 3], [4, 5, 6]) returns [(1, 4), (2, 5), (3, 6)]

also a test that shows zipping streams with multiple elements, and a test that expects error when zipping streams with misaligned length

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/716)*
<!-- Sent from Reviewable.io -->

nsthorat:
Review status: 0 of 3 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/dataset.ts, line 255 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/716#-L5VnD2gYhOi-ZtJ_5Az:-L5VnD2gYhOi-ZtJ_5B-:b3m67qm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/dataset.ts#L255)):*
> ```XML
> export function datasetFromZipped(datasets: Dataset[]) {
>   return datasetFromStreamFn(async () => {
>     const streams = await Promise.all(datasets.map((d) => d.getStream()));
> ```

remove parens from d

---

*[src/contrib/data/streams/data_stream.ts, line 87 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/716#-L5UYPV83R1jfhIlWPya:-L5Vsko-87RAd7lm5JRS:btxzlw9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L87)):*

What's the reasoning for returning a promise? We should push promises deeper to the lowest level possible (e.g. getNext() makes sense to return a promise), but not zip(). Likewise for other methods in contrib/data.

If we return a promise here, then we have to write:
```const result = (await (await streamFromZipped([a, b])).collectRemaining());```

instead, we should write:
```const result = await streamFromZipped([a, b]).collectRemaining();```

We followed this principle in dljs: you can chain operations as if they were synchronous, and only data() returns a promise:
```const meanSqLoss = await a.sub(b).square().mean().data();```
</blockquote></details>

+1 to this comment, I think this is the most important.

Sorry for coming to the party late, but in general, the API to the user shouldn't have multiple awaits. The final "next" or "collectRemaining" should return a promise, but creation of things in general shouldn't (e.g. CSVDataset.create.. internally this can do something in an asynchronous fashion but the user shouldn't await it. When the user awaits the final result, it will internally await multiple promises).

---

*[src/contrib/data/streams/data_stream.ts, line 471 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/716#-L5VqgWMm2mSc9xAjZ7g:-L5VqgWMm2mSc9xAjZ7h:b-3rn1as) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L471)):*
> ```XML
> class ZipperStream extends DataStream<DatasetElement> {
>   private currentPromise: Promise<ZipperState>;
>   count = 0;
> ```

can this be private?

---

*[src/contrib/data/streams/data_stream.ts, line 498 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/716#-L5Ud-n53wB6cwH1P_RD:-L5Vq8x9aaNDKsLYBN4y:b-3cw25j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4fc584f8c9540f43276283ad731029e1b3ee30af/src/contrib/data/streams/data_stream.ts#L498)):*

what happens if I want to zip primitives (early in the pipeline):
that is zip([1, 2, 3], [4, 5, 6])  returns `[(1, 4), (2, 5), (3, 6)]`. I think here we'll have a problem since both objects have the key 0, which does collision. We should align with [tf.data](https://www.tensorflow.org/api_docs/python/tf/contrib/data/Dataset#zip) and python , which seems to support cases like this
</blockquote></details>

+1, I think this is important. Or even zipping two objects into a tuple, like a Tensor (which currently will try to merge the objects).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/716)*

davidsoergel:
Abandoning per #873, but I'll revisit the outstanding comments here when recreating this PR against the new repo.

PR #718, opened 2018-02-14
Make contrib/data tests way faster

### LINKED ISSUE #704: Bring back default jasmine timeout limit for some dl.data tests
See #701 for context

davidsoergel:
Haha, I had forgotten about the bit where I intentionally made these tests super slow.  Oops!

Thanks for the timer warnings-- that was super helpful.  There's one ShuffleStream test that still barely triggers the warning on my (old & slow) machine (but only in FireFox).  I think this is tolerable.

Fixes #704.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/718)


dsmilkov:
:lgtm_strong: Thank you! As a general principle, we should try to make tests as short as possible while testing the desired functionality. The 500ms limit I chose is a starting point. If we can make the test run in 1ms instead of 20ms, it's still better. I'll be at some point in the future pushing this limit lower, as we add more tests and environments (which have multiplicative effect on # of tests), to keep eng productivity high

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/718#-:-L5Oxg_VWGUCmgWps8LH:b830yne)*

PR #752, opened 2018-02-19
Add check for empty xs list in Engine.gradients

caisq:
* Throw an Error immediately if an empty xs list is received.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/752)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -304,9 +304,7 @@ export class Engine implements TensorManager {
   gradients<T extends Tensor>(
       f: () => T, xs: Tensor[], dy?: T,
       allowNoGradients = false): {value: T, grads: Tensor[]} {
-    if (xs.length === 0) {
-      throw new Error('gradients() received an empty list of xs.');
-    }
+    util.assert(xs.length > 0, 'gradients() received an empty list of xs.');
 
     return tidy('gradients', () => {
       const y = f();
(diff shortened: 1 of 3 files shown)


src/engine.ts
```diff
@@ -304,6 +304,10 @@ export class Engine implements TensorManager {
   gradients<T extends Tensor>(
       f: () => T, xs: Tensor[], dy?: T,
       allowNoGradients = false): {value: T, grads: Tensor[]} {
+    if (xs.length === 0) {
```
ManrajGrover: We can make use of `util.assert` here.

src/gradients.ts
```diff
@@ -290,7 +290,15 @@ export class Gradients {
       }
     }
     // Prune non-trainable variables.
+    const originalVarCount = varList.length;
     varList = varList.filter(variable => variable.trainable);
+    if (varList.length === 0) {
```
ManrajGrover: We can make use of `util.assert` here as well.

dsmilkov:
:lgtm_strong: Thank you! Feel free to submit after addressing Manraj's comments.

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/752#-:-L5iPDcVYXOBQcNqZu4B:b-t2cg8t)*

PR #762, opened 2018-02-20
Make tidy(Promise) a warning and speed up schediling of WebGL programs

dsmilkov:
Webgl programs are sharing the same vertex program, so we don't need to bind vertex attributes every time we run a program.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/762)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/762#-:-L5tC5iqfJLkHOZgr0vd:b-jbsf8z)*

PR #763, opened 2018-02-21
Make tidy synchronous - do not allow tidy(f) where f returns a Promise

dsmilkov:
Also fix a bug: dl.getBackend() on already initialized environment should be a no-op

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/763)

PR #771, opened 2018-02-22
Rename streams to async_streams.

davidsoergel:
This is just a rename, in preparation for introducing contrib/data/sync_streams.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/771)


caisq:
Should it be called `asyncStream` to be consistent with the camel case naming in most of the other places?
https://deeplearnjs.org/docs/api/index.html

davidsoergel:
I just renamed the directory; snake case is consistent with filenames throughout DLJS.  The class names are still camelcase-- i.e. I didn't rename DataStream here (though I agree that may later become AsyncStream or something).

nsthorat:
Just curious - will one replace the other or will these two work together? Obviously the difference is async vs sync, but what are the different use cases?

davidsoergel:
They'll work together.

Anything involving Tensors has to be sync for the sake of GC.  That's because a) #763 disallows async tidy (for good reasons, I know), and b) we can't guarantee that all Tensors will be manually disposed because they'll be created in user code (via callbacks e.g. stream.map(f)) and then passed downstream, where that user code can't track them.

On the other hand many data sources are intrinsically async (e.g., reading a file or url). 

So the main async use case is "obtain (non-Tensor) data, do something with it, and emit a non-Tensor output"; this will generally be the outer loop of a data-stream-based program.  Within that async stream, there may be a map step, where the mapping function is an entire sync stream.  Inside that sync stream we can create Tensors, manipulate them, train and predict, etc., and every iteration can still be tidied.

Thus a likely scenario is that the interesting manipulations (math, filter, batch, etc.) happen in the sync phase, so the async streams may really need only next() and map().  I'm leaving all the other chainable manipulations in place for now since they're already implemented (and could still come in handy for non-Tensor data).  But, we may end up deleting unused ones once the sync counterparts are available.

<hr>
PS. I just thought of a way to maybe do manual GC in the async context after all (fixing b above).  Going to try it now; if that works we won't need this PR or the sync streams at all.

nsthorat:
Cool, let me know.

Also just curious if Sync vs Async is just a detail, and doesn't need to be separated into two parallel directories?

Tensor operations themselves have to be synchronous, and in general adding things like "filter" or "map" as data transformations should finish immediately, but next() doesn't have to be synchronous (e.g. just the part where you do tensor operations are in sync tidys(), or you simply use async functions to fetch and call dispose() on them).

Maybe we can sync up again over VC on Monday?

davidsoergel:
Abandoning this, after much mental churn, in favor of #785.

PR #775, opened 2018-02-23
Loss Operations: Adds Absolute Difference loss

ManrajGrover:
This PR adds absolute difference loss to the api.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/775)


nsthorat:
Sorry for the delay! Few comments inline.

---

Review status: 0 of 5 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/ops/loss_ops.ts, line 65 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6YF_cCFybGXsqWozZP:-L6YF_cCFybGXsqWozZQ:bw4xdb3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c34311547d6a5d23f1cc7aea84b6aab5dd5bb811/src/ops/loss_ops.ts#L65)):*
> ```XML
> 
>   /**
>    * Adds an Absolute Difference loss to the training procedure.
> ```

I know this is what TF says, but they have a Graph model, how about:

`Computes the absolute difference loss between two tensors.`

---

*[src/ops/loss_ops.ts, line 33 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwFXOZkAvK90KYYxs:-L6mwFXOZkAvK90KYYxt:bwuoq3t) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L33)):*
> ```XML
> export class LossOps {
>   /**
>    * Computes the weighted loss.
> ```

between two tensors

---

*[src/ops/loss_ops.ts, line 35 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwNVDkpv7iFATZQcP:-L6mwNVDkpv7iFATZQcQ:bmy08fl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L35)):*
> ```XML
>    * Computes the weighted loss.
>    *
>    * @param losses `Tensor` of shape `[batch_size, d1, ... dN]`.
> ```

remove ticks around Tensor, we parse the type automatically in the docs

---

*[src/ops/loss_ops.ts, line 38 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwS6ltpyRoDj4Ftwo:-L6mwS6ltpyRoDj4Ftwp:byn16po) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L38)):*
> ```XML
>    * @param weights `Tensor` whose rank is either 0, or the same rank as
>    *    `losses`, and must be broadcastable to `losses` (i.e., all
>    * dimensions must be either `1`, or the same as the corresponding
> ```

indent these 2 lines 4 more

---

*[src/ops/loss_ops.ts, line 44 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwoeGJ2qyRN0IxV5j:-L6mwoeGJ2qyRN0IxV5k:b-jyg7dx) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L44)):*
> ```XML
>   @operation
>   static computeWeightedLoss<T extends Tensor, O extends Tensor>(
>       losses: T, weights?: Tensor, reduction = Reduction.NONE): O {
> ```

TF has a default of Reduction.SUM_BY_NONZERO_WEIGHTS, can we keep that as well?

---

*[src/ops/loss_ops.ts, line 45 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwDO1PvPeiTe4xgiB:-L6mwDO2ZnZ0OE3nXB75:b-wgie91) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L45)):*
> ```XML
>   static computeWeightedLoss<T extends Tensor, O extends Tensor>(
>       losses: T, weights?: Tensor, reduction = Reduction.NONE): O {
>     if (weights === undefined) {
> ```

use `== null`

---

*[src/ops/loss_ops.ts, line 46 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mwXUf1StYX62Y46KS:-L6mwXUf1StYX62Y46KT:bdtcxeq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L46)):*
> ```XML
>       losses: T, weights?: Tensor, reduction = Reduction.NONE): O {
>     if (weights === undefined) {
>       weights = ops.scalar(1);
> ```

instead of doing it like this, can we conditionally do the ".mul" below? We will save a shader call and a texture.

---

*[src/ops/loss_ops.ts, line 80 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mxNxOIXTscKskAIqu:-L6mxNxPZ5p06qpt1aHZ:bdn9tmb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/loss_ops.ts#L80)):*
> ```XML
>   @operation
>   static absoluteDifference<T extends Tensor, O extends Tensor>(
>       labels: T, predictions: T, weights?: Tensor, reduction = Reduction.NONE):
> ```

TF has default of SUM_BY_NONZERO_WEIGHTS, can we keep that?

---

*[src/ops/lossop_tests.ts, line 3 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mxcmUWqgKy9hUtfO0:-L6mxcmUWqgKy9hUtfO1:bb6ht2i) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/lossop_tests.ts#L3)):*
> ```XML
> /**
>  * @license
>  * Copyright 2017 Google Inc. All Rights Reserved.
> ```

2018

---

*[src/ops/lossop_tests.ts, line 134 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L6mxmOEIBBo_T3vkMyY:-L6mxmOEIBBo_T3vkMyZ:b-qzjwbq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/ca76a4b0369dae3d16012541aa74113d6bc97ced/src/ops/lossop_tests.ts#L134)):*
> ```XML
>          Math.abs(3 - 6) * 2) /
>             20);
>   });
> ```

would you mind also testing computeWeightedLoss directly too?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-:-L6mxp2dTnfQ4s2biJKR:b9k8v5x)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/ops/loss_ops.ts, line 61 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-L7KfkTtU5LWdyy6xUH7:-L7KfkTtU5LWdyy6xUH8:bitjzkg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/b6524c6bbee287e38719a82c2f00351044acea18/src/ops/loss_ops.ts#L61)):*
> ```XML
>         loss = loss.div(ops.onesLike(losses).mul(weights).sum()) as O;
>       } else if (reduction === Reduction.SUM_BY_NONZERO_WEIGHTS) {
>         const mask = ops.where(
> ```

qq, in sum by non-zero weights, do we need to do all this? doesn't a multiplication by a weight of 0 do the same thing as manually checking?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/775#-:-L7KfqlZP1kcmfJ9yjmv:b-jbsf8z)*

ManrajGrover:
> qq, in sum by non-zero weights, do we need to do all this? doesn't a multiplication by a weight of 0 do the same thing as manually checking?

Hi, I'm not sure if I get it. Could you please elaborate more on it?

dsmilkov:
Left a few comments , but generally looks great!

---

Reviewed 1 of 2 files at r3, 1 of 2 files at r5.
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/ops/loss_ops.ts, line 55 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yFcEn43xRk-tRdC5a:-L8yFcEo9TEA7gfsy_RW:b-7592yj) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/loss_ops.ts#L55)):*
> ```XML
> 
>     if (reduction === Reduction.NONE) {
>       loss = weightedLoss as O;
> ```

`return weightedLoss;` here directly.

and remove the else { } and un-indent the code in the else.

---

*[src/ops/loss_ops.ts, line 58 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yGsstCOvTDzjVBSn3:-L8yGsstCOvTDzjVBSn4:b-w85oat) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/loss_ops.ts#L58)):*
> ```XML
>     } else {
>       loss = weightedLoss.sum() as O;
>       if (reduction === Reduction.MEAN) {
> ```

Make it explicit `if (reduction === Reduction.SUM) { return loss}`

---

*[src/ops/loss_ops.ts, line 59 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yH0Ib0QtkPVzTJPHX:-L8yH0Ib0QtkPVzTJPHY:b2iwwwm) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/loss_ops.ts#L59)):*
> ```XML
>       loss = weightedLoss.sum() as O;
>       if (reduction === Reduction.MEAN) {
>         loss = loss.div(ops.onesLike(losses).mul(weights).sum()) as O;
> ```

can this be simplified to `return loss.div(weights.sum())` ? 

---

*[src/ops/loss_ops.ts, line 63 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yHXBu7x3ui7X4Sfo7:-L8yHXBu7x3ui7X4Sfo8:b67rmlz) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/loss_ops.ts#L63)):*
> ```XML
>         const mask = ops.where(
>             weights.equal(ops.scalar(0)), ops.zerosLike(weights),
>             ops.onesLike(weights));
> ```

can this be simplified to:
```
const numNonZeros = weights.notEqual(0).sum();
return loss.div(numNonZeros);
``` 

---

*[src/ops/lossop_tests.ts, line 1 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yITu77BR4Y2yggFj9:-L8yITu77BR4Y2yggFjA:biuepop) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/lossop_tests.ts#L1)):*
> ```XML
> /**
> ```

rename this file to loss_ops_tests.ts so that the prefix `loss_ops` matches the file where the implementation is.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-:-L8yIkeq1DV02ji2N1Fk:bbxzo27)*

ManrajGrover:
@dsmilkov Addressed the comments

dsmilkov:
Reviewed 1 of 6 files at r6.
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/ops/loss_ops.ts, line 48 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91_Lr_9DeEz_607qHF:-L91YfhGEBP14_6RB6ey:bsny1zw) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L48)):*
> ```XML
>       reduction = Reduction.SUM_BY_NONZERO_WEIGHTS): O {
>     if (weights == null) {
>       weights = ops.onesLike(losses);
> ```

Keep weights null and check for their value inside the if statements so we can save shader calls and improve numerical stability. see comments below.

---

*[src/ops/loss_ops.ts, line 51 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91YPY24w9AOMSw5t0q:-L91YPY24w9AOMSw5t0r:bwr4xma) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L51)):*
> ```XML
>     }
> 
>     const weightedLoss = losses.mul(weights);
> ```

we can save a shader call if we check that weights are null here , that is:
`const weightedLoss = (weights == null) ? losses : losses.mul(weights);

---

*[src/ops/loss_ops.ts, line 52 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91XjN0A9461wAV-dzR:-L91XjN0A9461wAV-dzS:b78b3w) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L52)):*
> ```XML
> 
>     const weightedLoss = losses.mul(weights);
>     let loss;
> ```

Remove this loss variable. See comments below.

---

*[src/ops/loss_ops.ts, line 57 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91cAZQ5FoufGmUYUKn:-L91ZSabDBC-MMWmIbLn:bk5xaad) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L57)):*
> ```XML
>       return weightedLoss as O;
>     } else {
>       loss = weightedLoss.sum() as O;
> ```

Dont' precompute the sum of weightedLoss, since we won't use it in the Reduction.MEAN case. Do it inside the if statement. That is:
if (reduction == Reduction.SUM) {
  return weightedLoss.sum();
}
if (reduction == Reduction.MEAN) {
  // see comment below 
}

---

*[src/ops/loss_ops.ts, line 62 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91_cC42VbIIi6MPzK_:-L91_cC42VbIIi6MPzKa:b-ofor67) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L62)):*
> ```XML
>         return loss;
>       } else if (reduction === Reduction.MEAN) {
>         return loss.div(weights.sum());
> ```

`return (weights == null) ? weightedLoss.mean() : weightedLoss.sum().div(weights.sum())`

note we want mean() directly, since it avoids the potential overflow that x.sum().div(n) has. Also note we call weightedLoss.sum(), since the temp `loss` variable should go away.

---

*[src/ops/loss_ops.ts, line 65 at r6](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L91ciKaB9VBNFpJAlQl:-L91ciKaB9VBNFpJAlQm:b-71t3jn) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fec7077437e91d43a337cd0ccd076ef9bcfba3f1/src/ops/loss_ops.ts#L65)):*
> ```XML
>       } else if (reduction === Reduction.SUM_BY_NONZERO_WEIGHTS) {
>         const numNonZeros = weights.notEqual(ops.scalar(0)).sum();
>         return loss.div(numNonZeros);
> ```

weightedLoss.sum().div(numNonZeros)

---

*[src/ops/lossop_tests.ts, line 1 at r5](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-L8yITu77BR4Y2yggFj9:-L91crqZD-g3CIg3Dr3F:bxfzw8c) ([raw file](https://github.com/tensorflow/tfjs-core/blob/c5c7950bfbf594beea9dc1ecfa70ac8561c2e798/src/ops/lossop_tests.ts#L1)):*

rename this file to loss_ops_tests.ts so that the prefix `loss_ops` matches the file where the implementation is.
</blockquote></details>

My appologies for the typo: the filename must end with test.ts not tests.ts for the actual test runner to pick it up.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775)*

dsmilkov:
Thanks. Took another pass. There is a couple of small improvements we can do regarding perf and numerical stability, but this looks great already!

ManrajGrover:
@dsmilkov In `Reduction.SUM_BY_NONZERO_WEIGHTS`, if weights are undefined, `weights.notEqual` will throw an error. I'm not sure what to do in this case since we are not assigning weights as 1. I don't think we can totally get away from saving a shader call. Or can we? 

Tensorflow defaults to 1 in this case.

Also, Typescript is complaining about not returning anything at the end. Should we return `weightedLoss` as default?

dsmilkov:
When `weights` is `undefined`, you don't need to call `weights.notEqual()`. The number of non-zero weights is `losses.size` since all the weights are implicitly `1` of the same shape as `losses`

ManrajGrover:
@dsmilkov I had thought about throwing an error but then I realized error will automatically be thrown if anything other than `Reduction` is passed to `reduction` param.

dsmilkov:
:lgtm_strong: Great work Manraj!!

---

Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/775#-:-L9PjN_W5al9h4RxIHSu:b-10w32u)*

dsmilkov:
Hi Manraj,

Few tests are failing when I synced to master (see [log](https://travis-ci.org/tensorflow/tfjs-core/builds/363128118#L8978)). Nothing major, I think it's related to the fact that we made tf.div(a, b) be strict and expect a and b to be of the same dtype, to align with TF Python. Ping me when you fix those and I'll merge! Thank you!

PR #779, opened 2018-02-25
Resize input image in squeezenet demo

jbencook:
For some reason my browser is pulling in the cat image as a 226x226 image causing the squeezenet demo to fail. This change forces the input size to be 227x227. This fixes the demo for me and should make the required image size more explicit for other newcomers.

screen shot 2018-02-25 at 1 35 54 pm

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/779)

### DIFF
--- a/models/squeezenet/demo.html
+++ b/models/squeezenet/demo.html
@@ -22,8 +22,9 @@
     resultElement.innerText = 'Predicting...';
 
     const pixels = dl.fromPixels(cat);
+    const resized = dl.image.resizeBilinear(pixels, [227, 227]);
 
-    const result = squeezeNet.predict(pixels);
+    const result = squeezeNet.predict(resized);
 
     resultElement.innerText = '';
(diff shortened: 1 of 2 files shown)


jbencook:
I just realized someone else submitted [this fix](#489) a couple months ago. I'm happy to close my PR but the current state is pretty confusing if you're checking out this package for the first time. I agree with [their comment](https://github.com/PAIR-code/deeplearnjs/pull/489#issuecomment-354596818) that resizing inside the model is not ideal. But I added a more informative error in case someone else passes the wrong size image in.

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/779#-:-L6IuJE2trTabENvomRM:b-jbsf8z)*

PR #785, opened 2018-02-26
Dispose Tensors as needed during async stream processing.

davidsoergel:
This approach differs from the prior plan to introduce synchronous streams.  We can still discuss different designs, but this PR at least makes the current design work for now.

The idea is that streams remain fully async, pull-only, source streams.  Because next() is async, it is not possible to include it within a dl.tidy().  At the same time, Tensors returned from next() must be protected from being inadvertently tidied.  Thus, all Tensors returned from a stream have already had dl.keep() called on them, and so must be manually disposed.

This PR ensures that all Tensors produced as intermediate results during pipeline operation are automatically disposed, so that the only ones that the user is responsible for are those returned from next().

Note this disallows passing async user functions to map(), filter(), and forEach().  Those user functions will be automatically wrapped in dl.tidy().

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/785)


nsthorat:
Happy to chat more about this in person, would be good to understand next steps for the memory management stuff.

---

Review status: 0 of 7 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/batch_dataset.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JFhjjRab2ID2gI5Gv:-L6JFhjjRab2ID2gI5Gw:br668w1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset.ts#L19)):*
> ```XML
>  */
> 
> import * as dl from '../../index';
> ```

this will introduce a cycle, import keep directly from tracking.ts

---

*[src/contrib/data/batch_dataset_test.ts, line 18 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JGNBLnFESiFKxcEXh:-L6JGNBLnFESiFKxcEXi:byy0yjv) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L18)):*
> ```XML
>  */
> 
> import {DatasetBatch} from '..';
> ```

import from the file directly

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6JH3Tl0yUMqCcRKLfn:b77wj8h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*
> ```XML
> });
> 
> function consume(input: DatasetBatch): void {
> ```

can we put this in some common util somewhere, we seem to use it in several places?

And can we maybe just call this disposeBatch or something to that effect?

---

*[src/contrib/data/dataset.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHcZcB1uUHlpBpyxj:-L6JHcZcB1uUHlpBpyxk:b-706mww) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L287)):*
> ```XML
>         f(input);
>         consume(input);
>         return {};
> ```

why return {} at all?

---

*[src/contrib/data/dataset.ts, line 360 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHiPuu_dMzpKhS6ck:-L6JHiPuu_dMzpKhS6cl:b60m3nl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L360)):*
> ```XML
>  * Tensors are still present in the output.
>  */
> function consumePrep(input: DatasetElement): (output: DatasetElement) => void {
> ```

move this logic out to a util, this code looks quite similar to the code in batch_dataset

---

*[src/contrib/data/dataset.ts, line 384 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JN3ZWFbW_9nPNR5AD:-L6JN3ZWFbW_9nPNR5AE:bjbnjgt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L384)):*
> ```XML
> }
> 
> function retain(input: DatasetElement): DatasetElement {
> ```

can we call this something like keepDatasetElement so it's the same terminology as keep()?

---

*[src/contrib/data/dataset_test.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHwJQcGBQFYHuVbde:-L6JHwJQcGBQFYHuVbdf:b-7bchbt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L19)):*
> ```XML
>  */
> 
> import {keep} from '../../globals';
> ```

you can use keep via dl here (since tests use the public API).

---

*[src/contrib/data/dataset_test.ts, line 63 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOtU_4aQf-v5ya4XY:-L6JOtU_4aQf-v5ya4XZ:bkcuhj4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L63)):*
> ```XML
> }
> 
> describeWithFlags('Dataset', ALL_ENVS, () => {
> ```

use CPU_ENVs, no need to run this everywhere.

---

*[src/contrib/data/dataset_test.ts, line 213 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOmK5_TQfBYyL3H1j:-L6JOmK5_TQfBYyL3H1k:b-yyhiba) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L213)):*
> ```XML
>         return {};
>       })
>         .then(() => expect(count).toEqual(100))
> ```

use await in your tests for readability.

---

*[src/contrib/data/streams/data_stream.ts, line 150 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOQI7expcsdnOE7ir:-L6JOQI7expcsdnOE7is:bulbp3j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L150)):*
> ```XML
>    * @returns A `DataStream` of transformed elements.
>    */
>   map<OUT>(
> ```

is OUT a generic? If so, use "O".

---

*[src/contrib/data/streams/data_stream.ts, line 152 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMlmfkq98ErqRdrOC:-L6JMlmfkq98ErqRdrOD:b-xw1ksg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L152)):*
> ```XML
>   map<OUT>(
>       transform: (value: T) => OUT,
>       consumePrep?: (item: T) => ((output: OUT) => void),
> ```

add jsdocs, "consumePrep" on it's own doesn't tell me exactly what it does

---

*[src/contrib/data/streams/data_stream.ts, line 288 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMVdI3KkVGiWqSR3R:-L6JMVdI3KkVGiWqSR3S:b-jjzns0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L288)):*
> ```XML
>     while (this.count++ < this.maxCount) {
>       const skipped = await this.upstream.next();
>       if (this.consume != null) this.consume(skipped);
> ```

use {}

---

*[src/contrib/data/streams/data_stream.ts, line 402 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMgWTF1QWPWqu1_l4:-L6JMgWTF1QWPWqu1_l5:bqd1vzk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L402)):*
> ```XML
>       protected upstream: DataStream<T>,
>       protected predicate: (value: T) => boolean,
>       protected consume: (item: T) => void) {
> ```

this is optional, right?

---

*[src/contrib/data/streams/data_stream.ts, line 413 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMZgYC3YxE15yRDNa:-L6JMZgYC3YxE15yRDNb:b-jjzns0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L413)):*
> ```XML
>     if (this.predicate(item)) {
>       this.outputQueue.push(item);
>     } else if (this.consume != null)
> ```

use {}

---

*[src/contrib/data/streams/data_stream.ts, line 419 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JKTo5EHIP20EeocQ7:-L6JKTo5EHIP20EeocQ8:b-ock95z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L419)):*
> ```XML
> }
> 
> class MapStream<IN, OUT> extends QueueStream<OUT> {
> ```

can you make these I and O for consistency with generics, and everywhere else in this PR?

---

*[src/contrib/data/streams/data_stream.ts, line 438 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JK0Mj41sDmKGBKlGA:-L6JK0Mj41sDmKGBKlGB:b5qws61) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L438)):*
> ```XML
>     let mapped = this.transform(item);
> 
>     if (this.retain != null) mapped = this.retain(mapped);
> ```

use {} here and below (ts style)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6JLvvTi-G9yVvyNE-A:bzbro72)*

nsthorat:
Maybe that convo can be our sync tomorrow when Daniel gets back

---

Review status: 0 of 7 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6JPafH_EC5xvaC-AWI:b-2xouhz)*

davidsoergel:
You bet, happy to chat.  Thanks for the review!

---

Review status: 0 of 7 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/batch_dataset.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JFhjjRab2ID2gI5Gv:-L6JQmZNbnFGEsssND8F:bw4tlww) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset.ts#L19)):*

this will introduce a cycle, import keep directly from tracking.ts
</blockquote></details>

Oops quite so, thanks.  I fixed a pile of these already but apparently missed a couple.

---

*[src/contrib/data/batch_dataset_test.ts, line 18 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JGNBLnFESiFKxcEXh:-L6JRBfISvuuaGuTG-y3:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L18)):*

import from the file directly
</blockquote></details>

Done.

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6JZB3QwDeFwd-PdCgb:bx0gunu) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*

can we put this in some common util somewhere, we seem to use it in several places?

And can we maybe just call this disposeBatch or something to that effect?
</blockquote></details>

I did rename all these.  Re putting them in a common util class: they're all private.  Do you think the util consolidation is worth the cost of unnecessarily exporting them?  They are actually all a bit different, too-- sure a few lines here and there might be factored out, but not all that much.

---

*[src/contrib/data/dataset.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHcZcB1uUHlpBpyxj:-L6JRjfrOqOedNL8vC8Z:b-o4ga4u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L287)):*

why return {} at all?
</blockquote></details>

That's because the current implementation interprets null or undefined as "end of stream".  I'm already fixing this in the next PR by switching next() to return IteratorResult&lt;T> aka {value:T, done:boolean} throughout.

---

*[src/contrib/data/dataset.ts, line 360 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHiPuu_dMzpKhS6ck:-L6JZeGpUTARBcxYlKTx:b-cvmrhq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L360)):*

move this logic out to a util, this code looks quite similar to the code in batch_dataset
</blockquote></details>

See above: similar but not the same, and private.  Happy to move it out if you still prefer it that way.  Another approach would be to graduate DatasetElement and DatasetBatch to full classes, which could then have keep(), dispose(), and disposePrep() methods.

---

*[src/contrib/data/dataset.ts, line 384 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JN3ZWFbW_9nPNR5AD:-L6JSX9hygGzk9JwAFZ8:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L384)):*

can we call this something like keepDatasetElement so it's the same terminology as keep()?
</blockquote></details>

Done.

---

*[src/contrib/data/dataset_test.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHwJQcGBQFYHuVbde:-L6JSgOehHAxn_TGrgWH:b-9so0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L19)):*

you can use keep via dl here (since tests use the public API).
</blockquote></details>

Did that one for now.

There are tons of other direct imports though, in this code and in DLJS core tests.  I get that non-test code must use direct imports, but is there a reason to prefer one or the other style in tests?  Is consistency a good enough reason to pick one?

---

*[src/contrib/data/dataset_test.ts, line 63 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOtU_4aQf-v5ya4XY:-L6JTRH2uLsro0Zbbm5X:b-hu5hug) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L63)):*

use CPU_ENVs, no need to run this everywhere.
</blockquote></details>

Did this one.  I left batch_dataset_test and statistics_test alone for now, on the theory that they do Tensor manipulations that should be tested on GPU.

---

*[src/contrib/data/dataset_test.ts, line 213 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOmK5_TQfBYyL3H1j:-L6JUNu5lVB3P4wgkO1x:b7wd9dx) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L213)):*

use await in your tests for readability.
</blockquote></details>

Oh wow, for some reason I thought it() wouldn't accept an async function, but apparently it does.  Excellent-- these tests are fixed and I'll update the rest in a followup PR.

---

*[src/contrib/data/streams/data_stream.ts, line 150 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOQI7expcsdnOE7ir:-L6JXqw96rrwKPpU0WOF:b-m55pch) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L150)):*

is OUT a generic? If so, use "O".
</blockquote></details>

Done, though I'm curious why.  OUT is a perfectly valid generic type variable, and seems much clearer to me here.

---

*[src/contrib/data/streams/data_stream.ts, line 152 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMlmfkq98ErqRdrOC:-L6JoVFCM9S3vRHZp6C0:b-9pe9cl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L152)):*

add jsdocs, "consumePrep" on it's own doesn't tell me exactly what it does
</blockquote></details>

Oh jeez, I meant to do that and forgot, thanks.  Any interest in adding a tslint "completed-docs" rule to catch these in the future?  Low priority of course.

---

*[src/contrib/data/streams/data_stream.ts, line 288 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMVdI3KkVGiWqSR3R:-L6JZzTHdRevq7iDlbj_:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L288)):*

use {}
</blockquote></details>

Done.

---

*[src/contrib/data/streams/data_stream.ts, line 402 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMgWTF1QWPWqu1_l4:-L6JYG62_kar7UTxAmB3:b1f4ww8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L402)):*

this is optional, right?
</blockquote></details>

Yep good catch!

---

*[src/contrib/data/streams/data_stream.ts, line 413 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JMZgYC3YxE15yRDNa:-L6JZ3uorgQUsFfNQI3K:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L413)):*

use {}
</blockquote></details>

Done.

---

*[src/contrib/data/streams/data_stream.ts, line 419 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JKTo5EHIP20EeocQ7:-L6JYanqo45qGDNA5p9f:bb6un1j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L419)):*

can you make these I and O for consistency with generics, and everywhere else in this PR?
</blockquote></details>

Done

---

*[src/contrib/data/streams/data_stream.ts, line 438 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JK0Mj41sDmKGBKlGA:-L6JZ37zvOXzY4vcYP52:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L438)):*

use {} here and below (ts style)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6Jr43Pd9efj4eyMKR1:b765gxo)*

davidsoergel:
Review status: 0 of 7 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/dataset.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHcZcB1uUHlpBpyxj:-L6JtQ-nPTIk3hXcEKfd:b87l0he) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L287)):*

That's because the current implementation interprets null or undefined as "end of stream".  I'm already fixing this in the next PR by switching next() to return IteratorResult&lt;T> aka {value:T, done:boolean} throughout.
</blockquote></details>

Oh, and I realized that the forEach implementation consumes its inputs by default anyway, given a consumePrep function, so this whole bit was unnecessary anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785)*

nsthorat:
Few more nits ;) 

---

Review status: 0 of 7 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6M_N9rWY15xVq6MFee:bh7cdql) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*

I did rename all these.  Re putting them in a common util class: they're all private.  Do you think the util consolidation is worth the cost of unnecessarily exporting them?  They are actually all a bit different, too-- sure a few lines here and there might be factored out, but not all that much.
</blockquote></details>

I think it's still worth it to try to consolidate. File-level exports will never get exported at the module level unless explicitly told to do so, so I wouldn't worry about extra file-level exports.

Ways to consolidate:
- for disposeBatchPrep, it doesn't matter that the function has an argument, you can simply use `() =>` with no args since you don't use the argument (typescript will be happy). Then you can always taken an array of DatasetElements or a single DatasetElement and return that function.
- Same for extractTensorsFromElementArray, it can take a single one or an array.

---

*[src/contrib/data/dataset.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHcZcB1uUHlpBpyxj:-L6M_cvShf2iDFEQmCXt:b-xwaf1h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L287)):*

Oh, and I realized that the forEach implementation consumes its inputs by default anyway, given a consumePrep function, so this whole bit was unnecessary anyway.
</blockquote></details>

Awesome. One more small thing, can you s/undefined/null

---

*[src/contrib/data/dataset_test.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHwJQcGBQFYHuVbde:-L6Ma7sgQ57rIN15pt6g:bqrvt3p) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L19)):*

Did that one for now.

There are tons of other direct imports though, in this code and in DLJS core tests.  I get that non-test code must use direct imports, but is there a reason to prefer one or the other style in tests?  Is consistency a good enough reason to pick one?
</blockquote></details>

Yeah, we've been slowly migrating to this convention:

- Tests import public API as dl through './index' unless they are testing something not module-exported
- Internal library code does not import through index (because of cycles), it depends on the files that export our factory functions

---

*[src/contrib/data/dataset_test.ts, line 147 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Mc6PMANDGl5raek9-:-L6Mc6PMANDGl5raek90:b6lu39q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/dataset_test.ts#L147)):*
> ```XML
>     const items = await ds.collectAll();
>     expect(items.length).toEqual(100);
>     expect(dl.memory().numTensors).toEqual(200);
> ```

why 200? where does the 2 come from?

---

*[src/contrib/data/dataset_test.ts, line 154 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Mc8B-6xd7xiNuHtpx:-L6Mc8B-6xd7xiNuHtpy:bd0iqt8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/dataset_test.ts#L154)):*
> ```XML
>     expect(dl.memory().numTensors).toEqual(0);
>     await ds.skip(15).collectAll();
>     expect(dl.memory().numTensors).toEqual(170);
> ```

Why 170? Can you make it clear here either with a variable or a comment? If I just read this without understanding skip, I would assume this is testing that we have a leak :)

Other places in this file too that you assert a non-zero constant

---

*[src/contrib/data/streams/data_stream.ts, line 150 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOQI7expcsdnOE7ir:-L6Md4vEUBOZYbGGirTK:b-bdql15) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L150)):*

Done, though I'm curious why.  OUT is a perfectly valid generic type variable, and seems much clearer to me here.
</blockquote></details>

Convention :) 

---

*[src/contrib/data/streams/data_stream.ts, line 149 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McLtVU2tjOJdTHdPu:-L6McLtVU2tjOJdTHdPv:b-c5mel3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L149)):*
> ```XML
>    * @param predicate A function mapping a stream element to a transformed
>    *   element.
>    * @param consumePrep A function that tentatively schedules an item or its
> ```

instead of `consumePrep` maybe call this `cleanupFn` or `disposeFn` - it took me a little thinking to realize this is simply the callback that gets called at the end to dispose Tensors. Good to keep the same terminology everywhere (I don't see a use case where you'd use consumePrep for non-disposal).

Keeping the "Fn" suffix where you have these functions would be nice, and in general s/retain/keep and s/consume/dispose

---

*[src/contrib/data/streams/data_stream.ts, line 154 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Md71JK-mYiWqWPB8a:-L6Md71KRAAadBJr12Kt:bpo7lrl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L154)):*
> ```XML
>    *   function accepts an argument that may modify the list of objects to be
>    *   collected, e.g. by reversing the tentative decision made previously.
>    * @param retain A function that marks an item to be retained (i.e.,
> ```

Also probably call this `keepFn` for consistency.

---

*[src/contrib/data/streams/data_stream.ts, line 170 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McrkSoIuJeWMYmEkt:-L6McrkSoIuJeWMYmEku:b6vvhho) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L170)):*
> ```XML
>    *
>    * @param f A function to apply to each stream element.
>    * @param consumePrep A function that tentatively schedules an item or its
> ```

probably rename this as per above, right now the jsdoc doesn't match the param name

---

*[src/contrib/data/streams/data_stream.ts, line 179 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6MciF9Lk3neLU3OqfM:-L6MciF9Lk3neLU3OqfN:b-54dmy5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L179)):*
> ```XML
>       Promise<void> {
>     const consumePrep =
>         consume == null ? null : (item: T) => ((output: void) => consume(item));
> ```

this looks funny, should this just be

```
const consumePrep =
      consume == null ? null : (item: T) => () => consume(item);
```

---

*[src/contrib/data/streams/data_stream.ts, line 225 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McyM-3hHt2eqssQYa:-L6McyM-3hHt2eqssQYb:b-6w7r7u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L225)):*
> ```XML
>    * @param count The number of items to skip.  If a negative or undefined value
>    *   is given, the entire stream is returned unaltered.
>    * @param consume A function that finalizes an item, e.g. by marking it for
> ```

probably rename this to disposeFn as well

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6MdTCf_mPY8IoQB_Ik:b6vxdmy)*

davidsoergel:
Per offline discussion, this solution is much simpler, but trusts that a) the outer scope won't  unexpectedly dispose Tensors, and b) Tensors passed into pipeline stages may be safely disposed, i.e., they have no other usages.

---

Review status: 0 of 10 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6SgKPX7S5cQKKCIaFq:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*

I think it's still worth it to try to consolidate. File-level exports will never get exported at the module level unless explicitly told to do so, so I wouldn't worry about extra file-level exports.

Ways to consolidate:
- for disposeBatchPrep, it doesn't matter that the function has an argument, you can simply use `() =>` with no args since you don't use the argument (typescript will be happy). Then you can always taken an array of DatasetElements or a single DatasetElement and return that function.
- Same for extractTensorsFromElementArray, it can take a single one or an array.
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/dataset.ts, line 287 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHcZcB1uUHlpBpyxj:-L6SgSf7BSm4ZpqKv2cz:b-dux2qf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L287)):*

Awesome. One more small thing, can you s/undefined/null
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/dataset.ts, line 360 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHiPuu_dMzpKhS6ck:-L6SgVak0LXmoWjGIPkX:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset.ts#L360)):*

See above: similar but not the same, and private.  Happy to move it out if you still prefer it that way.  Another approach would be to graduate DatasetElement and DatasetBatch to full classes, which could then have keep(), dispose(), and disposePrep() methods.
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/dataset_test.ts, line 19 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JHwJQcGBQFYHuVbde:-L6Sgcrar4lQULVbW7OX:b-vmrj4e) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/dataset_test.ts#L19)):*

Yeah, we've been slowly migrating to this convention:

- Tests import public API as dl through './index' unless they are testing something not module-exported
- Internal library code does not import through index (because of cycles), it depends on the files that export our factory functions
</blockquote></details>

SGTM, thanks!

---

*[src/contrib/data/dataset_test.ts, line 147 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Mc6PMANDGl5raek9-:-L6SmyIDCiznKF6JoJtz:b-sbv3q1) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/dataset_test.ts#L147)):*

why 200? where does the 2 come from?
</blockquote></details>

Added comments to all these tests.

---

*[src/contrib/data/dataset_test.ts, line 154 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Mc8B-6xd7xiNuHtpx:-L6Sn0xQ0r7rvWcR8GaF:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/dataset_test.ts#L154)):*

Why 170? Can you make it clear here either with a variable or a comment? If I just read this without understanding skip, I would assume this is testing that we have a leak :)

Other places in this file too that you assert a non-zero constant
</blockquote></details>

Done.

---

*[src/contrib/data/streams/data_stream.ts, line 150 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JOQI7expcsdnOE7ir:-L6SgnmXw7cyqpb5vfa0:b-83tbea) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/streams/data_stream.ts#L150)):*

Convention :) 
</blockquote></details>

OK :)

---

*[src/contrib/data/streams/data_stream.ts, line 149 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McLtVU2tjOJdTHdPu:-L6Sgk1P7ijRXkaI1gZI:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L149)):*

instead of `consumePrep` maybe call this `cleanupFn` or `disposeFn` - it took me a little thinking to realize this is simply the callback that gets called at the end to dispose Tensors. Good to keep the same terminology everywhere (I don't see a use case where you'd use consumePrep for non-disposal).

Keeping the "Fn" suffix where you have these functions would be nice, and in general s/retain/keep and s/consume/dispose
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/streams/data_stream.ts, line 154 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Md71JK-mYiWqWPB8a:-L6Sgl45au7IhfazvozV:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L154)):*

Also probably call this `keepFn` for consistency.
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/streams/data_stream.ts, line 170 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McrkSoIuJeWMYmEkt:-L6SgrYolXrXxklWEZHR:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L170)):*

probably rename this as per above, right now the jsdoc doesn't match the param name
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/streams/data_stream.ts, line 179 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6MciF9Lk3neLU3OqfM:-L6Sgt8nB38QFY4SN2xA:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L179)):*

this looks funny, should this just be

```
const consumePrep =
      consume == null ? null : (item: T) => () => consume(item);
```
</blockquote></details>

obsoleted by new design.

---

*[src/contrib/data/streams/data_stream.ts, line 225 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6McyM-3hHt2eqssQYa:-L6SguZNEecc_mG_TFNx:b-g3e0v) ([raw file](https://github.com/pair-code/deeplearnjs/blob/4e951716195736bd6c27c0469080bf1d82b37d4a/src/contrib/data/streams/data_stream.ts#L225)):*

probably rename this to disposeFn as well
</blockquote></details>

obsoleted by new design.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6SmQ8tg5qCnKHRxHHu:b-jvbmcb)*

dsmilkov:
Reviewed 2 of 7 files at r1, 8 of 9 files at r3.
Review status: 10 of 11 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 460 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Sr0xXDr3QQsfUbsQx:-L6Sr0xXDr3QQsfUbsQy:bbdvyow) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/engine.ts#L460)):*
> ```XML
> // This is a copy of extractTensorsFromScopeResult but with no type constraints.
> // tslint:disable-next-line:no-any
> export function extractTensorsFromAny(result: any): Tensor[] {
> ```

Let's consolidate: move `extractTensorsFromScopeResult` to `util.ts` and rename to `extractTensorsFromContainer` and use it in both places. Rename also `ScopeResult` to `TensorContainer`, and similarly for the related types.

If you run into type constraints issues, feel free to relax the type, though you may not need to since the current type (`ScopeResult`) is already general and can describe any primitive/array/JSON object. The only thing that type prohibits is a `Promise<any>`, rightfully so.

---

*[src/tracking.ts, line 103 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsWQxylo8HQ1RIKe9:-L6Srxd-kyPWhY5jzpvz:b3cho12) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/tracking.ts#L103)):*
> ```XML
>   // DO NOT SUBMIT; stopgap.
>   // tslint:disable-next-line:no-any
>   static dispose(x: any) {
> ```

type it as x: TensorContainer (see comment above)

---

*[src/tracking.ts, line 105 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsfS1lcTDYt1dqeYg:-L6SsfS1lcTDYt1dqeYh:b-e9qfv4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/tracking.ts#L105)):*
> ```XML
>   static dispose(x: any) {
>     const tensors = extractTensorsFromAny(x);
>     for (const t of tensors) {
> ```

this is a bit unconventional since tensors is an Array, not an Object, and not supported in IE (pre Edge). Use:
```js
tensors.forEach(t => t.dispose());
```
for consistency with the rest of the codebase

---

*[src/contrib/index.ts, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SubHi56KjikIILw_T:-L6SubHi56KjikIILw_U:b1g684) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/index.ts#L20)):*
> ```XML
> // TODO(soergel) carefully consider what actually should be exposed here
> export {Dataset} from './data/dataset';
> export {datasetFromConcatenated, datasetFromElements} from './data/dataset';
> ```

let's not export these yet, since these probably will not be exposed to the user later, and we want to avoid modifying tests later. In your tests, import these 2 methods directly from their file, and import from 'index' only the things that are already exported in index.ts

---

*[src/contrib/data/batch_dataset.ts, line 93 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsyIql90fc6CSnPNL:-L6SsyIql90fc6CSnPNM:b-b200ci) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset.ts#L93)):*
> ```XML
>     }
>   }
>   for (const e of elements) {
> ```

if something is an array, use .forEach(e => e.dispose()) for consistency.

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6SwOftz0lBO4l-1hWU:bnn8ovz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*

obsoleted by new design.
</blockquote></details>

Since this is obsolete now that we have `dl.dispose()`, remove it?

---

*[src/contrib/data/batch_dataset_test.ts, line 22 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SvExXF1XFKkOrWwrB:-L6SvExXF1XFKkOrWwrC:b-3vyqi7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L22)):*
> ```XML
> import {TestDataset} from './dataset_test';
> 
> describeWithFlags('Dataset.batch()', CPU_ENVS, () => {
> ```

Use `ALL_ENVS` since we removed the byte packing settings  from `ALL_ENVS` and things are numerically stable. We will bring the byte-packing setting when we make it numerically stable.

---

*[src/contrib/data/batch_dataset_test.ts, line 30 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Sw6oSEb1Bds8TG6aD:-L6Sw6oSEb1Bds8TG6aE:bh0g2ad) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L30)):*
> ```XML
>         .then(batchStream => batchStream.collectRemaining().then(result => {
>           expect(result.length).toEqual(13);
>           for (const batch of result.slice(0, 12)) {
> ```

let's do forEach() for consistency throughout the codebase.

---

*[src/contrib/data/batch_dataset_test.ts, line 38 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Svb8ktAOnPpk7Xb1G:-L6Svb8ktAOnPpk7Xb1H:b2w3jl9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L38)):*
> ```XML
>           return result;
>         }))
>         .then((result) => {
> ```

any reason why we avoid async/await. Is it because jasmine/karma can't catch any error that way?

If that's the case, would this work:
```js
try {
  const a = await someAsyncF();
  ... 
} catch (e) {
  done.fail(e);
}
```

---

*[src/contrib/data/dataset.ts, line 54 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Swki-Gh0_4QfGSg-a:-L6Swki-Gh0_4QfGSg-b:b-8712y0) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/dataset.ts#L54)):*
> ```XML
>    * The dl.tidy() approach cannot be used in a asynchronous context.
>    */
>   abstract async getStream(): Promise<DataStream<DatasetElement>>;
> ```

Not for this PR, but wanted to start the discussion: Can we make getStream sync, or is there a fundamental reason? I want to push for only getNext() be async, and everything else be sync, to the extent possible.

---

*[src/contrib/data/dataset_test.ts, line 45 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SzvyWx4q6KQk5cDF1:-L6SzvyXQCHpoj5wzJOp:bm05nyb) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/dataset_test.ts#L45)):*
> ```XML
>       'string': `Item ${elementNumber}`
>     };
>     dl.keep(result['Tensor']);
> ```

just double-checking, with the new design, is `keep` still needed?

---

*[src/contrib/data/streams/data_stream.ts, line 356 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6T-EPM3FAIp7bLDEmB:-L6T-EPM3FAIp7bLDEmC:bws35bi) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/streams/data_stream.ts#L356)):*
> ```XML
> 
> // TODO(soergel): consider clean separation of synchronous pumpOne
> /*abstract class TransformingQueueStream<I, O> extends QueueStream<O> {
> ```

just checking: if commented out code meant be kept here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785)*

davidsoergel:
Review status: 10 of 11 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[src/tracking.ts, line 103 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsWQxylo8HQ1RIKe9:-L6TJh2xegIk-46CVsWh:b5uefj3) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/tracking.ts#L103)):*

I tried that; the problem is that currently streams can be of any type, e.g. DataStream<T>, and we want to be able to call dispose on any stream element.  I briefly went down the road of DataStream<T extends ScopeResult>, but stopped because people ought to be able to stream their own custom types.  And, we can't protect the dispose call with "if(x instanceof ScopeResult)" because ScopeResult is just a compile-time type, not a class.  So that's why I went with any here. WDYT?
</blockquote></details>

Oops let me try that again: I meant "e.g. DataStream&lt;T>" and then "down the road of DataStream&lt;T extends ScopeResult>".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785)*

davidsoergel:
Review status: 6 of 11 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

*[src/engine.ts, line 460 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Sr0xXDr3QQsfUbsQx:-L6TZA4LGcEuLdbEXNYT:b-c1u4rz) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/engine.ts#L460)):*

I think I need the `any` version (see below), but how about having `extractTensorsFromContainer` just call  `extractTensorsFromAny` (but with added type safety)?  That way we can prefer the Container version in general, and use the `any` version only when we specifically need it.
</blockquote></details>

(Preparing this part in another PR to be submitted before this one)

---

*[src/tracking.ts, line 105 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsfS1lcTDYt1dqeYg:-L6TK6dkvaE4EE35Ttle:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/tracking.ts#L105)):*

this is a bit unconventional since tensors is an Array, not an Object, and not supported in IE (pre Edge). Use:
```js
tensors.forEach(t => t.dispose());
```
for consistency with the rest of the codebase
</blockquote></details>

Done.

---

*[src/contrib/index.ts, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SubHi56KjikIILw_T:-L6TK_FOVxNAqjkAXwXK:b-a5xdnp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/index.ts#L20)):*

let's not export these yet, since these probably will not be exposed to the user later, and we want to avoid modifying tests later. In your tests, import these 2 methods directly from their file, and import from 'index' only the things that are already exported in index.ts
</blockquote></details>

I think users are very likely to want these, at least for writing their own tests.  Especially concatenating datasets seems like something the api should clearly provide.

---

*[src/contrib/data/batch_dataset.ts, line 93 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SsyIql90fc6CSnPNL:-L6TTX9XBf4JMtqABoJj:b-wkciwr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset.ts#L93)):*

if something is an array, use .forEach(e => e.dispose()) for consistency.
</blockquote></details>

Done here; I'll keep my eye out for other instances.

---

*[src/contrib/data/batch_dataset_test.ts, line 95 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6JH3Tl0yUMqCcRKLfm:-L6TX85Blw3a_edb3D6J:bb6un1j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/f80bb7a2664eb35ad0fe052d91707996a89fbdb8/src/contrib/data/batch_dataset_test.ts#L95)):*

Since this is obsolete now that we have `dl.dispose()`, remove it?
</blockquote></details>

Done

---

*[src/contrib/data/batch_dataset_test.ts, line 22 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SvExXF1XFKkOrWwrB:-L6TTdGolsJQxmCN5lqH:b3yi85f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L22)):*

Use `ALL_ENVS` since we removed the byte packing settings  from `ALL_ENVS` and things are numerically stable. We will bring the byte-packing setting when we make it numerically stable.
</blockquote></details>

I had changed it from ALL_ENVS to CPU_ENVS at @nsthorat's request, which I think was about saving time when the GPU test is unlikely to reveal any issue anyway (i.e., there's no math in here).  I don't have an opinion either way.  Just tell me what you guys agree ;)

---

*[src/contrib/data/batch_dataset_test.ts, line 30 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Sw6oSEb1Bds8TG6aD:-L6TWFThl3EgsnWnuj8B:b-xcs8gq) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L30)):*

let's do forEach() for consistency throughout the codebase.
</blockquote></details>

+1, thanks, I prefer that too and just forgot

---

*[src/contrib/data/dataset.ts, line 54 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6Swki-Gh0_4QfGSg-a:-L6TXTTXUEZCEOYV9GWD:b1yuaqe) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/dataset.ts#L54)):*

Not for this PR, but wanted to start the discussion: Can we make getStream sync, or is there a fundamental reason? I want to push for only getNext() be async, and everything else be sync, to the extent possible.
</blockquote></details>

Discussion underway in #798.

---

*[src/contrib/data/dataset_test.ts, line 45 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SzvyWx4q6KQk5cDF1:-L6TXtd5bBfKvPwnkUrU:b-wjcx1f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/dataset_test.ts#L45)):*

just double-checking, with the new design, is `keep` still needed?
</blockquote></details>

Nope, good catch, thanks!

---

*[src/contrib/data/streams/data_stream.ts, line 356 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6T-EPM3FAIp7bLDEmB:-L6TYv-5t6jW2U0Sohy-:b-ewqx0q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/streams/data_stream.ts#L356)):*

just checking: if commented out code meant be kept here?
</blockquote></details>

Oops no :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785)*

dsmilkov:
:lgtm_strong:

---

Review status: 2 of 8 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/contrib/index.ts, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SubHi56KjikIILw_T:-L6bztsqz6NszDxjwrqx:bxbjgrr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/index.ts#L20)):*

I think users are very likely to want these, at least for writing their own tests.  Especially concatenating datasets seems like something the api should clearly provide.
</blockquote></details>

Can we not deviate from https://www.tensorflow.org/api_docs/python/tf/contrib/data/Dataset#concatenate? I want to make sure there are no redundant methods in the user facing API. LGTM to keep the ball rolling, but let's resolve this comment in private.

---

*[src/contrib/data/batch_dataset_test.ts, line 22 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SvExXF1XFKkOrWwrB:-L6c-hKLFsuVgTcE79s6:b7jjhdl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/data/batch_dataset_test.ts#L22)):*

I had changed it from ALL_ENVS to CPU_ENVS at @nsthorat's request, which I think was about saving time when the GPU test is unlikely to reveal any issue anyway (i.e., there's no math in here).  I don't have an opinion either way.  Just tell me what you guys agree ;)
</blockquote></details>

Got it. CPU is fine then, This assumes our GPU works, and that's a fair assumption. Core can worry about tht.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-:-L6c1B0fUPCedgfa4Pxz:b-jbsf8z)*

davidsoergel:
Review status: 1 of 8 files reviewed at latest revision, 16 unresolved discussions, some commit checks pending.

---

*[src/contrib/index.ts, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785#-L6SubHi56KjikIILw_T:-L6cJno6bOjgrc0drL0A:bb6un1j) ([raw file](https://github.com/pair-code/deeplearnjs/blob/25f17ecc7bf252996f421d2e115bdb48d9d809a5/src/contrib/index.ts#L20)):*

Can we not deviate from https://www.tensorflow.org/api_docs/python/tf/contrib/data/Dataset#concatenate? I want to make sure there are no redundant methods in the user facing API. LGTM to keep the ball rolling, but let's resolve this comment in private.
</blockquote></details>

Done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/785)*

PR #791, opened 2018-02-27
Add a 'custom' backend to allow more than 'cpu' and 'gpu' backends.

nkreeger:
First idea at enabling the environment to allow a custom backend. I don't think I need to get into the weeds about setting the best-backend just yet. Happy to take another approach - but type def'ing constants is hard to work with outside of the package.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/791)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -186,7 +186,7 @@ function isWebGLGetBufferSubDataAsyncExtensionEnabled(webGLVersion: number) {
 }
 
 /** @docalias 'webgl'|'cpu' */
-export type BackendType = 'webgl'|'cpu';
+export type BackendType = 'webgl'|'cpu'|string;
 
 /** List of currently supported backends ordered by preference. */
 const SUPPORTED_BACKENDS: BackendType[] = ['webgl', 'cpu'];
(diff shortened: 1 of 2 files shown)


nsthorat:
I think this approach is good, let's wait for @dsmilkov though since he wrote the setBackend code.

dsmilkov:
:lgtm_strong:

---

Reviewed 2 of 2 files at r1.

---

*[src/environment_test.ts, line 266 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/791#-L6N4spaB6LUrSQiN3qG:-L6N4spaB6LUrSQiN3qH:b-kbhtgk) ([raw file](https://github.com/pair-code/deeplearnjs/blob/e694fe839fdb84870175396d9a1db500e61ef309/src/environment_test.ts#L266)):*
> ```XML
>   it('allow custom backend', () => {
>     const backend = new MathBackendCPU();
>     const success = ENV.addCustomBackend('custom', () => backend);
> ```

Hey Nick , this is good, but I think you should use ENV.registerBackend in your library instead of addCustomBackend.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/791#-:-L6N55V4OuVofIkWEdyn:b-jbsf8z)*

PR #793, opened 2018-02-28
Add gradients for tile op

caisq:
Currently, 1, 2, 3 and 4D cases are covered.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/793)


nsthorat:
:lgtm_strong:

This is awesome, thanks for doing this!

I was thinking you could reduce the size of the memory footprint by disposing the intermediates directly, but then realized you can't dispose() directly in a gradient right now... that's going to be a tricky one to fix.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/793#-:-L6PiDg19SDuAtqUGH-l:blg875c)*

caisq:
Added TODO item for memory optimization.

PR #794, opened 2018-02-28
2d atrous convolution and atrous depthwise convolution

oveddan:
This implements 2d atrous convolution and 2d atrous depthwise convolution, as described in [Rethinking Atrous Convolution for Semantic Image Segmentation](https://arxiv.org/abs/1706.05587). Refer to the [documentation for tf.nn.atrous_conv_2d](https://www.tensorflow.org/api_docs/python/tf/nn/atrous_conv2d) for an explanation of of how atrous convolution works.

Atrous convolution is currently implemented in TensorFlow in the following methods (the parameter name is listed next to each method):

* [tf.nn.conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d) - `dilations`
* [tf.nn.conv2d_backprop_filter](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_backprop_filter) - `dilations`
* [tf.nn.conv2d_backprop_input](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_backprop_input) - `dilations`
* [tf.nn.conv3d](https://www.tensorflow.org/api_docs/python/tf/nn/conv3d) - `dilations`
* [tf.nn.conv3d_backprop_filter_v2](https://www.tensorflow.org/api_docs/python/tf/nn/conv3d_backprop_filter_v2) - `dilations`
* [tf.nn.convolution](https://www.tensorflow.org/api_docs/python/tf/nn/convolution) - `dilation_rate`
* [tf.nn.atrous_conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/atrous_conv2d) -`rate`
* [tf.nn.atrous_conv2d_transpose](https://www.tensorflow.org/api_docs/python/tf/nn/atrous_conv2d_transpose) - `rate`
* [tf.nn.depthwise_conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d) - `rate`
* [tf.nn.depthwise_conv2d_native](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d_native) - `dilations`
* [tf.nn.depthwise_conv2d_native_backprop_filter](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d_native_backprop_filter) - `dilations`
* [tf.nn.depthwise_conv2d_native_backprop_input](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d_native_backprop_input) - `dilations`
* [tf.nn.dilation2d](https://www.tensorflow.org/api_docs/python/tf/nn/dilation2d) - `rates`
* [tf.nn.quantized_conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/quantized_conv2d) - `dilations`

This PR modifies the webgl and cpu ops `depthwiseConv2d` and `conv2d` to take the parameter `dilationRate` and perform atrous convolution if the rate width or height is greater than 1.  In its current state, it is potentially a breaking change in the api for `conv1d` and `conv2d` as it adds the positional argument `dilations`.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/794)

### DIFF
--- a/src/kernels/backend_cpu.ts
+++ b/src/kernels/backend_cpu.ts
@@ -863,6 +863,8 @@ export class MathBackendCPU implements KernelBackend {
   conv2d(x: Tensor4D, filter: Tensor4D, convInfo: Conv2DInfo): Tensor4D {
     const filterHeight = convInfo.filterHeight;
     const filterWidth = convInfo.filterWidth;
+    const dilationHeight = convInfo.dilationHeight;
+    const dilationWidth = convInfo.dilationWidth;
     const padLeft = convInfo.padInfo.left;
     const padTop = convInfo.padInfo.top;
     const y = ops.buffer<Rank.R4>(convInfo.outShape, x.dtype);
@@ -871,17 +873,24 @@ export class MathBackendCPU implements KernelBackend {
       for (let d2 = 0; d2 < convInfo.outChannels; ++d2) {
         for (let yR = 0; yR < convInfo.outHeight; ++yR) {
           const xRCorner = yR * convInfo.strideHeight - padLeft;
-          const xRMin = Math.max(0, xRCorner);
-          const xRMax = Math.min(convInfo.inHeight, filterHeight + xRCorner);
           for (let yC = 0; yC < convInfo.outWidth; ++yC) {
             const xCCorner = yC * convInfo.strideWidth - padTop;
-            const xCMin = Math.max(0, xCCorner);
-            const xCMax = Math.min(convInfo.inWidth, filterWidth + xCCorner);
+
             let dotProd = 0;
-            for (let xR = xRMin; xR < xRMax; ++xR) {
-              const wR = xR - xRCorner;
-              for (let xC = xCMin; xC < xCMax; ++xC) {
-                const wC = xC - xCCorner;
+            for (let wR = 0; wR < filterHeight; wR++) {
+              const xR = xRCorner + wR * dilationHeight;
+
+              if (xR < 0 || xR >= convInfo.inHeight) {
+                continue;
+              }
+
+              for (let wC = 0; wC < filterWidth; wC++) {
+                const xC = xCCorner + wC * dilationWidth;
+
+                if (xC < 0 || xC >= convInfo.inWidth) {
+                  continue;
+                }
+
                 for (let d1 = 0; d1 < convInfo.inChannels; ++d1) {
                   const pixel = x.get(b, xR, xC, d1);
                   const weight = filter.get(wR, wC, d1, d2);
@@ -989,6 +998,8 @@ export class MathBackendCPU implements KernelBackend {
       Tensor4D {
     const filterHeight = convInfo.filterHeight;
     const filterWidth = convInfo.filterWidth;
+    const dilationHeight = convInfo.dilationHeight;
+    const dilationWidth = convInfo.dilationWidth;
     const padLeft = convInfo.padInfo.left;
     const padTop = convInfo.padInfo.top;
     const chMul = convInfo.outChannels / convInfo.inChannels;
@@ -998,18 +1009,24 @@ export class MathBackendCPU implements KernelBackend {
       for (let d1 = 0; d1 < convInfo.inChannels; ++d1) {
         for (let yR = 0; yR < convInfo.outHeight; ++yR) {
           const xRCorner = yR * convInfo.strideHeight - padLeft;
-          const xRMin = Math.max(0, xRCorner);
-          const xRMax = Math.min(convInfo.inHeight, filterHeight + xRCorner);
           for (let yC = 0; yC < convInfo.outWidth; ++yC) {
             const xCCorner = yC * convInfo.strideWidth - padTop;
-            const xCMin = Math.max(0, xCCorner);
-            const xCMax = Math.min(convInfo.inWidth, filterWidth + xCCorner);
             for (let q = 0; q < chMul; ++q) {
               let dotProd = 0;
-              for (let xR = xRMin; xR < xRMax; ++xR) {
-                const wR = xR - xRCorner;
-                for (let xC = xCMin; xC < xCMax; ++xC) {
-                  const wC = xC - xCCorner;
+              for (let wR = 0; wR < filterHeight; ++wR) {
+                const xR = xRCorner + wR * dilationHeight;
+
+                if (xR < 0 || xR >= convInfo.inHeight) {
+                  continue;
+                }
+
+                for (let wC = 0; wC < filterWidth; ++wC) {
+                  const xC = xCCorner + wC * dilationWidth;
+
+                  if (xC < 0 || xC >= convInfo.inWidth) {
+                    continue;
+                  }
+
                   const pixel = x.get(b, xR, xC, d1);
                   const weight = filter.get(wR, wC, d1, q);
                   dotProd += pixel * weight;
@@ -1021,6 +1038,7 @@ export class MathBackendCPU implements KernelBackend {
         }
       }
     }
+
     return y.toTensor();
   }
 
--- a/src/kernels/webgl/conv_gpu.ts
+++ b/src/kernels/webgl/conv_gpu.ts
@@ -29,6 +29,8 @@ export class Conv2DProgram implements GPGPUProgram {
     const padLeft = convInfo.padInfo.left;
     const strideHeight = convInfo.strideHeight;
     const strideWidth = convInfo.strideWidth;
+    const dilationHeight = convInfo.dilationHeight;
+    const dilationWidth = convInfo.dilationWidth;
     const filterHeight = convInfo.filterHeight;
     const filterWidth = convInfo.filterWidth;
 
@@ -52,14 +54,14 @@ export class Conv2DProgram implements GPGPUProgram {
         // ? = to be determined. : = across all values in that axis.
         float dotProd = 0.0;
         for (int wR = 0; wR < ${filterHeight}; wR++) {
-          int xR = xRCorner + wR;
+          int xR = xRCorner + wR * ${dilationHeight};
 
           if (xR < 0 || xR >= ${convInfo.inHeight}) {
             continue;
           }
 
           for (int wC = 0; wC < ${filterWidth}; wC++) {
-            int xC = xCCorner + wC;
+            int xC = xCCorner + wC * ${dilationWidth};
 
             if (xC < 0 || xC >= ${convInfo.inWidth}) {
               continue;
--- a/src/kernels/webgl/conv_gpu_depthwise.ts
+++ b/src/kernels/webgl/conv_gpu_depthwise.ts
@@ -32,6 +32,8 @@ export class DepthwiseConv2DProgram implements GPGPUProgram {
     const padLeft = convInfo.padInfo.left;
     const strideHeight = convInfo.strideHeight;
     const strideWidth = convInfo.strideWidth;
+    const dilationHeight = convInfo.dilationHeight;
+    const dilationWidth = convInfo.dilationWidth;
     const filterHeight = convInfo.filterHeight;
     const filterWidth = convInfo.filterWidth;
     const channelMul = convInfo.outChannels / convInfo.inChannels;
@@ -56,14 +58,14 @@ export class DepthwiseConv2DProgram implements GPGPUProgram {
         float dotProd = 0.0;
         // TODO(dsmilkov): Flatten the two for loops and vec4 the operations.
         for (int wR = 0; wR < ${filterHeight}; wR++) {
-          int xR = xRCorner + wR;
+          int xR = xRCorner + wR * ${dilationHeight};
 
           if (xR < 0 || xR >= ${xNumRows}) {
             continue;
           }
 
           for (int wC = 0; wC < ${filterWidth}; wC++) {
-            int xC = xCCorner + wC;
+            int xC = xCCorner + wC * ${dilationWidth};
 
             if (xC < 0 || xC >= ${xNumCols}) {
               continue;
--- a/src/ops/conv.ts
+++ b/src/ops/conv.ts
@@ -431,8 +431,8 @@ export class ConvOps {
   @operation
   static depthwiseConv2d<T extends Tensor3D|Tensor4D>(
       input: T, filter: Tensor4D, strides: [number, number]|number,
-      pad: 'valid'|'same'|number, dilations: [number, number]|number = [1, 1],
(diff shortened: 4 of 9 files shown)


oveddan:
This is not ready to merge yet, it's just been opened to get some feedback/review.  Particular things that I would like feedback on:

* Both backend_cpu`'s `conv2d` and `depthwiseConv2d` were inconsistent with the way their corresponding gpu methods were written, and I found it challenging to figure out how to use the dilation parameters in the cpu method.    I refactored those cpu methods to have similar logic to their corresponding gpu methods, and was able to more easily use dilation.  I'm not sure if the cpu methods were different by design for performance reasons.
* For testing atrous convolution, a standard filter was converted into the dilated version by padding it with 0s in between the values.  Then atrous convolution was performed using the standard filter, and standard convolution was done with the dilated version of the filter, and the results were compared for similarity. 
* The change to the apis of `conv1d` and `conv2d` would be a breaking change as it adds a positional argument of `dilations` in the middle of the arguments.  It may be better to add that parameter to the end, or just to create a separate method `atrousConv2d`.  

Besides the above requested feedback, still what needs to be done in this PR:
- [ ] raise errors when doing backprop on convolution layers where dilation rates are greater than 1
- [ ] If we are sticking with the breaking change, update all the models and examples in the repo to account for this new parameter.
- [ ] Add more tests for `depthwiseConv2d` where dilation is greater than 1.

dsmilkov:
Hi Dan,
This is an amazing PR. Replies below:

- Don't worry much about the backend_cpu implementations. We mostly use it for sanity checking our webgl implementations. Refactoring is fine.
- It's ok for testing atrous in a slightly different way.
- Regarding API, I left a comment below. We are ok with being a breaking change, but have some request for the order of params and adding dataFormat now.

Thanks again for the awesome PR.

---

Review status: 0 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/ops/conv.ts, line 129 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-L6swdmJhRmTaQI2V-1B:-L6swdmJhRmTaQI2V-1C:bslalms) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8ace2ee8d7da963d532090af80d515abd3789bb9/src/ops/conv.ts#L129)):*
> ```XML
>   static conv2d<T extends Tensor3D|Tensor4D>(
>       x: T, filter: Tensor4D, strides: [number, number]|number,
>       dilations: [number, number]|number = [1, 1], pad: 'valid'|'same'|number,
> ```

Let's add dilations after padding, but before dimRoundingMode. Also can you add dataFormat = 'NHWC' before the dilations, so we don't break users again when we add data format later.

Basically the signature should be x, filter, strides, pad, data_format='NHWC', dilations=[1, 1].

Throw an error if dataFormat is not "NHWC" for now. We can add the functionality later.

To the same for conv1d

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L6sxL8gLtL-YPH28yUM:b-y5m0gj)*

oveddan:
Regarding testing atrous convolution, I struggled to properly setup a test for `depthwiseConv2d` when there is dilation,  inDepth, and chMul greater than 1.

What I'm trying to do is populate the weight tensor with 0s to represent dilation, first columnwise then rowwise (first two dimensions) but I can't figure out the proper way.   

Here is the test code:

```typescript
    const fSize = 2;
    const pad = 'same';
    const stride = 1;
    const dilation = 2;
    const chMul = 2;
    const inDepth = 2;

    const x = dl.tensor4d(
        [
          0.675707, 0.758567, 0.413529, 0.963967, 0.217291, 0.101335, 0.804231,
          0.329673, 0.924503, 0.728742, 0.180217, 0.210459, 0.133869, 0.650827,
          0.047613, 0.554795, 0.653365, 0.442196
        ],
        [1, 3, 3, inDepth]);
    const w = dl.tensor4d(
        [
          0.347154, 0.386692, 0.327191, 0.483784, 0.591807, 0.24263, 0.95182,
          0.174353, 0.592136, 0.623469, 0.988244, 0.660731, 0.946534, 0.0801365,
          0.864889, 0.874602
        ],
        [fSize, fSize, inDepth, chMul]);
    // adding a dilation rate is equivalent to using a filter
    // with 0s for the dilation rate
    const fSizeDilated = fSize + (fSize - 1) * (dilation - 1);
    const wDilated = dl.tensor4d(
        [
          0.347154, 0, 0.386692,  0, 0, 0, 0.327191, 0, 0.483784,
          0.591807, 0, 0.24263,   0, 0, 0, 0.95182,  0, 0.174353,
          0.592136, 0, 0.623469,  0, 0, 0, 0.988244, 0, 0.660731,
          0.946534, 0, 0.0801365, 0, 0, 0, 0.864889, 0, 0.874602
        ],
        [fSizeDilated, fSizeDilated, inDepth, chMul],
    );

    const result = dl.depthwiseConv2d(x, w, stride, pad, dilation);

    const expectedResult = dl.depthwiseConv2d(x, wDilated, stride, pad);

    expect(result.shape).toEqual(expectedResult.shape);
    expectArraysClose(result, expectedResult);
```

This test fails because the matrix does not get populated the right way due to me not putting the 0s in the proper positions.  How should I write something that populates it the right way?  I've also tried something like:

```typescript
const wDilated = dl.tensor4d(
        [
          [[
            [0.347154, 0, 0.386692, 0, 0, 0, 0.327191, 0, 0.483784],
            [0.591807, 0, 0.24263, 0, 0, 0, 0.95182, 0, 0.174353]
          ]],
          [[
            [0.592136, 0, 0.623469, 0, 0, 0, 0.988244, 0, 0.660731],
            [0.946534, 0, 0.0801365, 0, 0, 0, 0.864889, 0, 0.874602]
          ]]
        ],
        [fSizeDilated, fSizeDilated, inDepth, chMul],
    );
```

but get the error: 	`Error: Error creating a new Tensor. Inferred shape (2,1,2,9) does not match the provided shape (3,3,2,2). Shapes 3,3,2,2 and 2,1,2,9 must match`

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/ops/conv.ts, line 129 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-L6swdmJhRmTaQI2V-1B:-L70JluOerVjRWW2vLnk:bud230g) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8ace2ee8d7da963d532090af80d515abd3789bb9/src/ops/conv.ts#L129)):*

Let's add dilations after padding, but before dimRoundingMode. Also can you add dataFormat = 'NHWC' before the dilations, so we don't break users again when we add data format later.

Basically the signature should be x, filter, strides, pad, data_format='NHWC', dilations=[1, 1].

Throw an error if dataFormat is not "NHWC" for now. We can add the functionality later.

To the same for conv1d
</blockquote></details>

Fixed in 25a532b.

The ordering here is slightly inconsistent, in order to match the ordering in Tensorflow:
* In `conv1d`, as in [tf.nn.conv1d](https://www.tensorflow.org/api_docs/python/tf/nn/conv1d), it's ordered `stride`, `pad`, `dataFormat`, `dilation`
* In `conv2d`, as in [tf.nn.conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d), it's ordered `strides`, `pad`, `dataFormat`, `dilations`
* In `depthwiseConv2d`, as in [tf.nn.depthwiseConv2d](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d), it's ordered `strides`, `pad`, `dilations`, `dataFormat`

Let me know if this is ok with you, or if you think it would be better to keep the ordering consistent within *deeplearn.js*.

For `conv1d` it accepts `NWC` or `NCW` but throws an error if it's not `NWC`

Also, TF doesn't have a `dilation` or `rate` parameter for [tf.nn.conv1d](https://www.tensorflow.org/api_docs/python/tf/nn/conv1d) - so not sure if we should get rid of that parameter that was added in this PR for that method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L70MlQ3PCkRc9do431b:b-bq3cuf)*

dsmilkov:
Sorry for the long delay. This is amazing work and would love to check this:

Regarding the test, maybe easier to populate the weights using a tensor buffer and using `buffer.set(value, ...coords)`:
```ts
const buf = dl.buffer<Rank.R4>([fSizeDilated, fSizeDilated, inDepth, chMul]);
buf.set(0.347154, 0, 0, 0, 0); // [0,0] in filter, 0 inDepth, 0 chMul
buf.set(0, 0, 1, 0, 0); // [0,1] in filter, 0 inDepth, 0 chMul
buf.set(0.386692, 0, 2, 0, 0); // [0, 2] in filter, 0 inDepth, 0 chMul
.....
wDilated = buf.toTensor();
```

Left 1 comment about the API. Thanks for summarizing the differences so nicely.

---

Reviewed 2 of 12 files at r1, 1 of 3 files at r2, 3 of 5 files at r4, 3 of 4 files at r5.

---

*[src/ops/conv.ts, line 129 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-L6swdmJhRmTaQI2V-1B:-L7r1lQFvfmuolghsht1:b-dy22vg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8ace2ee8d7da963d532090af80d515abd3789bb9/src/ops/conv.ts#L129)):*
<details><summary><i>Previously, oveddan (Dan Oved) wrote…</i></summary><blockquote>

Fixed in 25a532b.

The ordering here is slightly inconsistent, in order to match the ordering in Tensorflow:
* In `conv1d`, as in [tf.nn.conv1d](https://www.tensorflow.org/api_docs/python/tf/nn/conv1d), it's ordered `stride`, `pad`, `dataFormat`, `dilation`
* In `conv2d`, as in [tf.nn.conv2d](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d), it's ordered `strides`, `pad`, `dataFormat`, `dilations`
* In `depthwiseConv2d`, as in [tf.nn.depthwiseConv2d](https://www.tensorflow.org/api_docs/python/tf/nn/depthwise_conv2d), it's ordered `strides`, `pad`, `dilations`, `dataFormat`

Let me know if this is ok with you, or if you think it would be better to keep the ordering consistent within *deeplearn.js*.

For `conv1d` it accepts `NWC` or `NCW` but throws an error if it's not `NWC`

Also, TF doesn't have a `dilation` or `rate` parameter for [tf.nn.conv1d](https://www.tensorflow.org/api_docs/python/tf/nn/conv1d) - so not sure if we should get rid of that parameter that was added in this PR for that method.



</blockquote></details>

Thanks for the detailed list. Let's go with being consistent within deeplearn,
so have depthwiseConv2D take `strides, pad, dataFormat, dilations`.

Conv1 accepting nwc or ncw sounds great.
Keep dilation and rate in conv1d in deeplearn to be consistent across the board.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L7r2QomdIOKkntw5alb:b3tsz0t)*

oveddan:
Review status: 5 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/ops/conv.ts, line 129 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-L6swdmJhRmTaQI2V-1B:-L7wDkX82zyLKctGDcCi:b-m2aeb8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/8ace2ee8d7da963d532090af80d515abd3789bb9/src/ops/conv.ts#L129)):*

Thanks for the detailed list. Let's go with being consistent within deeplearn,
so have depthwiseConv2D take `strides, pad, dataFormat, dilations`.

Conv1 accepting nwc or ncw sounds great.
Keep dilation and rate in conv1d in deeplearn to be consistent across the board.
</blockquote></details>

Fixed 281335f0ed8f72e578a93d3dcc941705dbd533ba

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794)*

dsmilkov:
This looks great. Let me know when the depthwise tests are added, and we'll be good to go. Thanks for the amazing work Dan!

---

Reviewed 4 of 4 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L7zjr1QgjjyFY3qqCkx:bqswgzp)*

oveddan:
So I added a couple depthwise conv2d tests for atrous convolution, but in a different way than you suggested.   It would have been challenging to build up a 3x3x2x2 filter then set each value one by one.

It was easier to do it by building individual 2d filters with the desired values, then to `stack` them:

```typescript
  const w =
        dl.stack(
              [
                dl.tensor2d(
                    [0.614293, 0.0648011, 0.101113, 0.452887], [fSize, fSize]),
                dl.tensor2d(
                    [0.0582746, 0.426481, 0.872743, 0.765767], [fSize, fSize])
              ],
              2)
            .expandDims(3) as dl.Tensor4D;

    // adding a dilation rate is equivalent to using a filter
    // with 0s for the dilation rate
    const fSizeDilated = fSize + (fSize - 1) * (dilation - 1);
    const wDilated =
        dl.stack(
              [
                dl.tensor2d(
                    [0.614293, 0, 0.0648011, 0, 0, 0, 0.101113, 0, 0.452887],
                    [fSizeDilated, fSizeDilated]),
                dl.tensor2d(
                    [0.0582746, 0, 0.426481, 0, 0, 0, 0.872743, 0, 0.765767],
                    [fSizeDilated, fSizeDilated])
              ],
              2)
            .expandDims(3) as dl.Tensor4D;
```

I added a couple tests that do this, one with dilation of 2 and chMul of 1, and another with dilation of 2 and chMul of 2 (resulting in a 3x3x2x2 dilated filter).

---

Review status: 8 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L80X6LRWdtRgoxbNMLj:b-yqf1nv)*

dsmilkov:
:lgtm_strong: That's really smart. This looks ready to submit. Thanks again for the amazing work!

---

Reviewed 1 of 1 files at r7.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/794#-:-L82aMWOTicpQrN-ZJKz:b-ifnvkc)*

PR #798, opened 2018-02-28
Make getStream() sync

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/798)

### DIFF
--- a/src/contrib/data/dataset.ts
+++ b/src/contrib/data/dataset.ts
@@ -20,9 +20,8 @@ import * as seedrandom from 'seedrandom';
 
 import {BatchDataset} from './batch_dataset';
 import {computeDatasetStatistics, DatasetStatistics} from './statistics';
-import {DataStream} from './streams/data_stream';
+import {DataStream, streamFromFunction} from './streams/data_stream';
 import {streamFromConcatenated} from './streams/data_stream';
-import {streamFromFunction} from './streams/data_stream';
 import {streamFromItems} from './streams/data_stream';
 import {DatasetElement} from './types';
 
@@ -45,7 +44,7 @@ export abstract class Dataset {
    * Provide a new stream of elements.  Note this will also start new streams
    * from any underlying `Dataset`s.
    */
-  abstract async getStream(): Promise<DataStream<DatasetElement>>;
+  abstract getStream(): DataStream<DatasetElement>;
 
   // TODO(soergel): Make Datasets report whether repeated getStream() calls
   // produce the same result (e.g., reading from a file) or different results
@@ -95,9 +94,7 @@ export abstract class Dataset {
   filter(filterer: (value: DatasetElement) => boolean | Promise<boolean>):
       Dataset {
     const base = this;
-    return datasetFromStreamFn(async () => {
-      return (await base.getStream()).filter(filterer);
-    });
+    return datasetFromStreamFn(() => base.getStream().filter(filterer));
   }
 
   /**
@@ -111,9 +108,7 @@ export abstract class Dataset {
   map(transform: (value: DatasetElement) => DatasetElement |
           Promise<DatasetElement>): Dataset {
     const base = this;
-    return datasetFromStreamFn(async () => {
-      return (await base.getStream()).map(transform);
-    });
+    return datasetFromStreamFn(() => base.getStream().map(transform));
   }
 
   /**
@@ -144,9 +139,8 @@ export abstract class Dataset {
    */
   concatenate(dataset: Dataset): Dataset {
     const base = this;
-    return datasetFromStreamFn(async () => {
-      return (await base.getStream()).concatenate(await dataset.getStream());
-    });
+    return datasetFromStreamFn(
+        () => base.getStream().concatenate(dataset.getStream()));
   }
 
   /**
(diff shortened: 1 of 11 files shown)


davidsoergel:
I appreciate the motivation to reduce the async surface, but I'm not sure it's the right decision here.

A stream may require some initialization logic, and we can't know in advance if that may be unavoidably async.  In the simplest case, consider consider fetching a url with training data.  Logically the fetch happens once at the outset, and so makes sense as part of getStream().  Since fetch is async, getStream() must be, too.  Then the next() calls just walk through the fetched blob.

If getStream is sync, then we have to put any async initialization logic in next, i.e.

    async next(): T {
       if(!this.isInitialized()) {
            await this.initialize();
       }
       ...
    }

It just seems distasteful to me to have to do that test on every single next() call, when we already know that it will fire exactly once.

---

Also, why does it matter if getStream() is async?  We know the user will be operating in an async context soon enough anyway, when they call next().  A typical pattern is conceptually like

    async trainMyModel() {
        const stream = await this.dataset.getStream();
        await stream.forEach(batch=>this.model.train(batch));
        }
    }

But in practice the stream is hidden from the user anyway when they're operating at the Dataset level:

    async trainMyModel() {
        await this.dataset.forEach(batch=>this.model.train(batch));
    }

I.e., the Dataset-level API already bundles the getStream() and repeated next() calls into a single async call.

So I guess my question is: under what circumstance would someone want to create a stream, but not use it right away?

And, would it address your concern to make getStream() private, so that users feel they're simply operating on the Dataset as a whole without worrying about the underlying mechanics?  That mental model was my intent here anyway.

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-:-L6UPGu8wX5klE4EXzrU:brhkquy)*

dsmilkov:
If you can make getStream() private, that would be amazing.
My motivation behind this change is to simplify the test code of datasets, which gives me a window to the user facing API.

For example, if you remove calls to `getStream()` in csv_dataset_test, and minimize the number of promises the user has to await in order to read a row from a csv file, I'll be happy to drop this PR

davidsoergel:
Sounds good, I'll try making getStream private and see what happens!

---

Review status: 0 of 11 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-:-L6UXoiYhdlz49gEJqBS:b4n7ijd)*

dsmilkov:
Per offline discussion, reducing async methods improves debugability and simplifies code. This is ready for review. I think this can be orthogonal to making getStream private, both of which will simplify things a bit.

davidsoergel:
:lgtm_strong:

---

Reviewed 7 of 11 files at r1, 4 of 4 files at r2.

---

*[src/contrib/data/streams/data_stream.ts, line 453 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-L6XM5uyW_xSOG4Yn-7Y:-L6XM5uyW_xSOG4Yn-7Z:b-a25j8z) ([raw file](https://github.com/pair-code/deeplearnjs/blob/637b18935beb4a36662e3af4b10b7b55c6f4281d/src/contrib/data/streams/data_stream.ts#L453)):*
> paralellize
nit: typo

---

*[src/contrib/data/streams/data_stream.ts, line 506 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-L6XGVMj9_BwDai5Z89t:-L6XGVMj9_BwDai5Z89u:b-x48nq9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/637b18935beb4a36662e3af4b10b7b55c6f4281d/src/contrib/data/streams/data_stream.ts#L506)):*
> ```XML
>   next(): Promise<T> {
>     this.refill();
>     return this.buffer.shift();
> ```

Maybe add a comment: "This shift will never throw an error because the buffer is always full after a refill.  If the stream is exhausted, the buffer will be full of Promises that will resolve to the end-of-stream signal."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-:-L6XMiYX3XvuQ8rKeV_w:b-jbsf8z)*

dsmilkov:
Review status: 10 of 11 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[src/contrib/data/streams/data_stream.ts, line 453 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-L6XM5uyW_xSOG4Yn-7Y:-L6XcDDQ5_TTCqjnIIEd:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/637b18935beb4a36662e3af4b10b7b55c6f4281d/src/contrib/data/streams/data_stream.ts#L453)):*

> paralellize
nit: typo
</blockquote></details>

Done.

---

*[src/contrib/data/streams/data_stream.ts, line 506 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798#-L6XGVMj9_BwDai5Z89t:-L6XcEzhqt4umczlahJt:bcvjgzp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/637b18935beb4a36662e3af4b10b7b55c6f4281d/src/contrib/data/streams/data_stream.ts#L506)):*

Maybe add a comment: "This shift will never throw an error because the buffer is always full after a refill.  If the stream is exhausted, the buffer will be full of Promises that will resolve to the end-of-stream signal."
</blockquote></details>

Good idea. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/798)*

PR #803, opened 2018-03-02
Implemented Eager Version of the Complementary Color Prediction

jamesmtchou:
Updated the Complementary Color Prediction model to use the Eager Mode.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/803)


demos/complementary-color-prediction-eager/complementary-color-prediction-eager.ts
```diff
@@ -0,0 +1,415 @@
...
+    return dl.tidy(() => {
+      let cost: dl.Scalar = dl.scalar(0);
+      const { input, target } = this.data;
+      const lastBatchIndex = input.length - this.batchSize;
+      for (let i = 0; i <= lastBatchIndex; i += this.batchSize) {
+        const batchData = {
+          input: input.slice(i, i + this.batchSize),
+          target: target.slice(i, i + this.batchSize),
+        };
+        if (shouldFetchCost) {
```
pankajvishwani: I think, it should be the following:
```
      const batchCost = this.train1Batch(batchData, shouldFetchCost);
      if(shouldFetchCost){
        cost = cost.add(batchCost);
      }
```
Otherwise train1Batch will only be called when shouldFetchCost is set to true.

pankajvishwani: I think, it should be the following:
```
      const batchCost = this.train1Batch(batchData, shouldFetchCost);
      if(shouldFetchCost){
        cost = cost.add(batchCost);
      }
```
Otherwise train1Batch will only be called when shouldFetchCost is set to true.

pankajvishwani:
Can you also add code to serialize/de-serialize weights and biases? Something like this:
```
	serializeWeights(){
		const result = [];
		this.weights.forEach(t => {
			result.push({shape: t.shape, data: Array.prototype.slice.call(t.dataSync())});
		});
		return result;
	}

	serializeBiases(){
		const result = [];
		this.biases.forEach(v => {
			result.push(Array.prototype.slice.call(v.dataSync()));
		});
		return result;
	}

```

rwieruch:
Thanks for working on this PR. I guess it will help lots of people to update their NNs written with deeplearn.js from an older version 👍

dsmilkov:
Thanks for the great PR. We are in the process of modularizing and moving demos/examples out of the main repo. In this process, we want to reduce the number of demos that we maintain, and enable the community to share their demos more easily. We encourage you to take this code and make a new github repo for it. Feel free to post a link of that new repo here on this issue and we'll keep track of that for the future. Thank you!

rwieruch:
@dsmilkov do I understand correctly that you want to hand over the example projects to the community? It would be great to have at least one example/tutorial though for people to learn about the status quo usage of deeplearn.js. I would love to migrate my projects ([0](https://github.com/javascript-machine-learning/mnist-neural-network-deeplearnjs), [1](https://github.com/javascript-machine-learning/color-accessibility-neural-network-deeplearnjs)) to the recent version, but still waiting for a best practice approach which demonstrates the eager mode.

dsmilkov:
We are definitely going to have our own examples and tutorials. We are currently in a flux state where we are in the process of (re)writing several of our examples and tutorials that we will host in another repo and will remove the `demos/` folder from this repo. The new examples/demos will have more of an educational focus. Stay tuned!

dsmilkov:
Hi,

The team had some discussion and we decided to keep the number of [examples](https://github.com/tensorflow/tfjs-examples) small, so that it is easier to maintain. We would love to see this in an [Observable notebook](https://beta.observablehq.com/) and happy to provide guidance while moving the code there. We are also happy to see it use the higher-level [Layers API](https://github.com/tensorflow/tfjs-layers).

jamesmtchou:
@dsmilkov Yes! Will be glad to migrate the code to observable notebook & Layers API (<- very excited about this; major pain point while using eager mode) and this thread when that's done.

PR #807, opened 2018-03-02
Add a python script to write grouped weights and a manifest JSON to disk, and a loadWeights method on the JS side.

nsthorat:
This PR adds the functionality to serialize numpy arrays to disk, sharding and gathering into chunks of 4MB. Weights are organized by "groups", the bytes of the weights in each group get concatenated together, sharded into arraybuffers, and written to disk. This means arrays of different dtypes can be intermixed in this packing.

This allows us to optimally shard for Chrome's network cache max size (4MB) and it also allows us to fetch weights by group, without the need to fetch all groups.

#### python
- Adds `./scripts/write_weights.py` which contains the method write_weights. This method takes a group configuration, `weights_groups` that is a list of groups. Each group is a list of weight entries. Each weight entry looks like the following:
```
entry = {
  'name': 'weight1',
  'data': np.array([1, 2, 3], 'float32')
}
```
Weights groups would then look like:
```
weight_groups = [
  [group_0_entry1, group_0_entry2],
  [group_1_entry1, group_1_entry2],
]
```
Currently, this only supports numpy arrays as the data field. In the future, we can possibly take more types.
- `write_weights` also takes a `write_dir`, which is the directory to write the files to.
- `write_weights` also takes a `shard_size_bytes` which allows the user to specify the size of the shards in bytes.
- `write_weights` also takes a `write_manifest`, which allows you to turn off writing the manifest JSON to disk.

#### typescript
- Adds a `loadsWeights` method which takes a weights manifest object, a path prefix, and optionally the weights to fetch. If not providing the list of weights to fetch, all weights will be fetched.
- `loadWeights` abstracts away all the sharding logic, collecting underlying arraybuffers that may be packed and sharded, returning a map of names to `Tensors`.

This PR also:
- Adds infrastructure for unit testing python code which gets run by travis. Python files suffixed with `_test.py` will be run by the test discoverer. All pip dependencies can be declared in requirements.txt (which is the python standard).

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/807)


dsmilkov:
:lgtm_strong: Really nice work, especially how quickly you wrote this.

---

Reviewed 9 of 10 files at r2.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[scripts/run_python_tests.sh, line 18 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8v3Zfr08FGrcz3BF:-L6n8v3Zfr08FGrcz3BG:b2bowk6) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/run_python_tests.sh#L18)):*
> ```Shell
> 
> pip install -r ./scripts/requirements.txt --user
> python -m unittest discover scripts "*_test.py"
> ```

travis has python by default? sweet. not a priority, but in case it's easy, might be worth trying to run this in python3 by adding additional line:

```python3 -m unittest discover scripts "*_test.py"```
```


---

*[scripts/write_weights.py, line 53 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8PJqWvjpOG6VRWR6:-L6n8PJqWvjpOG6VRWR7:bu82v2f) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L53)):*
> Chrome

fyi: not just chrome, firefox, safari too

---

*[scripts/write_weights.py, line 54 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8dKeTOpwMDzrn-zW:-L6n8dKeTOpwMDzrn-zX:bz65br4) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L54)):*
> ```Python
>     shard_size_bytes: The size of shards in bytes. Defaults to 4MB, which is
>       Chrome's max file size for caching.
>     write_manifest: Whether to write the manifest JSON to disk.
> ```

append: `Defaults to true`

---

*[scripts/write_weights.py, line 56 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nDh_fg5-XAto6P4DS:-L6nDh_fg5-XAto6P4DT:bwifyve) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L56)):*
> ```Python
>     write_manifest: Whether to write the manifest JSON to disk.
>   Returns:
>     The weights manifest JSON string.
> ```

absolutely love this detailed docstring. For completeness, would be awesome to show a sample manifest json here in the `Returns:` section

---

*[scripts/write_weights.py, line 99 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nCTXosP5lw8zPHNPe:-L6nCTXosP5lw8zPHNPf:bdilti) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L99)):*
> ```Python
> 
>         bytes = data.tobytes()
>         group_bytes += bytes
> ```

if possible, consider making `group_bytes` a `BufferedWriter` to avoid mem copy every time you append. Looks like the [python2 version](https://docs.python.org/2/library/io.html#io.BufferedWriter) and [python3 version](https://docs.python.org/3/library/io.html#io.BufferedWriter) have the same API.

---

*[src/weights_loader.ts, line 35 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nETG0e8f3r2cyMyf8:-L6nETG0e8f3r2cyMyf9:b-e48xdf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader.ts#L35)):*
> ```XML
> const DTYPE_VALUE_SIZE_MAP: {[dtype: string]: number} = {
>   'float32': 4,
>   'int32': 4
> ```

avoiding bool for now? sgtm for now, just double checking

---

*[src/weights_loader.ts, line 38 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nEJBOxtNAqB7r5L2b:-L6nEJBOxtNAqB7r5L2c:b-cosa6i) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader.ts#L38)):*
> ```XML
> };
> 
> export class WeightsLoader {
> ```

Why have a class at all? Looks like loadWeights can be a standalone function.

---

*[src/weights_loader_test.ts, line 417 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nFdW4i9cTxXnmiy8t:-L6nFdW4i9cTxXnmiy8u:b-3kkcj7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader_test.ts#L417)):*
> ```XML
> 
>     const weightsNamesToFetch = ['doesntexist'];
>     dl.loadWeights(manifest, './', weightsNamesToFetch)
> ```

would something like this work:

```js
try {
 await dl.loadWeights();
 done();
} catch (ex) {
  done.fail(ex);
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-:-L6nG-LzNUSZYK7z49Zq:b-enn0jf)*

nsthorat:
Review status: 3 of 9 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

*[scripts/run_python_tests.sh, line 18 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8v3Zfr08FGrcz3BF:-L6nQc2luZSGckVzR3Zn:b-o0dw5q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/run_python_tests.sh#L18)):*

travis has python by default? sweet. not a priority, but in case it's easy, might be worth trying to run this in python3 by adding additional line:

```python3 -m unittest discover scripts "*_test.py"```
</blockquote></details>

Yeah, turns out it does. Added python 3 unit tests, and had to change some code to make it actually python3 compatible.

---

*[scripts/write_weights.py, line 53 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8PJqWvjpOG6VRWR6:-L6nIroOM_oZy-NZPETS:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L53)):*

> Chrome

fyi: not just chrome, firefox, safari too
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 54 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6n8dKeTOpwMDzrn-zW:-L6nIF3FGVpU3Mi8Ezi2:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L54)):*

append: `Defaults to true`
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 56 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nDh_fg5-XAto6P4DS:-L6nIHefdwRtGhhCZjqb:b-fflhqt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L56)):*

absolutely love this detailed docstring. For completeness, would be awesome to show a sample manifest json here in the `Returns:` section
</blockquote></details>

Great idea. Done.

---

*[scripts/write_weights.py, line 99 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nCTXosP5lw8zPHNPe:-L6nUkx0FWQIUxIrHlDP:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/scripts/write_weights.py#L99)):*

if possible, consider making `group_bytes` a `BufferedWriter` to avoid mem copy every time you append. Looks like the [python2 version](https://docs.python.org/2/library/io.html#io.BufferedWriter) and [python3 version](https://docs.python.org/3/library/io.html#io.BufferedWriter) have the same API.
</blockquote></details>

Done.

---

*[src/weights_loader.ts, line 35 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nETG0e8f3r2cyMyf8:-L6nHUd9QhaVWq89AoJ8:b4ckhul) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader.ts#L35)):*

avoiding bool for now? sgtm for now, just double checking
</blockquote></details>

Yep, avoiding it for right now. We can come back to this pretty easily, there's an explicit exception for non-supported dtypes.

---

*[src/weights_loader.ts, line 38 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nEJBOxtNAqB7r5L2b:-L6nHvf_E712PqsDzq39:b-vp5lvf) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader.ts#L38)):*

Why have a class at all? Looks like loadWeights can be a standalone function.
</blockquote></details>

Good catch, I was originally thinking to @doc it, but I think it makes more sense for it to not make it to docs.

---

*[src/weights_loader_test.ts, line 417 at r2](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nFdW4i9cTxXnmiy8t:-L6nKyrlD4gx4gA_UNwt:b-hhj7ix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/727571ba50d6df086a3c683169ab9bc792e32eae/src/weights_loader_test.ts#L417)):*

would something like this work:

```js
try {
 await dl.loadWeights();
 done();
} catch (ex) {
  done.fail(ex);
}
```
</blockquote></details>

Yes! This works great, thank you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807)*

nsthorat:
Thanks for the review. Will wait for @caisq to make sure this is sufficient from the python side.

caisq:
Review status: 3 of 9 files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[scripts/write_weights.py, line 19 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6ns3OnlUGMzKGzuRaV:-L6ns3OnlUGMzKGzuRaW:b-m8701y) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L19)):*
> import numpy as np
To be consistent with google python style, "import numpy as np" should be moved to an import section below this first one. The section ordering is:
1) built-in 
2) third-party libraries such as numpy
3) modules from the library itself.
The sections should be separated by an empty line.

---

*[scripts/write_weights.py, line 29 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nscmpeyIPTpX019CH:-L6nscmpeyIPTpX019CI:bv8fak7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L29)):*
> small weights are packed into 4MB shards.
Not sure what this means. Does this mean that even if you have only one 4-byte weight, you will end up with a 4MB file?

---

*[scripts/write_weights.py, line 49 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1BxC8Cxqgt-I5E9f:-L6o1BxC8Cxqgt-I5E9g:b487yvm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L49)):*
> all groups
--> all groups and all entries?

---

*[scripts/write_weights.py, line 52 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIe7c7zb5LDbxqdoY:-L6oIe7c7zb5LDbxqdoZ:bc5lvjj) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L52)):*
> ```Python
>       numpy ndarray.
>     write_dir: A directory to write the files to.
>     shard_size_bytes: The size of shards in bytes. Defaults to 4MB, which is
> ```

You probably want to raise a ValueError below if shard_size_bytes is <= 0.

---

*[scripts/write_weights.py, line 63 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1aWU-5ozrKKTLvoI:-L6o1aWU-5ozrKKTLvoJ:bknqv16) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L63)):*
> [{
Add ```json or ```javascript, whichever one works?

---

*[scripts/write_weights.py, line 64 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o2Dz3LTBRqHT5FRtX:-L6o2Dz3LTBRqHT5FRtY:b-iki8bd) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L64)):*
> group0-000001-of-000002

The name could be easier to follow if it is formatted as 'group0-shard000001-of-000002'. Also, I wonder whether it is necessary to have so many zeros and hyphens. Why not just "group0-shard1of2"? I also find it a little inconsistent to use zero-based indices for groups, but 1-based indices for shards. I suggest using 1-based  throughout, because it's more intuitive and less prone to confusion.

---

*[scripts/write_weights.py, line 64 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oL1ew_0hqWaUQ3G2V:-L6oL1ew_0hqWaUQ3G2W:b-3pilkg) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L64)):*
> filepaths'
I thought in the design doc we converged on 'paths' instead of 'filepaths'? 'paths' is more succinct and general.

---

*[scripts/write_weights.py, line 69 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1Pb6iix7sGbUZfZN:-L6o1Pb6iix7sGbUZfZO:b1g0nng) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L69)):*
>  [{
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>
>       'filepaths': ['group0-000001-of-000002', 'group0-000002-of-000002'],
>       'weights': [{
>         'name': 'weight1',
>         'shape': [1000, 1000],
>         'dtype': 'float32'
>       }]
>     }
</details>
I'd like to see the indentation become stricter here. Maybe use an auto JSON formatter to do it.

---

*[scripts/write_weights.py, line 83 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1lQeWDwqk8NdxUf2:-L6o1lQeWDwqk8NdxUf3:bct17z8) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L83)):*
> ```Python
>     write_manifest=True):
>   if not isinstance(weight_groups, list):
>     raise Exception('weight_groups must be a list of groups')
> ```

For tensorflow and keras (and I guess ML-oriented Python in general), it is more typical to raise `ValueError` instead of  the very low level class `Exception`. Same elsewhere.

---

*[scripts/write_weights.py, line 83 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oFMMBqwCEica8rtPB:-L6oFMMBqwCEica8rtPC:b-sq2v8t) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L83)):*
> Exception('weight_groups must be a list of groups')
You might want to consider relaxing the requirement a little: it could also be a tuple or any iterable from what I see in the code below.

---

*[scripts/write_weights.py, line 84 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oFY3BvTDFYPQt3q2B:-L6oFY3BvTDFYPQt3q2C:b36zes5) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L84)):*
> len(weight_groups) == 0:
Google Python style suggests: `if not weight_groups:`
See https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

---

*[scripts/write_weights.py, line 86 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oGTzX587T2_tHijAV:-L6oGTzX587T2_tHijAW:b-7shc0m) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L86)):*
> i in range(len(weight_groups)):
You can avoid using `range` and simplify the code below a little by using `enumerate`.

Also, this function is a little long. You may consider refactor this sanity check loop into a module-private helper function.

---

*[scripts/write_weights.py, line 87 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oG6hysOMCE2PhLJef:-L6oG6hysOMCE2PhLJeg:bp41znl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L87)):*
> isinstance(weight_groups[i], list):
Like above: you just need this to be an iterable.

---

*[scripts/write_weights.py, line 103 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oI3LClI4GNocPKdEq:-L6oI3LClI4GNocPKdEr:bbh6vxt) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L103)):*
> ```Python
>             'array')
> 
>   weight_names = {}
> ```

This can done more efficiently as:
weight_names = set()

and below:
weight_names.add(name)
if name in weight_names:
   ...

---

*[scripts/write_weights.py, line 119 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oHFc5duyUW50X5j8i:-L6oHFc5duyUW50X5j8j:bm8b27d) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L119)):*
> Exception('Error dumping weight ' + name + ' dtype ' +
>               data.dtype.name + ' from not supported.')
FYI, for Keras this is probably fine, because Keras weights default to float32. But with generic tensorflow graphs this will probably cause relatively frequent troubles, because tensorflow allows various dtypes such as float32, int16, etc. You might want to add a TODO item for automatic dtype conversion during this function and the loading of the saved artifacts.

---

*[scripts/write_weights.py, line 130 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIIBOEvKvvAen-4DU:-L6oIIBOEvKvvAen-4DV:b-lcqfvp) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L130)):*
>  weight_names[name] = True
See my comment about using set() above.

---

*[scripts/write_weights.py, line 133 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIPhZgXigSPsnJ6qS:-L6oIPhZgXigSPsnJ6qT:b-hsrk1q) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L133)):*
> [d for d in data.shape]
You can do: `list(data.shape)`

---

*[scripts/write_weights_test.py, line 17 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJgK5oaIDsJGypUUB:-L6oJgK5oaIDsJGypUUC:b6xup6h) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L17)):*
> import numpy as np
See my comment about putting imports in sections above.

---

*[scripts/write_weights_test.py, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJldXFlOKoMBw3AdZ:-L6oJldXFlOKoMBw3Ad_:bebgity) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L20)):*
> import shutil
Make sure that in each import section, the imports are sorted alphabetically.

---

*[scripts/write_weights_test.py, line 25 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJ1RCevAE1qGjvtXz:-L6oJ1RCevAE1qGjvtY-:bpqeb7s) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L25)):*
>  
nit: According to Google Python style, there should be two empty lines separating class, function and constant definitions.

---

*[scripts/write_weights_test.py, line 28 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJzuIlh_7t0-mUHKM:-L6oJzuIlh_7t0-mUHKN:b-c385o9) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L28)):*
> path.exists(
Use os.path.isdir. (Same below).

---

*[src/weights_loader.ts, line 173 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oMGXd-qfwT95_HvO8:-L6oMGXd-qfwT95_HvO9:b7wp6xn) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/src/weights_loader.ts#L173)):*
> weightsTensorMap[weightsEntry.manifestEntry.name]
Throw an Error if duplicate weights names are found?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807)*

nsthorat:
Review status: 3 of 9 files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[scripts/write_weights.py, line 19 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6ns3OnlUGMzKGzuRaV:-L6qZY99N88kFHLi7Xxv:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L19)):*

> import numpy as np
To be consistent with google python style, "import numpy as np" should be moved to an import section below this first one. The section ordering is:
1) built-in 
2) third-party libraries such as numpy
3) modules from the library itself.
The sections should be separated by an empty line.
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 29 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6nscmpeyIPTpX019CH:-L6q_S-MEFbavuZwGQC2:b-bxx7ya) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L29)):*

> small weights are packed into 4MB shards.
Not sure what this means. Does this mean that even if you have only one 4-byte weight, you will end up with a 4MB file?
</blockquote></details>

No, it will be a single 4 byte file (if it's the only weight per group). I updated the docstring to make this clearer.

---

*[scripts/write_weights.py, line 49 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1BxC8Cxqgt-I5E9f:-L6q_bmj2VCIzMJs5GGK:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L49)):*

> all groups
--> all groups and all entries?
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 52 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIe7c7zb5LDbxqdoY:-L6qaZVdEL3yucJZ4Bcp:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L52)):*

You probably want to raise a ValueError below if shard_size_bytes is <= 0.
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 63 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1aWU-5ozrKKTLvoI:-L6qn4XWTuIRHHLz8IDI:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L63)):*

> [{
Add ```json or ```javascript, whichever one works?
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 64 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oL1ew_0hqWaUQ3G2V:-L6rKwdz4eh8NTGdKMfR:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L64)):*

> filepaths'
I thought in the design doc we converged on 'paths' instead of 'filepaths'? 'paths' is more succinct and general.
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 64 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o2Dz3LTBRqHT5FRtX:-L6rLI_Pg49uRK1RB1W0:b5q5fza) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L64)):*

> group0-000001-of-000002

The name could be easier to follow if it is formatted as 'group0-shard000001-of-000002'. Also, I wonder whether it is necessary to have so many zeros and hyphens. Why not just "group0-shard1of2"? I also find it a little inconsistent to use zero-based indices for groups, but 1-based indices for shards. I suggest using 1-based  throughout, because it's more intuitive and less prone to confusion.
</blockquote></details>

I agree. I accepted suggestion. I did it this way for some parity with TF checkpoints, but that's not important.

---

*[scripts/write_weights.py, line 69 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1Pb6iix7sGbUZfZN:-L6rL2MOIJgycaI3WKZI:b-aprsiy) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L69)):*

>  [{
>       'filepaths': ['group0-000001-of-000002', 'group0-000002-of-000002'],
>       'weights': [{
>         'name': 'weight1',
>         'shape': [1000, 1000],
>         'dtype': 'float32'
>       }]
>     }
I'd like to see the indentation become stricter here. Maybe use an auto JSON formatter to do it.
</blockquote></details>

What is the problem with this indentation? I think this is the clearest way to present this block.

---

*[scripts/write_weights.py, line 83 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6o1lQeWDwqk8NdxUf2:-L6rOtEvjlOFatS4PO-1:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L83)):*

For tensorflow and keras (and I guess ML-oriented Python in general), it is more typical to raise `ValueError` instead of  the very low level class `Exception`. Same elsewhere.
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 83 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oFMMBqwCEica8rtPB:-L6rOtSakHQZ9MXhmi-e:b6n3yrc) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L83)):*

> Exception('weight_groups must be a list of groups')
You might want to consider relaxing the requirement a little: it could also be a tuple or any iterable from what I see in the code below.
</blockquote></details>

I'm going to gently push back here. I don't want to allow any iterables (I don't want to get bitten by the "pass a string is fine because it's iterable" bug).

---

*[scripts/write_weights.py, line 84 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oFY3BvTDFYPQt3q2B:-L6rPACSH-IGL4dAAOat:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L84)):*

> len(weight_groups) == 0:
Google Python style suggests: `if not weight_groups:`
See https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 86 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oGTzX587T2_tHijAV:-L6rP47FJ4uRCQas0xu7:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L86)):*

> i in range(len(weight_groups)):
You can avoid using `range` and simplify the code below a little by using `enumerate`.

Also, this function is a little long. You may consider refactor this sanity check loop into a module-private helper function.
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 87 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oG6hysOMCE2PhLJef:-L6rPB-87aVgd5M6h8-S:b4cee5u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L87)):*

> isinstance(weight_groups[i], list):
Like above: you just need this to be an iterable.
</blockquote></details>

I don't want to allow any iterable here (e.g. string is iterable). I really want to restrict it exactly to lists.

---

*[scripts/write_weights.py, line 103 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oI3LClI4GNocPKdEq:-L6rPIj4BRP9KAopsbs7:bh06whm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L103)):*

This can done more efficiently as:
weight_names = set()

and below:
weight_names.add(name)
if name in weight_names:
   ...
</blockquote></details>

Can't use set. Doesn't work in both 2 and 3, and didn't want to introduce a dependency on six.

---

*[scripts/write_weights.py, line 119 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oHFc5duyUW50X5j8i:-L6rPOdq9EH__1hMDdbt:b-am0ztw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L119)):*

> Exception('Error dumping weight ' + name + ' dtype ' +
>               data.dtype.name + ' from not supported.')
FYI, for Keras this is probably fine, because Keras weights default to float32. But with generic tensorflow graphs this will probably cause relatively frequent troubles, because tensorflow allows various dtypes such as float32, int16, etc. You might want to add a TODO item for automatic dtype conversion during this function and the loading of the saved artifacts.
</blockquote></details>

Yep, I'm aware of our lack of support for other dtypes, I've just implemented the ones we've implemented in deeplearn.js (besides boolean which needs more careful though since we represent a bit with a full byte in JS world, but the wire transfer format should probably only encode bits).

Added a TODO at the top.

---

*[scripts/write_weights.py, line 130 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIIBOEvKvvAen-4DU:-L6rPtE6u7wfaCcZ5Pyc:b-i8uemr) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L130)):*

>  weight_names[name] = True
See my comment about using set() above.
</blockquote></details>

Can't use set, not supported in 2 and 3 which is why I used a dict.

---

*[scripts/write_weights.py, line 133 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIPhZgXigSPsnJ6qS:-L6rPz2EN8IKo12mMuVK:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L133)):*

> [d for d in data.shape]
You can do: `list(data.shape)`
</blockquote></details>

Done.

---

*[scripts/write_weights_test.py, line 17 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJgK5oaIDsJGypUUB:-L6rPzl7ow2vMvhAQaCv:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L17)):*

> import numpy as np
See my comment about putting imports in sections above.
</blockquote></details>

Done.

---

*[scripts/write_weights_test.py, line 20 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJldXFlOKoMBw3AdZ:-L6rQ2jwT5Oi9aN2eKzI:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L20)):*

> import shutil
Make sure that in each import section, the imports are sorted alphabetically.
</blockquote></details>

Done.

---

*[scripts/write_weights_test.py, line 25 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJ1RCevAE1qGjvtXz:-L6rQ4-JRadJkWSSX735:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L25)):*

>  
nit: According to Google Python style, there should be two empty lines separating class, function and constant definitions.
</blockquote></details>

Done.

---

*[scripts/write_weights_test.py, line 28 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oJzuIlh_7t0-mUHKM:-L6rQ8SFzqZbKW5p_84W:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights_test.py#L28)):*

> path.exists(
Use os.path.isdir. (Same below).
</blockquote></details>

Done.

---

*[src/weights_loader.ts, line 173 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oMGXd-qfwT95_HvO8:-L6rQenVVC1jfVNYKvDG:b-d5wzc2) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/src/weights_loader.ts#L173)):*

> weightsTensorMap[weightsEntry.manifestEntry.name]
Throw an Error if duplicate weights names are found?
</blockquote></details>

Done. Just generally this should never happen since we control the manifest, but I see your point if someone manually edits it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807)*

caisq:
Review status: 3 of 9 files reviewed at latest revision, 33 unresolved discussions, some commit checks pending.

---

*[scripts/write_weights.py, line 130 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIIBOEvKvvAen-4DU:-L6rSzFO20goyq9XM117:b-nowetm) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L130)):*

Can't use set, not supported in 2 and 3 which is why I used a dict.
</blockquote></details>

That doesn't sound right. set exists in Python3: https://docs.python.org/3/tutorial/datastructures.html#sets
What error are you getting?

---

*[scripts/write_weights.py, line 27 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rRunb7TinSKHf7po3:-L6rRunb7TinSKHf7po4:bhawxl) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L27)):*
> ```Python
> # TODO(nsthorat): Support more than just float32 and int32 for weight dumping.
> DTYPE_BYTES = {'float32': 4, 'int32': 4}
> ```

nit: need two empty lines here per Google Python style

---

*[scripts/write_weights.py, line 28 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rRxuYHTPTolnbbwqB:-L6rRxuYHTPTolnbbwqC:byi2qdw) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L28)):*
> """

nit: this doc string should live in the body of the python function. (that is how Python doc generators work.) See example: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/ops.py#L69

---

*[scripts/write_weights.py, line 85 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rSQYKIUgv7gbdpofN:-L6rSQYKIUgv7gbdpofO:b-coebky) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L85)):*
> ```Python
>     weight_groups, write_dir, shard_size_bytes = 1024 * 1024 * 4,
>     write_manifest=True):
>   assert_weight_groups_valid(weight_groups)
> ```

Rename as _assert_weight_groups_valid() to indicate it is module-private (unless you want to use the funciton in other modules, which I think you don't). Same for the other helper functions.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807)*

nsthorat:
Review status: 3 of 9 files reviewed at latest revision, 33 unresolved discussions, some commit checks failed.

---

*[scripts/write_weights.py, line 130 at r3](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6oIIBOEvKvvAen-4DU:-L6rUoviSU5Esihz7Sx9:b-dx288u) ([raw file](https://github.com/pair-code/deeplearnjs/blob/efd5cef37b693971c99a4d7b5990bd2d9cb3dba2/scripts/write_weights.py#L130)):*

That doesn't sound right. set exists in Python3: https://docs.python.org/3/tutorial/datastructures.html#sets
What error are you getting?
</blockquote></details>

Set exists in python 3 out of the box, in Python 2 you have to import Set, which fails in Python 3.

---

*[scripts/write_weights.py, line 27 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rRunb7TinSKHf7po3:-L6rVMibO1VMllPM-TUY:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L27)):*

nit: need two empty lines here per Google Python style
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 28 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rRxuYHTPTolnbbwqB:-L6rVMQPL12ziAdi2qBb:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L28)):*

> """

nit: this doc string should live in the body of the python function. (that is how Python doc generators work.) See example: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/ops.py#L69
</blockquote></details>

Done.

---

*[scripts/write_weights.py, line 85 at r4](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807#-L6rSQYKIUgv7gbdpofN:-L6rVJaeKfHPQctifImm:b-896fix) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c4a1502e5094692df927fc3bb4fe1a89b1b33bd6/scripts/write_weights.py#L85)):*

Rename as _assert_weight_groups_valid() to indicate it is module-private (unless you want to use the funciton in other modules, which I think you don't). Same for the other helper functions.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/807)*

PR #809, opened 2018-03-04
Introduce flatmap, using a queue; don't use queues for map, filter, or batch.

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/809)


davidsoergel:
Abandoning per #873.

PR #810, opened 2018-03-04
Streaming parser for IDX files (the native format of MNIST), and associated IDXDataset.

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/810)


davidsoergel:
Abandoning per #873.

PR #821, opened 2018-03-07
Remove deprecated API

dsmilkov:
Remove:
 - math
 - static methods on Tensor
 - add `normalized` to dl.multinomial() and expose to the end user
 - error if f passed in tidy() returns a Promise.

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/821)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 21 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/tensor.ts, line 297 at r1](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/821#-L70Ai96DAMjQXRvlXWV:-L70Ai96DAMjQXRvlXWW:b-njcnm7) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fc9965a5327ec3ed3a4e77783adaf271901126fc/src/tensor.ts#L297)):*
> ```XML
>   get(...locs: number[]) {
>     this.throwIfDisposed();
>     return this.buffer().get(...locs);
> ```

just to be clear, we're relying on dataSync() caching locally and buffer() being a cheap wrapper, right?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/pair-code/deeplearnjs/821#-:-L70B9i94ru8bpmuocPz:b-jbsf8z)*

dsmilkov:
Review status: 0 of 34 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/tensor.ts, line 297 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/821#-L70Ai96DAMjQXRvlXWV:-L70GaLggKCObIcT-kLa:b-gxgs4y) ([raw file](https://github.com/pair-code/deeplearnjs/blob/fc9965a5327ec3ed3a4e77783adaf271901126fc/src/tensor.ts#L297)):*

just to be clear, we're relying on dataSync() caching locally and buffer() being a cheap wrapper, right?
</blockquote></details>

That's what I expected, but found out that buffer() is not that cheap, so keeping the old implementation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/821)*

PR #825, opened 2018-03-08
Unary Ops: Adds log1p to the api

ManrajGrover:
This PR adds unary operator `log1p` to the api.

https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.log1p.html#numpy.log1p

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/825)

### DIFF
--- a/src/kernels/backend.ts
+++ b/src/kernels/backend.ts
@@ -109,6 +109,7 @@ export interface KernelBackend extends TensorStorage, BackendTimer {
   pow<T extends Tensor>(a: T, b: Tensor): T;
   exp<T extends Tensor>(x: T): T;
   log<T extends Tensor>(x: T): T;
+  log1p<T extends Tensor>(x: T): T;
   sqrt<T extends Tensor>(x: T): T;
 
   square<T extends Tensor>(x: T): T;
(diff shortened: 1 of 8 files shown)


nsthorat:
:lgtm_strong:

Awesome stuff! One minor comment.

---

Review status: 0 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/ops/unary_ops.ts, line 144 at r1](https://reviewable.io:443/reviews/pair-code/deeplearnjs/825#-L7KgWi2rHeXrFNknFUP:-L7KgWi2rHeXrFNknFUQ:bj1ad17) ([raw file](https://github.com/pair-code/deeplearnjs/blob/c31f3117013b862554dbc9caede59e28dab2a3d1/src/ops/unary_ops.ts#L144)):*
> ```XML
>   static log1p<T extends Tensor>(x: T): T {
>     const grad = (dy: T) => {
>       return {x: () => dy.divStrict(x.add(ops.scalar(1).toFloat()))};
> ```

do you need to call toFloat() here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/pair-code/deeplearnjs/825#-:-L7Kgdk8FeEyZHvJbDJX:b-z4zzc3)*

PR #831, opened 2018-03-09
Update 'Training for ML Beginners' to work with API found in version 0.5.1

lukebelliveau:
While walking through the [Training For ML Beginners](https://deeplearnjs.org/docs/tutorials/ml_beginners.html), I found the example wasn't quite compatible with the current API version published on npm. This updates the tutorial to the current API.

I confirmed this works, but I'm sure there are some things that don't look quite right - I'm new here :). If y'all are interested in this change, I'm happy to keep working on it if it needs revisions.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/831)

### DIFF
--- a/docs/tutorials/ml_beginners.md
+++ b/docs/tutorials/ml_beginners.md
@@ -37,14 +37,14 @@ import * as dl from 'deeplearn';
 // Step 1. Set up variables, these are the things we want the model
 // to learn in order to do prediction accurately. We will initialize
 // them with random values.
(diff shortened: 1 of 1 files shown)


dsmilkov:
Hi @lukebelliveau,

You seem to be using the deprecated API, which we will remove in 0.6.0. The tutorial is written to use only the api that is on [this page](https://deeplearnjs.org/docs/api/index.html).

PR #842, opened 2018-03-10
Unary Ops: Adds exp2 to the api

ManrajGrover:
This PR adds unary operation `dl.exp2` to the api.

https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.exp2.html#numpy.exp2

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/842)


dsmilkov:
Hey Manraj, since we are aligning our API with TF and `exp2` doesn't exist in TensorFlow, let's drop this op.

Thanks for the understanding!

PR #843, opened 2018-03-10
Unary Ops: Adds dl.log2 to the api

ManrajGrover:
This PR adds unary operation `dl.log2` to the api.

https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.log2.html#numpy.log2

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/843)


dsmilkov:
Hey Manraj, since we are aligning our API with TF and exp2 doesn't exist in TensorFlow, let's drop this op.

Thanks for the understanding!

PR #844, opened 2018-03-10
Unary Ops: Adds dl.log10 to the api

ManrajGrover:
This PR adds unary operation `dl.log10` to the api.

https://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.log10.html#numpy.log10

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/844)


dsmilkov:
Same here. Since we are aligning our API with TF and exp2 doesn't exist in TensorFlow, let's drop this op.

Thanks for the understanding!

PR #859, opened 2018-03-17
WebGL Backend: Centralize async reading of data, postpone disposal if a pending read

dsmilkov:
All the work involves only the WebGL backend.

- Centralize async reading of data in a single setTimeout loop
- Postpone disposal if there is a pending read
- Use LRU cache to page textures out of the GPU memory and into CPU (Avoids crashing browser)
- Some optimizations:
  - Cache the `ext.GPU_DISJOINT_EXT` parameter since reading it is expensive
  - Do binary search for queryIsDone since checking is expensive
  - Reuse download buffers for readPixels to minimize js memory trashing

Did a stress test using demos/mnist_eager where I made tensor.dispose() a no-op. There is significant slowdown (~10x) due to constant downloads and js GC trashing memory, but the browser doesn't crash. The optimizations I did were motivated by the profiling from this stress test.

Fixes https://github.com/tensorflow/tfjs/issues/394

This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/859)


nsthorat:
:lgtm_strong:

Nice work.

Can you add a unit tests where you mock window screen width and allocate a bunch of tensors and make sure nothing blows up?

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/kernels/backend_webgl.ts, line 90 at r3](https://reviewable.io/reviews/tensorflow/tfjs-core/859#-LEGuQ8j3JGLF9gsBraQ:-LEGuQ8j3JGLF9gsBraR:b-tc0mpl) ([raw file](https://github.com/tensorflow/tfjs-core/blob/cb26c21b074e0f2d97879ed5cd045db02626c5a8/src/kernels/backend_webgl.ts#L90)):*
> ```XML
> }
> 
> // Empirically determined constant used to decide the number of bytes on GPU
> ```

this makes it seem like this is the number of bytes. Maybe we can specify the bytes are this constant * screen area * dpi

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/859#-:-LEGuC1cBH0Q2Ac2Rl_5:btvo6xx)*

PR #868, opened 2018-03-19
Unary Ops: Adds tf.reciprocal

ManrajGrover:
This PR adds unary operation `tf.reciprocal` to the API.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/868)


ManrajGrover:
Hi @nsthorat,

> Thanks for these awesome changes. When you're adding these in the future, let's make sure to try to not implement every TF op, since it could get out of hand quickly. At this point, we have pretty good coverage and we should try to add these based on real world use cases.

Sure, it makes sense to prioritize and add operations based on use case. This will also help keep bundle size small.

> Another point (which you've been following), one way to tell if you should write a backend (webgl shader) or not is whether TF uses the C API or implements the op in TensorFlow PY.
>
> If you see docs here: https://www.tensorflow.org/api_docs/python/tf/reciprocal, you'll see it's implemented in "gen_math_ops" which is the generated binding to the C API (this means they have a C implementation of tf.reciprocal and we should also). You're doing that correctly here, but just wanted to make sure we're on the same page.

Thanks for sharing this. I've been analyzing Tensorflow implementations and outputs before implementing them here.

PR #881, opened 2018-03-22
Inform backends when tensors reshape.

nkreeger:
Some backends might need to know when a Tensor reshapes to update any backing. This change introduces a new API on the backend to inform backends that a Tensor is reshaped.

I'm not 100% how to test this since most of this functionality is protected through the Tensor API. Open to suggestions if you have any.

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/881)


dsmilkov:
This is obsolete now due to #893 right?

PR #887, opened 2018-03-23
Binary Ops: Adds tf.mod operation

### LINKED ISSUE #884: add mod op?
```js
dl.tensor1d([11,22,33,44,55,66,77,88,99]).mod(10);
// [1,2,3,4,5,6,7,8,9]
```

ManrajGrover:
This PR adds `tf.mod` operation to the API.

Fixes #884

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/887)


ManrajGrover:
While working on this I found that due to low precision, values like `0.5999999` get floored to `0.5` which technically was `0.6` and `floor(0.6)` should be `0.6`. Creating an issue for this if not already opened.

dsmilkov:
Nice work Manraj. Few comments, but we are very close.

---

Reviewed 7 of 8 files at r2.
Review status: 7 of 8 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/kernels/webgl/binaryop_gpu.ts, line 79 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92frOhASX8R4GisIyO:-L92frOhASX8R4GisIyP:buu1x6) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/kernels/webgl/binaryop_gpu.ts#L79)):*
> ```XML
>     return rem;
>   } else {
>     return mod(rem + b, b);
> ```

do you need this check in glsl and the special case? Based on https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mod.xhtml it looks like the built-in `mod(x,y)` is computed as `x - y * floor(x/y)` which is exactly aligned with TF's mod.

---

*[src/ops/binary_ops.ts, line 373 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92i-LmAG0xjgyPdFAa:-L92i-LmAG0xjgyPdFAb:biyd2yq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/ops/binary_ops.ts#L373)):*
> ```XML
>    * const b = dl.tensor1d([1, 2, 9, 4]);
>    *
>    * a.mod(b).print();  // or dl.mod(a, b)
> ```

`tf.` in the doc snippet.

---

*[src/ops/binary_ops.ts, line 393 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92h4MKB3mw4ylOtgCr:-L92h4MKB3mw4ylOtgCs:b7ox9t8) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/ops/binary_ops.ts#L393)):*
> ```XML
>     broadcast_util.assertAndGetBroadcastShape(a.shape, b.shape);
>     const der = (dy: Tensor) => {
>       const derA = () => dy.mul(onesLike(a));
> ```

mod supports broadcasting, which means you'll have to reduce the derA to have the shape of `a`, which means you have to dy.sum(reductionAxis). See gradient in our tf.mul() for an example: https://github.com/tensorflow/tfjs-core/blob/master/src/ops/binary_ops.ts#L259

---

*[src/ops/binary_ops_test.ts, line 341 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92e3Z31GnvkFiFxv5B:-L92e3Z31GnvkFiFxv5C:bxg7xof) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/ops/binary_ops_test.ts#L341)):*
> ```XML
> 
>     expect(result.shape).toEqual(a.shape);
>     expectArraysClose(result, [0.09999999, 0.19999996, 0.15, 0.05000016]);
> ```

instead of using values like 0.09999 and 0.199999996, use the correct expected values like 0.1 and 0.2. If the test fails with those values, it will be highly informative, since it tells us that we should improve the implementation to be more numerically stable.

---

*[src/ops/binary_ops_test.ts, line 376 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92f0kO2Y0rTlFnKi68:-L92f0kO2Y0rTlFnKi69:b-g4epwr) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/ops/binary_ops_test.ts#L376)):*
> ```XML
> 
>     expect(result.shape).toEqual(a.shape);
>     expectArraysClose(result, [0.5, 0.0999999, 0.5, 0.20000017]);
> ```

same here: use the correct math values e.g. 0.1. Likewise for the other tests in this file.

---

*[src/ops/binary_ops_test.ts, line 461 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-L92htrMAQIJCIv4fWme:-L92htrMAQIJCIv4fWmf:b-wnkxtb) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8a9cc02eec7df13a64348f0c4bf4c5a3584415fe/src/ops/binary_ops_test.ts#L461)):*
> ```XML
>       3 * -1 * Math.floor(0.7 / 0.7), 4 * -1 * Math.floor(0.91 / 0.15)
>     ]);
>   });
> ```

add a gradient test that uses broadcasting.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-:-L92i38D22vbXtDzI-cJ:b-ncc2dn)*

ManrajGrover:
Thanks @dsmilkov 

> do you need this check in glsl and the special case? Based on
> https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/mod.xhtml it looks like the built-in
> mod(x,y) is computed as x - y * floor(x/y) which is exactly aligned with TF's mod.

Actually, using plain mod(x, y) was failing for following case:

```js
const a = dl.scalar(0.6);
const b = dl.tensor1d([0.5, 3, -0.1, -4]);
const result = dl.mod(a, b);
```

Particularly, case when `mod(0.6, -0.1)`, expected value is `-0.1` but webgl implementation returns `1.4901161193847656e-8`. I've made changes in test cases and pushed with webgl implementation for clarity.

Incorporating changes for broadcast.

dsmilkov:
Hi @ManrajGrover , I just verified that in TF tf.mod(0.6, -0.1) is returning `-0.1`, however I believe that to be a bug on the TF side. According to the documentation mod(x,y) = x - floor(x / y) * y.
If we break that down: mod(0.6, -0.1) = 0.6 - floor(0.6/-0.1) * (-0.1) = 0.6 - floor(-6)*(-0.1) = 0.6 -(-6)*(-0.1) = 0.6 +6*(-0.1) = 0.6 - 0.6 = 0 which is what WebGL gives.

tldr; I'd go with the built-in mod() function in glsl.

ManrajGrover:
Hi @dsmilkov, agreed. I had done some digging in Tensorflow code and found the [implementation](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/kernels/binary_ops.cc#L87) checks the sign of params before calculating mod.

On reading more, I found that Tensorflow does this to keep it consistent with `tf.floordiv` which follows C semantics for negative numbers. I ran a [small code](https://ideone.com/Y1vAMn) for it and found the result to be matching.

dsmilkov:
Thanks Manraj for the deep dive. Let's go with the simpler/faster built-in `mod` and I'm willing to be inconsistent with TF CPU for now, since the condition where we don't match is not very realistic. Also I'm curious how TF CUDA does it and whether that's consistent with TF CPU.

Anyhow, only 2 comments left:
1) fixing the gradient for broadcasting
2) adding a test for gradient with broadcast.

Thank you!

---

Reviewed 8 of 8 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-:-L9Al2nm14Hky2lwrBKw:bbtpn3m)*

dsmilkov:
:lgtm_strong: Thank you for the patience and all the great detective work investigating fmod. Really nicely done.

---

Reviewed 2 of 2 files at r4.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/887#-:-L9QMYGy9AX5avCDitsi:b-h1raxu)*

PR #926, opened 2018-04-06
Make core tests reusable

dsmilkov:
The goal of this PR is to make the core tests reusable across different backends.

- `describeWithFlags` instead of list of features, now takes constraints that need to be satisfied in order for the test to run.
- `ALL_ENVS` becomes `{}`, which means there are no constraints.
- `WEBGL_ENVS` becomes `{BACKEND: 'webgl'}` which means run this test only when the `BACKEND` environment feature is `webgl`. Likewise for `CPU_ENVS`.
- Adds the following user-facing API:
  - `tf.test_util.setBeforeAll`/`setBeforeEach(f)`
  - `tf.test_util.setAfterAll`/`setAfterEach(f)`
  - `tf.test_util.setTestEnvFeatures(listOfFeatures)`. An example list: `[{BACKEND: 'custom'}]`
- Constraints a few core tests to run only in WebGL, since they are using browser-specific API.
- `ENV.registerBackend()` now takes an optional priority number (defaults to 1), which is used by `getBestBackendType` to find the best backend.
- Splits `toPixels` tests into browser-specific (which use canvas) and non-browser specific

Also fixes a bug with vertex buffer attribute binding, which allows us to re-create webgl backends in beforeAll, instead of beforeEach, speeding up our tests by 3x.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/926)

### DIFF
--- a/src/debug_mode_test.ts
+++ b/src/debug_mode_test.ts
@@ -22,11 +22,11 @@ import * as util from './util';
 
 describeWithFlags('debug on', ALL_ENVS, () => {
   beforeAll(() => {
-    dl.ENV.set('DEBUG', true);
+    tf.ENV.set('DEBUG', true);
   });
 
   afterAll(() => {
-    dl.ENV.set('DEBUG', false);
+    tf.ENV.set('DEBUG', false);
   });
 
   it('debug mode does not error when no nans', () => {
@@ -66,7 +66,7 @@ describeWithFlags('debug on', ALL_ENVS, () => {
 
 describeWithFlags('debug off', ALL_ENVS, () => {
   beforeAll(() => {
-    dl.ENV.set('DEBUG', false);
+    tf.ENV.set('DEBUG', false);
   });
 
   it('no errors where there are nans, and debug mode is disabled', () => {
--- a/src/debug_mode_test.ts
+++ b/src/debug_mode_test.ts
@@ -20,11 +20,15 @@ import * as dl from './index';
 import {ALL_ENVS, describeWithFlags, expectArraysClose} from './test_util';
 import * as util from './util';
 
-const ALL_ENVS_DEBUG = ALL_ENVS.map(env => Object.assign({'DEBUG': true}, env));
-const ALL_ENVS_NO_DEBUG =
-    ALL_ENVS.map(env => Object.assign({'DEBUG': false}, env));
+describeWithFlags('debug on', ALL_ENVS, () => {
+  beforeAll(() => {
+    dl.ENV.set('DEBUG', true);
+  });
 
-describeWithFlags('debug on', ALL_ENVS_DEBUG, () => {
+  afterAll(() => {
+    dl.ENV.set('DEBUG', false);
+  });
+
   it('debug mode does not error when no nans', () => {
     const a = dl.tensor1d([2, -1, 0, 3]);
     const res = dl.relu(a);
@@ -60,7 +64,11 @@ describeWithFlags('debug on', ALL_ENVS_DEBUG, () => {
   });
 });
 
-describeWithFlags('debug off', ALL_ENVS_NO_DEBUG, () => {
+describeWithFlags('debug off', ALL_ENVS, () => {
+  beforeAll(() => {
+    dl.ENV.set('DEBUG', false);
+  });
+
   it('no errors where there are nans, and debug mode is disabled', () => {
     const a = dl.tensor1d([2, NaN]);
     const res = dl.relu(a);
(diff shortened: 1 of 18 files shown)


dsmilkov:
Reviewed 2 of 13 files at r2.
Review status: 2 of 13 files reviewed at latest revision, all discussions resolved.

---

*[src/engine.ts, line 288 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-L9rAVuP1p84qizf8IJh:-L9rAVuP1p84qizf8IJi:bbima9e) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9f439d950337f2a083b372c598ffd06ce8ff8313/src/engine.ts#L288)):*
> ```XML
>   }
> 
>   dispose() {}
> ```

engine should not dispose the backend, since it doesn't own it - it just uses it. Also, we re-create an engine between each unit test, but we don't necessarily want to re-create the backend.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926)*

nsthorat:
:lgtm_strong:

---

Reviewed 9 of 13 files at r2.
Review status: 10 of 13 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/environment.ts, line 223 at r3](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-L9uFn5m8IxRfpa1foRi:-L9uFn5m8IxRfpa1foRj:b6qzi1b) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4aceba454299547f3668a708ab9bed2a62e0b31d/src/environment.ts#L223)):*
> ```XML
>    */
>   @doc({heading: 'Environment'})
>   static setBackend(backendType: string, safeMode = false) {
> ```

I dont have a good answer right now but it's quite nice that the union string shows up in our docs:
https://js.tensorflow.org/api/0.9.0/#setBackend

Maybe not a huge deal and we can just add some documentation at some point.

Just something to think about.

---

*[src/environment.ts, line 366 at r3](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-L9uF4Sr7hZMXJTdwkc9:-L9uF4Sr7hZMXJTdwkcA:b1eqs4a) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4aceba454299547f3668a708ab9bed2a62e0b31d/src/environment.ts#L366)):*
> ```XML
>       boolean {
>     if (name in this.registry) {
>       throw new Error(`${name} backend was already registered`);
> ```

QQ - if 2 backends are registered with the same name, this often means we have a double bundle and then the page is fatally broken. What do you think about warning instead of throwing here so not to break the page?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-:-L9uGUx_E2mIWS00jANH:b-jbsf8z)*

dsmilkov:
Reviewed 12 of 13 files at r2.
Review status: 10 of 16 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/environment.ts, line 223 at r3](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-L9uFn5m8IxRfpa1foRi:-L9vERSF2QVZSXsfsWSE:b-ucqnnm) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4aceba454299547f3668a708ab9bed2a62e0b31d/src/environment.ts#L223)):*

I dont have a good answer right now but it's quite nice that the union string shows up in our docs:
https://js.tensorflow.org/api/0.9.0/#setBackend

Maybe not a huge deal and we can just add some documentation at some point.

Just something to think about.
</blockquote></details>

Yeah, we can add more documentation in the @param desc, when nodejs comes out. And a link to how to enable it.

---

*[src/environment.ts, line 366 at r3](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926#-L9uF4Sr7hZMXJTdwkc9:-L9vEmDx0-Tg6cquAYbU:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4aceba454299547f3668a708ab9bed2a62e0b31d/src/environment.ts#L366)):*

QQ - if 2 backends are registered with the same name, this often means we have a double bundle and then the page is fatally broken. What do you think about warning instead of throwing here so not to break the page?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/926)*

PR #938, opened 2018-04-10
Add support for a tensor array of size 1 in concat.

adarob:
Fixes tensorflow/tfjs#73

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/938)


bileschi:
Should an empty list of tensors (length zero) be allowed?

PR #945, opened 2018-04-12
[WIP] batchNorm gradients

easadler:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/945)

### DIFF
--- a/src/ops/batchnorm.ts
+++ b/src/ops/batchnorm.ts
@@ -21,6 +21,7 @@ import {Tensor, Tensor1D, Tensor2D, Tensor3D, Tensor4D} from '../tensor';
 import {Rank} from '../types';
 import * as util from '../util';
 
+import * as ops from './ops';
 import {operation} from './operation';
 
 export class BatchNormOps {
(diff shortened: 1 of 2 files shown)


easadler:
@nsthorat hey just got started. Wanted to get some advice from you before diving in.

I would love it if you could help walk me through the gradients portion a little bit. I'm confused on how to calculate them.

nsthorat:

---

*[src/ops/batchnorm.ts, line 197 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/945#-L9uuQmH9WnWtJ0Elu6K:-L9uuQmH9WnWtJ0Elu6L:b-byldxr) ([raw file](https://github.com/tensorflow/tfjs-core/blob/18ad3fa346dd3d21becd58fde4c8ada310c6d1a8/src/ops/batchnorm.ts#L197)):*
> ```XML
> 
>     const grad = (dy: Tensor4D) => {
>       return {x: () => dy, mean: () => dy, variance: () => dy};
> ```

Alright, so here you want to return a value for all of the tensor inputs (x, mean, variance, scale, offset).

Each of those values will be the partial derivative of the output with respect to that variable, times dy (dy is actually de/dy).

For example, `de/dx = de/dy * dy/dx`

So your job here now is to complete dy / d(variable) for each of those inputs.

Check out the equation for batchnorm: https://www.tensorflow.org/api_docs/python/tf/nn/batch_normalization

Now you just need to compute all of the partial derivatives.

---

*[src/ops/batchnorm.ts, line 205 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/945#-L9uvAKF9mWyI19Wp_VS:-L9uvAKF9mWyI19Wp_VT:b-l0u9u0) ([raw file](https://github.com/tensorflow/tfjs-core/blob/18ad3fa346dd3d21becd58fde4c8ada310c6d1a8/src/ops/batchnorm.ts#L205)):*
> ```XML
>             varianceEpsilon, batchnormReshape4D(scale),
>             batchnormReshape4D(offset)),
>         {x: x4D, mean, variance}, grad);
> ```

here you will have to also specify the other inputs so we can compute derivatives.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/945)*

nsthorat:
Added a comment, hope this helps.

easadler:
@nsthorat thanks that was super helpful! I will copy tensorflow!  I see that beta and gamma are offset and scale now. Sometimes reading the documentation is important!

Thanks again!

easadler:
@nsthorat hey just made some progress. I've got a couple questions if you don't mind helping. I will give finishing this up another stab in a day or so.

- I've been using the batch norm [paper](https://arxiv.org/pdf/1502.03167.pdf) for the partial derivatives. These are consistent with tensorflow until I get to `derMean`. When I use `tf.gradients` I get a different result, but when I redo the equation from scratch in tensorflow, it lines up with the results in this PR. I could just be implementing it wrong. Maybe you see something or this formula is incorrect?
- I'm having trouble reshaping the gradient for x. This could be because of issues with my formulas, but perhaps there is something else that I am missing.

nsthorat:
Hey @easadler, if the current tests work for the subset of gradients, let's merge this as is, there's actually an internal person who needs batchNorm who can finish your work up! How does that sound to you? Would you just leave a comment saying which derivative isn't yet working?

To respond to your questions directly, I usually check out wolfram alpha to compute non-trivial partial derivatives, you could try that for the gradients you are trying to compute. Going forward you should try that, it's really helpful!

caisq:
Thanks @easadler for starting this!

easadler:
Thanks guys :).

PR #951, opened 2018-04-14
Add erf ops

Lewuathe:
# Purpose

[As TensorFlow does](https://www.tensorflow.org/api_docs/python/tf/erf), TensorFlow.js can provide `erf` ops to calculate error function. The actual algorithm calculates the value approximately with elementary functions. See [Approximation with elementary functions](https://en.wikipedia.org/wiki/Error_function#Approximation_with_elementary_functions).

screen shot 2018-04-14 at 19 19 49

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/951)


nsthorat:
Hey @Lewuathe, just curious, do you need these ops / kernels for something?

In general, we're now at a point where we're going to try to stay use-case oriented instead of trying to port every exotic op that TensorFlow just for parity, it's going to make it harder for us to implement other backends and the bundle size is also starting to get bloated. If there is some use case you need this for we'll definitely get it in.

Lewuathe:
@nsthorat I see. Thank you for a feedback. Actually, it is not use case driven. I thought we might need to port ops in TF (python) as many as possible. That's the main reason.
So if there is no use case for `erf` generally, we can close this.

nsthorat:
Thanks for understanding!

dsmilkov:
:lgtm_strong: Since you've already done all this work, we will get this in. Left couple small comments. This is already looking really good. Nice work @lewuathe . Glad to have you back!!

---

Reviewed 5 of 9 files at r1.
Review status: 5 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/ops/unary_ops.ts, line 746 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/951#-LAFYnwlDZRmVEcoXm4D:-LAFYnwlDZRmVEcoXm4E:b-r6809x) ([raw file](https://github.com/tensorflow/tfjs-core/blob/476a3fac1edea13cbcb847052f54f7a17b977389/src/ops/unary_ops.ts#L746)):*
> ```XML
>   @doc({heading: 'Operations', subheading: 'Basic math'})
>   @operation
>   static erf<T extends Tensor>(x: T): T {
> ```

error if x is not of dtype 'int32 or 'float' and cast to float32 before calling the backend.

```ts
util.assert(x.dtype === 'int32' || x.dtype === 'float32',
    'Input dtype must be `int32` or `float32`');
if (x.dtype === 'int32') {
  x = x.toFloat();
}
```

---

*[src/ops/unary_ops_test.ts, line 2058 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/951#-LAFZz24EWFHeftYkr36:-LAFZz24EWFHeftYkr37:b-sccsxr) ([raw file](https://github.com/tensorflow/tfjs-core/blob/476a3fac1edea13cbcb847052f54f7a17b977389/src/ops/unary_ops_test.ts#L2058)):*
> ```XML
>     const a = tf.tensor2d(values, [2, 2]);
>     const result = tf.erf(a);
>     const expected = [0.2227, 0.3286, 0.4283, 0.5204];
> ```

The actual `expectArraysClose` won't test for more than `1e-3`, but we are going to be reducing that threshold as we improve numerical precision. To get us ready for that, can you add more decimal places in the expected results for these tests. For e.g. I just used http://colab.research.google.com/ to test this input and got:
```
[0.2227026 , 0.32862678, 0.42839235, 0.5204999 ]
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/951#-:-LAF_ZhrC3mFPzsZtFxt:bpoek6g)*

dsmilkov:
:lgtm_strong: Thanks!

---

Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/951#-:-LAW7QNTBNUGbdSe1BeX:bfdlkw5)*

PR #957, opened 2018-04-15
Finish gradient for tf.matMul

jgartman:
This PR finishes the implementation of the gradient for `tf.matMul`. https://github.com/tensorflow/tfjs/issues/108

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/957)


dsmilkov:
Thank you! One suggestion!

---

Reviewed 2 of 2 files at r1.

---

*[src/ops/matmul.ts, line 65 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957#-LADsQYMCeSonH12EmNJ:-LADsQYMCeSonH12EmNK:b-bzlln) ([raw file](https://github.com/tensorflow/tfjs-core/blob/0b9e8739afb10e302f1b5130ad0ece1112110b52/src/ops/matmul.ts#L65)):*
> ```XML
>           b: () => a.toFloat().matMul(dy, true, false)
>         };
>       } else if (!transposeA && transposeB) {
> ```

looks like you can have only one case instead of 4 cases, using the `transposeA` and `transposeB` flipped in the right now: (please double check me on this):

```ts
a: () => dy.matMul(b.toFloat(), transposeA, !transposeB),
b: () => a.toFloat().matMul(dy, !transposeA, transposeB);
```

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957#-:-LADuMVb3ZFCD7WU2qH9:b-chro68)*

jgartman:

---

*[src/ops/matmul.ts, line 65 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957#-LADsQYMCeSonH12EmNJ:-LAG97Fl3AnPSFO7YaJW:b-u4dnsi) ([raw file](https://github.com/tensorflow/tfjs-core/blob/0b9e8739afb10e302f1b5130ad0ece1112110b52/src/ops/matmul.ts#L65)):*

looks like you can have only one case instead of 4 cases, using the `transposeA` and `transposeB` flipped in the right now: (please double check me on this):

```ts
a: () => dy.matMul(b.toFloat(), transposeA, !transposeB),
b: () => a.toFloat().matMul(dy, !transposeA, transposeB);
```

</blockquote></details>

Thanks, I might be misunderstanding but I'm not sure that would work.  For example, in the case of `!transposeA && transposeB`suppose `a` has shape `[3 2]` and `b` has shape `[1 2]`.  When we take the product `ab^T` the resulting matrix will have shape `[3 1]` so the gradient `dy` should also have shape `[3 1]`.  If we calculate `db` by `a^Tdy^T` then we would be trying to multiply matrices with shapes `[2 3][1 3]`.  Something similar would also happen in the `transposeA && !transposeB` case.  Let me know if I'm missing something.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957)*

dsmilkov:

---

*[src/ops/matmul.ts, line 65 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957#-LADsQYMCeSonH12EmNJ:-LAGEL_tEfxlPcVyv-kV:bflo3dy) ([raw file](https://github.com/tensorflow/tfjs-core/blob/0b9e8739afb10e302f1b5130ad0ece1112110b52/src/ops/matmul.ts#L65)):*
<details><summary><i>Previously, jgartman (Josh Gartman) wrote…</i></summary><blockquote>

Thanks, I might be misunderstanding but I'm not sure that would work.  For example, in the case of `!transposeA && transposeB`suppose `a` has shape `[3 2]` and `b` has shape `[1 2]`.  When we take the product `ab^T` the resulting matrix will have shape `[3 1]` so the gradient `dy` should also have shape `[3 1]`.  If we calculate `db` by `a^Tdy^T` then we would be trying to multiply matrices with shapes `[2 3][1 3]`.  Something similar would also happen in the `transposeA && !transposeB` case.  Let me know if I'm missing something.
</blockquote></details>

Ah you are completely right! I double checked. Thanks for being super careful!

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957)*

dsmilkov:
:lgtm_strong: Really nice work. Thank you Josh!

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/957#-:-LAGEUaB2-V4-2l51dnZ:bcy7xho)*

PR #958, opened 2018-04-16
Preserve license header in index.js

Lewuathe:
Preserve license header in the main JS file (`dist/index.js`) by using `/*!`.

see: https://github.com/tensorflow/tfjs/issues/66

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/958)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-/**
+/*!
  * @license
  * Copyright 2017 Google Inc. All Rights Reserved.
  * Licensed under the Apache License, Version 2.0 (the "License");


nsthorat:
@dsmilkov can you look at this? This will only add the license to index.js, do you understand the scope of the issue you filed?

nsthorat:
Just chatted with @dsmilkov, I think we should drop this since it doesn't add much. Thanks a lot for the contribution though.

Lewuathe:
@nsthorat I see. Thanks for feedback. I'll close this.

PR #960, opened 2018-04-16
Implement floordiv for integer division.

nsthorat:
Also adds a note for contributing.

Fixes https://github.com/tensorflow/tfjs/issues/136

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/960)


src/ops/arithmetic_test.ts
```diff
@@ -29,6 +29,26 @@ describeWithFlags('div', ALL_ENVS, () => {
...
   });
 
+  it('integer division implements floor divide', () => {
+    const a = tf.tensor1d([-5, -4, 3, 2], 'int32');
+    const c = tf.tensor1d([3, 2, 2, 2], 'int32');
+
+    const r = tf.div(a, c);
+
+    expect(r.dtype).toEqual('int32');
+    expectArraysClose(r, [-2, -2, 1, 1]);
```
oveddan: I believe that this is actually rounding. 
Shouldn't the first value here, -2, be -1?  `Math.floor(-5/3)` should be -1.

oveddan: 👍 
One more issue, if you do tf.div with 3/3, you get a 0:

```javascript
    const a = tf.scalar(3, 'int32');
    const c = tf.scalar(3, 'int32');

    const r = tf.div(a, c);

    expect(r.dtype).toEqual('int32');
    expectArraysClose(r, [1]);
```
results in an error:
	Error: Arrays differ: actual[0] = 0, expected[0] = 1.

PR #972, opened 2018-04-20
Move @types/jasmine to dependencies.

adarob:
@types/jasmine is required if an @tensorflow/tfjs-core is used directly since test_util is in the index.

https://github.com/tensorflow/magenta-js/pull/38
@iansimon

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/972)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "@types/jasmine": "~2.5.53",
     "@types/node": "~9.6.0",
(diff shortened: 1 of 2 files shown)


ry:
I hit this issue too, but I jasmine is a rather larger dependency (123kb min, https://bundlephobia.com/result?p=jasmine@3.1.0) that is orthogonal to TFJS. 
Ideally only the tests would import test_util...

adarob:
Thanks!

PR #976, opened 2018-04-22
Add rank, size ops

Lewuathe:
# Purpose

As described [here](https://github.com/tensorflow/tfjs/issues/188#issuecomment-381845538), MobileNet V2 depends on the rank and size as an ops implementation. Like TensorFlow, they can be returned as rank0 tensor (scalar). 

# Note
We may need to consider consistency with `rank` in eager mode because it returns a rank as just a number not a tensor.

see: https://github.com/tensorflow/tfjs/issues/188#issuecomment-381845538

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/976)

### DIFF
--- a/src/ops/array_ops.ts
+++ b/src/ops/array_ops.ts
@@ -1232,6 +1232,25 @@ export class ArrayOps {
   static print<T extends Tensor>(x: T, verbose = false): void {
     console.log(tensor_util.tensorToString(x, verbose));
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


Lewuathe:
Ah, I see. So we don't need `rank`, `size` ops if we can convert `SavedModel`. Thanks.
I'm going to close this.

PR #978, opened 2018-04-22
Improve precision for Safari by using float16 on iOS and encoding floats as bytes before downloading.

nsthorat:
This PR enables iOS Safari and Desktop Safari to have higher floating point precision (vs now which is fixed point precision). This means we also support numbers larger than 20,000 (and less than -20,000). To do this, we must use different types of textures for uploading, rendering, and downloading.

On all platforms, uploading to float32 textures is supported.

|         | Render           | Download  |
| ------------- |:-------------:| -----:|
| Chrome / Firefox      | float32 | float32 |
| Safari Desktop      | float32      |   bytes |
| iOS Safari | float16      |    bytes |

To do this, we remove the `WEBGL_FLOAT_TEXTURE_ENABLED` bit and introduce two new bits, `WEBGL_RENDER_FLOAT32_ENABLED` and `WEBGL_DOWNLOAD_FLOAT_ENABLED` which enable each of the respective code paths depending on the device.

Because we have multiple types of textures, the texture manager now knows about the type of textures, and will cache based on the texture type. I introduced a separation of `LogicalTextureType` which includes "upload", "render" and "download", so from the perspective of the WebGL backend we ask the texture manager for a texture of this type. The texture manager will convert that to a `PhysicalTextureType` depending on environment flags, and create a texture based on that type. This means we maximize cache utilization across flag settings.

Because we can't download floating point textures directly, we have to first pass it through an IEEE 32 bit float encoder program which samples from a floating point texture and writes to an unsigned byte texture. In JavaScript, we simply wrap that `ArrayBuffer` in a `Float32Array` to decode on the CPU-side. This only happens when the user asks to download a Tensor from the GPU (via readSync in BackendWebGL).

This PR also:
- Turns off Firefox testing locally. This speeds up testing for local development.
- On Browserstack, we now run Chrome, Firefox, Safari, and iOS Safari.
- For iOS devices (render float16) we reduce the precision in unit tests to 1e-1. Late, we can see if we can improve precision.
- Turns off a few tests for non-Chrome browsers. I filed issues to track turning them back on (all IO handler related).
- Refactors gpgpu_context so that the texture configuration is computed a single time when the context is created, instead of computing them whenever needed. This cleans up the file a lot.
- Removes all the code paths for packing.
- Swaps `isNaN` for a new implementation that works on all platforms (tested on windows edge).
- Reduces the size of some values in unit tests to pass in 1e-1 setting on iOS.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/978)


dsmilkov:
The change is ok

caisq:
Just a general question: does this change have any implications for performance on non-mobile browsers?

---

---

*[src/environment.ts, line 154 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymBuZEw4RkMH7aWO2:-LEymBuZEw4RkMH7aWO3:b-g4v9jp) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/environment.ts#L154)):*
> ```
> b
> ```

Just curious about the naming: maybe FrameBuffer instead of Framebuffer? 

---

*[src/io/browser_http_test.ts, line 273 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymmfL83YI3Tb1843O:-LEymmfL83YI3Tb1843P:bg0ls1x) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/io/browser_http_test.ts#L273)):*
> ```
> 'existing body'
> ```

Since this test seems to run on Chrome only now, why do we need to remove FormData here?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-:-LEyn3P4E6EAvrDG9Q_C:b-lqygxb)*

nsthorat:

---

*[src/environment.ts, line 154 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymBuZEw4RkMH7aWO2:-LEynvVlC5Y4LHiF2Nu0:b1hpovz) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/environment.ts#L154)):*

> ```
> b
> ```

Just curious about the naming: maybe FrameBuffer instead of Framebuffer? 
</blockquote></details>

Framebuffer is actually lower case (see the gl.createFramebuffer native call below).

---

*[src/io/browser_http_test.ts, line 273 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymmfL83YI3Tb1843O:-LEynyN39s7K3zYrO7iF:b-n3zogd) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/io/browser_http_test.ts#L273)):*

> ```
> 'existing body'
> ```

Since this test seems to run on Chrome only now, why do we need to remove FormData here?
</blockquote></details>

Well, I did part of the work for you -- when you turn back on Safari this would fail. :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/978)*

dsmilkov:
:lgtm_strong: This is AMAZING!!!!

---

Reviewed 8 of 27 files at r1, 20 of 23 files at r3, 5 of 10 files at r4.

---

*[src/jasmine_util.ts, line 172 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEyju0t7ks6hNJGa3ac:-LEyju0u1ls_SDvXhBCD:b6i1dgi) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/jasmine_util.ts#L172)):*
> ```XML
> }
> 
> export function expectArrayBuffersEqual(
> ```

move to `test_util.ts`?

---

*[src/kernels/backend_webgl.ts, line 141 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymEeU5sGf2WC-VuPA:-LEymEeV7e0OLpTTnrAP:b-6b8ce4) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/kernels/backend_webgl.ts#L141)):*
> ```XML
>       texture: null,
>       texShape: null,
>       texType: LogicalTextureType.FLOAT_RENDER
> ```

perhaps rename to `usage: TextureUsage`

---

*[src/kernels/webgl/gpgpu_util.ts, line 47 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEynNzb6e-1VskjLe0X:-LEynNzc6yPSNXXqpktV:bf76dbt) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/kernels/webgl/gpgpu_util.ts#L47)):*
> ```XML
> 
>   defaultNumChannels: number;
>   textureTypeHalfFloat: number;
> ```

much cleaner!

---

*[src/kernels/webgl/tex_util.ts, line 23 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEylWUH2Nfw8EqW_btb:-LEylHafBN6qfnLrMWEl:b-rfbx71) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/kernels/webgl/tex_util.ts#L23)):*
> ```XML
>   FLOAT_RENDER,
>   FLOAT_UPLOAD,
>   UNSIGNED_BYTE
> ```

Add PIXELS and FLOAT_DOWNLOAD as logical (remove unsigned_byte)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-:-LEyoGok0gpuytTT9UHZ:b-g5tiy2)*

nsthorat:

---

*[src/environment.ts, line 154 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymBuZEw4RkMH7aWO2:-LEyp99NEI9zMenq4_81:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/environment.ts#L154)):*

Framebuffer is actually lower case (see the gl.createFramebuffer native call below).
</blockquote></details>

Done.

---

*[src/io/browser_http_test.ts, line 273 at r4](https://reviewable.io/reviews/tensorflow/tfjs-core/978#-LEymmfL83YI3Tb1843O:-LEyp8ok7nsEYNkX3Fdd:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/ae5a9461e1d6aefae496354fda814609285eb6fd/src/io/browser_http_test.ts#L273)):*

Well, I did part of the work for you -- when you turn back on Safari this would fail. :)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/978)*

lygstate:
Question, on some GPU, such as Mali T720, that does support OpenGL ES 3.1, but doesn't support for EXT_color_buffer_float, would this be able to upload float32 texture? And downloading like iOS does?

PR #980, opened 2018-04-22
Stop recording to the tape while a kernel runs

dsmilkov:
Some kernels might call higher-level ops to work around limitations. Because of this, we need to stop recording to the tape while the kernel runs.

Also use built-in inversesqrt in GLSL

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/980)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -103,16 +103,18 @@ export class Engine implements TensorManager {
     };
     const scopeName = this.activeScope.name;
 
+    // Stop recording to a tape when running a kernel.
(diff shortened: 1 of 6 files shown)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-core/980#-:-LAifr7Y0wpBdaoV6NnR:b-jbsf8z)*

PR #992, opened 2018-04-25
add tests for tf.conv1d gradients

easadler:
This PR adds tests to make sure that the gradients for`tf.convo1d` match tensorflow. I believe this has to do with tensorflow/tfjs#108. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/992)


dsmilkov:

---

*[src/ops/conv.ts, line 111 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/992#-LAxMDlbAnfVz8G1m7Bz:-LAxMDlbAnfVz8G1m7C-:b-58i63) ([raw file](https://github.com/tensorflow/tfjs-core/blob/fa39d7763953d523934f989a53cdb07149f37634/src/ops/conv.ts#L111)):*
> ```XML
>                    .as3D(filter.shape[0], filter.shape[1], filter.shape[2])
>       };
>     };
> ```

Since conv1d calls conv2d, both are high-level ops, and conv2d has already gradients, we don't need to explicitly implement gradients for conv1d - it will happen automatically. What is veru useful, and you already have started on, is to write bunch of unit tests to make sure the gradient works and compare it with TF in python (http://colab.research.google.com is very useful for testing TF Python) Thanks!

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/992)*

dsmilkov:
Your tests found a bug in our CPU backend, which we just fixed in #1014. I will sync this PR to master, and your tests should pass, at which point I'll merge this in. Thank you!

dsmilkov:
:lgtm_strong:

---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/992#-:-LBaTW6_1FmWIcnIr6mx:b-jbsf8z)*

PR #999, opened 2018-04-26
Optimize encodeFloatArray() and decodeToFloatArray() to be 2x faster

janhuenermann:
Hi,

I reduced function calls (map etc.) in `encodeFloatArray()` and `decodeToFloatArray()` which leads to a 2x+ improvement. Given the importance of these functions in browsers such as Safari 11, which have no support for float-texture reading, I think this will contribute to overall performance improvements.

-Jan

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/999)

### DIFF
--- a/src/kernels/webgl/tex_util.ts
+++ b/src/kernels/webgl/tex_util.ts
@@ -87,10 +87,10 @@ export function encodeFloatArray(floatArray: Float32Array): Uint8Array {
     value = floatArray[i / 4];
 
     if (isNaN(value)) {
-      uintArray[i] = BYTE_NAN_VALUE
-      uintArray[i + 1] = BYTE_NAN_VALUE
-      uintArray[i + 2] = BYTE_NAN_VALUE
-      uintArray[i + 3] = BYTE_NAN_VALUE
+      uintArray[i] = BYTE_NAN_VALUE;
+      uintArray[i + 1] = BYTE_NAN_VALUE;
+      uintArray[i + 2] = BYTE_NAN_VALUE;
+      uintArray[i + 3] = BYTE_NAN_VALUE;
       continue ;
     }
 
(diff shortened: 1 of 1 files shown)


janhuenermann:
I signed it! :)

nsthorat:
Hi @janhuenermann thanks so much for this. I'm actually about to completely change how Safari / iOS work. In fact, we will not need this fixed point math to do the conversion (since you can render to floats, we can IEEE 754 pack to bytes on the GPU, download bytes into a TypedArray and allow the browser to do conversion from bytes => float).

Basically this method will be going away in a week (and also currently there is no test coverage for this since we turned off unit tests on Safari since precision was so bad).

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/999#-:-LB1ZvUKEjYNLS-2bQ89:bm470ia)*

janhuenermann:
Alright. This is awesome. I was going crazy about my models not converging in Safari.. 😆 And was already wondering why tfjs isn't making use of float texture writing, which is available in Safari (and then converting it to uint8's later on, on the GPU..). Cool then, excited for that release! But if you need any help in that regard, let me know :)

nsthorat:
Awesome!! Thanks so much. I was deep in WebKit code this weekend because it turns out iOS doesn't let you upload half-floats, it only lets you write to half floats, so the config will be:

Safari Desktop: Upload 32 bit floats, render to 32 bit floats, download IEEE 754 floats packed as unsigned bytes.
iOS Safari: Upload 32 bit floats, render to 16 bit floats, download IEEE 754 floats packed as unsigned bytes.

Once I get some time I'm going to finish this up.

dsmilkov:
Closing this PR since we agreed that the code being optimized is going away with #978. Thanks @janhuenermann and looking forward to more collaboration!

PR #1000, opened 2018-04-26
Align tf.resizeNearestNeighbor with TF

Lewuathe:
TensorFlow uses `floor` instead of `round` when `alignCorners` is specified. 

see: https://github.com/tensorflow/tfjs/issues/221


nsthorat:
Thanks for the fix!

PR #1003, opened 2018-04-27
Add IOHandler subtype BrowserLocalStorage.

caisq:
* `tf.io.browserLocalStorage()` is the public factory function.
* Adjust the interface`SaveResult` slightly: Remove the required
  `success` boolean field. Failures in saving and loading will always be
  thrown as `Error`s.
* Add helper functions:
  * `stringByteLength`: required for calculating the byte length of
    the JSON model topology and weight specs.
  * `arrayBufferToBase64String`
  * `base64StringToArrayBuffer`
  The last two helper functions are required because local storage
  values cannot by binary arrays or `ArrayBuffer`s.

Towards tensorflow/tfjs#13

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1003)


src/io/local_storage.ts
```diff
@@ -0,0 +1,227 @@
...
+  constructor(modelPath: string) {
+    if (!(window && window.localStorage)) {
+      throw new Error(
+          'The current environment does not support local storage.');
+    }
+    this.LS = window.localStorage;
+
+    if (modelPath == null) {
+      throw new Error(
+          'For local storage, modelPath must not be null or undefined.');
```
ericdnielsen: I would suggest using the same error string for both this Error and the next one.  We don't people playing whack-a-mole.  Following the discussion last year on TF errors, something like "Local Storage requires a modelPath that is no empty, underfined, nor null."  Maybe add on a "Suggested fix/use: use a unique descriptive name, possibly including a version token such as 'supermodel-01'"

src/io/local_storage.ts
```diff
@@ -0,0 +1,227 @@
...
+   * See the documentation to `browserLocalStorage` for details on the saved
+   * artifacts.
+   *
+   * @param modelArtifacts The model artifacts to be stored.
+   * @returns An instance of SaveResult.
+   */
+  async save(modelArtifacts: ModelArtifacts): Promise<SaveResult> {
+    if (!(window && window.localStorage)) {
+      throw new Error(
+          'The current environment does not support local storage.');
```
ericdnielsen: Do we want an io.io_util.enumerateSupportedHandlers? To be able to provide the list of support handlers in the error message?  In either case can we provide better details "LocalStorage requires a browser that supports XYZ, this program is not being run in a browser or not in a browser that supports that standard."

src/io/local_storage.ts
```diff
@@ -0,0 +1,227 @@
...
+
+        return {modelArtifactsInfo};
+      } catch (err) {
+        // If saving failed, clean up all items saved so far.
+        for (const key in this.paths) {
+          this.LS.removeItem(this.paths[key]);
+        }
+
+        throw new Error(
+            `Failed to save model '${this.modelPath}' to local storage: ` +
```
ericdnielsen: Does err contain any information that we can use to say why it failed?  Do we need to do size checks before we attempt to write?  Are there possible permissions issues we can pre-check?

caisq:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/io/local_storage.ts, line 76 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LB3lMQuEZHXFGqy8t0v-r1-76:-LB71i9iB_yiXaU_VrLt:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/591212200901b67efd71461f5935933dd9db1887/src/io/local_storage.ts#L76)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

I would suggest using the same error string for both this Error and the next one.  We don't people playing whack-a-mole.  Following the discussion last year on TF errors, something like "Local Storage requires a modelPath that is no empty, underfined, nor null."  Maybe add on a "Suggested fix/use: use a unique descriptive name, possibly including a version token such as 'supermodel-01'"
</blockquote></details>

Done.

---

*[src/io/local_storage.ts, line 103 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LB3lMQuEZHXFGqy8t0v-r1-103:-LB70UIc0kmR5dpp64UV:b28wrpr) ([raw file](https://github.com/tensorflow/tfjs-core/blob/591212200901b67efd71461f5935933dd9db1887/src/io/local_storage.ts#L103)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Do we want an io.io_util.enumerateSupportedHandlers? To be able to provide the list of support handlers in the error message?  In either case can we provide better details "LocalStorage requires a browser that supports XYZ, this program is not being run in a browser or not in a browser that supports that standard."
</blockquote></details>

Good point. I added a TODO item here to improve the helpfulness of this error message. Ideas include:
1. Link to a page under js.tensorflow.org
2. When we have the classes and logic ready, automatically determine what the supported IOHandlers are given the current environment and print them out.

---

*[src/io/local_storage.ts, line 138 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LB3lMQuEZHXFGqy8t0v-r1-138:-LB71jJ531e7W7_ExY82:b-nj41zm) ([raw file](https://github.com/tensorflow/tfjs-core/blob/591212200901b67efd71461f5935933dd9db1887/src/io/local_storage.ts#L138)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Does err contain any information that we can use to say why it failed?  Do we need to do size checks before we attempt to write?  Are there possible permissions issues we can pre-check?
</blockquote></details>

There is no API to check the browser local storage quota. Hence there are pages like this that figure this out for you by saving incrementally large artifacts: https://arty.name/localstorage.html

I added a few words here pointing to size quota as a likely cause.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1003)*

nsthorat:
:lgtm_strong:

Nice work!

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/io/io_utils.ts, line 166 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUyex0fMmjw8F1YTW:-LBHUyex0fMmjw8F1YTX:bpsr8xx) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L166)):*
> ```XML
>  * @returns A string that base64-encodes `buffer`.
>  */
> export function arrayBufferToBase64String(buffer: ArrayBuffer) {
> ```

return type

---

*[src/io/io_utils.ts, line 167 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUigCBVUk9mWzZWTi:-LBHUigCBVUk9mWzZWTj:b3ue8lf) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L167)):*
> ```XML
>  */
> export function arrayBufferToBase64String(buffer: ArrayBuffer) {
>   return btoa(String.fromCharCode.apply(null, new Uint8Array(buffer)));
> ```

optional: use spread operator. 

`String.fromCharCode(...new Uint8Array(buffer));`

---

*[src/io/io_utils.ts, line 176 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUpny-pmtvzbzhXC4:-LBHUpny-pmtvzbzhXC5:bpsr8xx) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L176)):*
> ```XML
>  * @returns Decoded `ArrayBuffer`.
>  */
> export function base64StringToArrayBuffer(str: string) {
> ```

return type

---

*[src/io/local_storage.ts, line 138 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LB3lMQuEZHXFGqy8t0v-r1-138:-LBHY9v74WSzwf-6A_nd:booyp3m) ([raw file](https://github.com/tensorflow/tfjs-core/blob/591212200901b67efd71461f5935933dd9db1887/src/io/local_storage.ts#L138)):*

There is no API to check the browser local storage quota. Hence there are pages like this that figure this out for you by saving incrementally large artifacts: https://arty.name/localstorage.html

I added a few words here pointing to size quota as a likely cause.
</blockquote></details>

Can we also log the native err here?

---

*[src/io/local_storage.ts, line 67 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHXOef-8bGw9LDQdiV:-LBHXOef-8bGw9LDQdiW:b3p3bfr) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/local_storage.ts#L67)):*
> ```XML
>   protected readonly paths: {[key: string]: string};
> 
>   constructor(modelPath: string) {
> ```

s/modelPath/modelName? you refer to this as name in error messages

---

*[src/io/local_storage.ts, line 116 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHXzE04EO_MisGO6In:-LBHXzE04EO_MisGO6Io:brvqk20) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/local_storage.ts#L116)):*
> ```XML
>       const modelArtifactsInfo: ModelArtifactsInfo = {
>         dateSaved: new Date(),
>         modelTopologyType: 'KerasJSON',
> ```

how will this get wired through? will this be a field in ModelArtifacts?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-:-LBHYmRO8XJU2HoLTChi:b-rtmsva)*

caisq:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/io/io_utils.ts, line 166 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUyex0fMmjw8F1YTW:-LBJXYZd62laDdlTYE3x:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L166)):*

return type
</blockquote></details>

Done.

---

*[src/io/io_utils.ts, line 167 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUigCBVUk9mWzZWTi:-LBJXn-KDOWwZimOHWYs:b-5h45v3) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L167)):*

optional: use spread operator. 

`String.fromCharCode(...new Uint8Array(buffer));`
</blockquote></details>

Hmm. The results seem different

```js
btoa(String.fromCharCode.apply(null, new Uint8Array([1,2,3])))
// vs.
btoa(...new Uint8Array([1,2,3]))
```

Let's keep it this way.

---

*[src/io/io_utils.ts, line 176 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUpny-pmtvzbzhXC4:-LBJYAjCC9t_y5sgSgQP:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L176)):*

return type
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1003)*

nsthorat:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[src/io/io_utils.ts, line 167 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1003#-LBHUigCBVUk9mWzZWTi:-LBLVmPSEbUgInL9arxE:btmzvx6) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b56527f4053f40be770b7d74f1a43f128b608421/src/io/io_utils.ts#L167)):*

Hmm. The results seem different

```js
btoa(String.fromCharCode.apply(null, new Uint8Array([1,2,3])))
// vs.
btoa(...new Uint8Array([1,2,3]))
```

Let's keep it this way.
</blockquote></details>

Just for posterity (no need to update, just good for JS foo), the correct line is:

`btoa(String.fromCharCode(...new Uint8Array(buffer)));`

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1003)*

PR #1010, opened 2018-04-30
Distribution Normal from Tensors

thibo73800:
This pull request is a first implementation of the class available in tensorflow python:
tf.distributions.Normal
The current method on tensorflow.js output a non-differentiable tensorflow. 

This pull request implements a differentiable normal distribution from tensors (mean and stds).

The original issue: [https://github.com/tensorflow/tfjs/issues/231#issuecomment-385482092](https://github.com/tensorflow/tfjs/issues/231#issuecomment-385482092)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1010)


dsmilkov:
Thanks for this awesome contribution!

Instead of starting a new namespace to match `tf.distributions`, it would be better to just implement gradients for `mean` and `std` in [`tf.randomNormal`](https://js.tensorflow.org/api/0.10.0/#randomNormal) to match TensorFlow's [`tf.random_normal`](https://www.tensorflow.org/api_docs/python/tf/random_normal). To do that, allow `mean` and `std` to be either a number or a Tensor. Does this sound ok?

Cheers!

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/ops/normal.ts, line 40 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-LBb-5v_1vhd5L_Yt1AY:-LBb-5v_1vhd5L_Yt1AZ:b-k2n4ir) ([raw file](https://github.com/tensorflow/tfjs-core/blob/3f6fa5fb74dd1b101947688cbed884cc88b49077/src/ops/normal.ts#L40)):*
> ```XML
>    * @param scale Floating point tensor|number; the stddevs of the
>    * distribution(s).
>    * @param dtype The data type of the output.
> ```

Remove dtype since tf.distributions.Normal doesn't have it and ints also don't make much sense here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-:-LBb-Vj6EsP-3oDq6Hhm:b1k0whr)*

dsmilkov:
We want to postpone growing the codebase too quicky until there is a good use-case driven request. In other words, we can start `tf.distributions.*` when there is a real use-case that can't benefit from `tf.randomUniform` and `tf.randomNormal`.

---

Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-:-LBb1Ib64hGZ3RbYoKI1:b-48f840)*

thibo73800:
Ok, I just finish to implement gradients for mean and std and add unit tests. The normal behavior of the method is still working. All the tests passed on my computer (yarn lint and yarn test) but travis through this error after all the tests passed:

`The command "if [ "$BROWSERSTACK_KEY" ]; then yarn test-travis; fi" exited with 1.`

I am not sure to understand where does the problem comes from...

thibo73800:
@dsmilkov  Travis is now happy :)

dsmilkov:
Hi @thibo73800 I don't see any gradients being implemented. We can still get this in after you simplify the method to take only scalar (see comment below). Also if you can update the title of the PR and the top-level comment to reflect what this PR will do. 

---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/ops/array_ops.ts, line 428 at r3](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-LCG0i5h8um0pEIejaar:-LCG0i5h8um0pEIejaas:bb2wrd7) ([raw file](https://github.com/tensorflow/tfjs-core/blob/21d3cab23124bce0b26c8ab197d579fcc5c560ce/src/ops/array_ops.ts#L428)):*
> ```XML
>    * const means = tf.tensor1d([0., 5., 10.]);
>    * const stds = tf.tensor1d([1., 1., 1.]);
>    * tf.randomNormal([2, 2], means, stds).print();
> ```

to align with TF's API: https://www.tensorflow.org/api_docs/python/tf/random_normal let's force the mean and std to be either a number of a Scalar (0-rank tensor).

---

*[src/ops/array_ops_test.ts, line 759 at r3](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-LCG0-p90cn9zu_sJd9X:-LCG0-p90cn9zu_sJd9Y:btbd0nz) ([raw file](https://github.com/tensorflow/tfjs-core/blob/21d3cab23124bce0b26c8ab197d579fcc5c560ce/src/ops/array_ops_test.ts#L759)):*
> ```XML
>   });
> 
>   it('should return a int32 2D of values from differents normal distribution',
> ```

i don't think this description reflects what is tests below. Also no need to test for int32 since randomNormal of int32 doesn't make sense.

---

*[src/ops/array_ops_test.ts, line 796 at r3](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-LCG0GVg6e71VpH3vB5O:-LCG0GVg6e71VpH3vB5P:bihetsh) ([raw file](https://github.com/tensorflow/tfjs-core/blob/21d3cab23124bce0b26c8ab197d579fcc5c560ce/src/ops/array_ops_test.ts#L796)):*
> ```XML
>   });
> 
>   it('should return a float32 3D of values from differents normal distribution',
> ```

it's not really a 3d tensor tested below.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1010#-:-LCG0pX9-q1CcROah7QB:b2z2n10)*

dsmilkov:
Hi @thibo73800 ,  I just noticed that TensorFlow's tf.random_normal [is not differentiable](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/random_ops.py#L83). That is, you can't compute the gradient of the output of random_normal w.r.t `mean` or `std_dev`, so we don't really have to make tf.random_normal accept both number and tensors as inputs.

I'm curious, what is your original use-case (project) that motived this PR?

thibo73800:
@dsmilkov  I am working on a reinforcement learning algorithm, running into the browser. I use the Actor critic algorithm for continuous value, the predicted value follow a Gaussian distribution. In python, I would use tf.distributions.Normal(mean, std). In this case, mean and std are tensors coming from a neural network... Since tf.random_normal don't allow tensor, I can't apply backpropagation on the result.

PR #1011, opened 2018-04-30
Loss Operations: Move Reduction under losses namespace

ManrajGrover:
This PR moves `Reduction` to losses namespace to align with Tensorflow

https://www.tensorflow.org/api_docs/python/tf/losses/Reduction

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1011)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,7 +46,6 @@ export {DataType, Rank, ShapeMap} from './types';
 
 export * from './ops/ops';
 export {LSTMCellFunc} from './ops/lstm';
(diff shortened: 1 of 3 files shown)


ManrajGrover:
Hi Daniel,

I agree this might break code for users. Just to be clear, will we be incorporating this change when major release happens since some users might expect it to be under `tf.losses`?

ManrajGrover:
I think we can close this.

dsmilkov:
Thanks for the understanding

PR #1012, opened 2018-04-30
Add tf.io.browserDownloads() and tf.io.browserFiles()

caisq:
* `tf.io.browserDownloads()` causes the browser to download a model's artifact as files.
  * In the case of ` tf.Model` (Keras-style model), two files consistent with the tensorflowjs_converter format  will be donwnloaded:
    * A JSON file that contains model topology and weight manifest
    * A binary file that contains the weights of the model
* `tf.io.browserFiles()` supports loading model artifacts from files such as user-selected files.

Towards https://github.com/tensorflow/tfjs/issues/13

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012)


ry:
Wouldn't it make more sense for this to live in tfjs-layers instead of core? `tf.Model` isn't even defined in core.

caisq:
@ry thanks for the comment. It's true that `tf.Model` is not a core construct. But *eventually* we want to support saving both Keras-style `tf.Model`s and non-Keras styles (such as the `FrozenModel`s loaded from converted TF SavedModels). The latter doesn't support saving yet, but we have plans to support that. @pyu10055

nsthorat:
Let's chat about this tomorrow -- overall we think that it may make sense to actually decouple the IOHandler because it's really hard to name these top-level functions that merge load/save. Let's chat about this tomorrow (all of these comments below still apply).

---

---

*[src/io/files.ts, line 1 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBR6T42-wXhECbInm9D:-LBR6T439WZj7jNztu40:bvo59m8) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L1)):*
> ```XML
> /**
> ```

I think we should call this something like browser_files.ts. I don't want to pack all of the node and browser stuff into a single typescript file.

---

*[src/io/files.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLuZVCFCMfNJ5NhcN:-LBSLuZVCFCMfNJ5NhcO:b-yh3yv7) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L32)):*
> ```XML
> const DEFAULT_FILE_NAME_PREFIX = 'model';
> const DEFAULT_JSON_EXTENSION_NAME = '.json';
> const DEFAULT_WEIHGT_DATA_EXTENSION_NAME = '.weights.bin';
> ```

WEIGHT

---

*[src/io/files.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLjd04uaVjPfIfyvl:-LBSLjd04uaVjPfIfyvm:bjqfy1e) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L34)):*
> ```XML
> const DEFAULT_WEIHGT_DATA_EXTENSION_NAME = '.weights.bin';
> 
> export interface DownloadTriggerConfig {
> ```

how about BrowserDownloadTriggerConfig

---

*[src/io/files.ts, line 42 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSMPtzDjbdSQYCNf_-:-LBSMPu-0YHhTJkUwKwF:bun3qsy) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L42)):*
> ```XML
>    * create to bind the download to.
>    */
>   jsonAnchor?: HTMLAnchorElement;
> ```

Let's remove this config entirely. I think we just want to trigger downloads immediately, and always create temporary anchors. If the user wants to build a UI that lets you click a link, they can add an onclick handler which calls your download trigger method.

---

*[src/io/files.ts, line 59 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLl5t0CPjUMa8cWNp:-LBSLl5t0CPjUMa8cWNq:bt7owc0) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L59)):*
> ```XML
> }
> 
> export class DownloadTrigger implements IOHandler {
> ```

how about BrowserDownloadTrigger

---

*[src/io/files.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSMxTU-v5fsiDG0VB3:-LBSMxTU-v5fsiDG0VB4:b-ui075x) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L66)):*
> ```XML
>   private readonly trigger: boolean;
> 
>   constructor(fileNames?: string|string[], config?: DownloadTriggerConfig) {
> ```

can this be a string => string object instead of an array, or alternatively have multiple filename arguments?

---

*[src/io/files.ts, line 67 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLpRp0D-roSMSKzZt:-LBSLpRp0D-roSMSKzZu:b-7zfmck) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L67)):*
> ```XML
> 
>   constructor(fileNames?: string|string[], config?: DownloadTriggerConfig) {
>     if (!window) {
> ```

window == null

---

*[src/io/files.ts, line 164 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOHyf7IMKOd5imWuu:-LBSOHyf7IMKOd5imWuv:b-u0xhen) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L164)):*
> ```XML
> }
> 
> export class Files implements IOHandler {
> ```

BrowserFileChooser? Something to disambiguate since this will not be used in node.

Also, can you split this file into two files? It's better to have 1:1 classes to files, just like in Java.

---

*[src/io/files.ts, line 167 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSO9uMAiVmagRXMiU1:-LBSO9uMAiVmagRXMiU2:b-xfvpqt) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L167)):*
> ```XML
>   private readonly files: File[];
> 
>   constructor(files?: File[]) {
> ```

can you just stick a private here?

How many files are you expecting to support here? There's a throw in load() for files with non-2 length.

Also, are you assuming order of the files? We have standardized filenames for everything, so it would be good to support arbitrary ordering.

---

*[src/io/files.ts, line 182 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOlTa0dcMzGzeGmpa:-LBSOlTa0dcMzGzeGmpb:bngjw9a) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L182)):*
> ```XML
>       const jsonReader = new FileReader();
>       // tslint:disable-next-line:no-any
>       jsonReader.onload = (event: any) => {
> ```

can you type this? I think it may be HTMLEvent

---

*[src/io/files.ts, line 205 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOwQADPqx3najVGZK:-LBSOwQADPqx3najVGZL:b-osos0o) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L205)):*
> ```XML
>         if (weightGroup.paths.length !== 1) {
>           reject(new Error(
>               `When uploading user-selected files, we current support only ` +
> ```

why is this? if we're using a file chooser you can easily select many files. I think we should support that use case.

---

*[src/io/files.ts, line 237 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSQ9Oa5YnGluKnTG6f:-LBSQ9Oa5YnGluKnTG6g:b-2z4xp9) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L237)):*
> ```XML
>  *     {units: 1, inputShape: [10], activation: 'sigmoid'}));
>  * const artifactsInfo = await model.save(tf.io.triggerDownloads(
>  *     ['mymodel.json', 'mymodel.weights.bin'])):
> ```

it seems better for this API to have weights / model separated like this:

```
tf.io.triggerDownloads(
    'myModel.json', ['mymodel.weights.bin1', 'mymodel.weights.bin2']);
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-:-LBSSzrw9p0dB-tOWrBO:bs1d1m6)*

caisq:
Had conversation offline. Will keep the `IOHandler` interface as it is nice for many sroutes that have semmetry between saving and loading. The classes for downloading and uploading files will remain as they are (i.e., two separate files).
But the public helper functions will be renamed:
tf.io.triggerDownloads ==> tf.io.browserDownloads
tf.io.files ==> tf.io.browserFiles

In order to minimize potential confusion with native files in node.js that we will support in the future.

Also, just a heads-up that the complexity of the code in this PR has increased a little as a result of our decision to support  uploading multiple (sharded) weight file in the browser.

---

Review status: 0 of 5 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/io/files.ts, line 1 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBR6T42-wXhECbInm9D:-LBXUnScDGSPAL7TsbnY:bwn1x08) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L1)):*

I think we should call this something like browser_files.ts. I don't want to pack all of the node and browser stuff into a single typescript file.
</blockquote></details>

Done renaming this file and the associated test file.

---

*[src/io/files.ts, line 32 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLuZVCFCMfNJ5NhcN:-LBXdoiB0ns5N0UuBl7n:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L32)):*

WEIGHT
</blockquote></details>

Done.

---

*[src/io/files.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLjd04uaVjPfIfyvl:-LBXddj_-ybX8KvzQmru:bxb0y0k) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L34)):*

how about BrowserDownloadTriggerConfig
</blockquote></details>

This config interface has been removed.

---

*[src/io/files.ts, line 42 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSMPtzDjbdSQYCNf_-:-LBXdgRW3dWpzCzb-89A:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L42)):*

Let's remove this config entirely. I think we just want to trigger downloads immediately, and always create temporary anchors. If the user wants to build a UI that lets you click a link, they can add an onclick handler which calls your download trigger method.
</blockquote></details>

Done.

---

*[src/io/files.ts, line 59 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLl5t0CPjUMa8cWNp:-LBaXsf321KWNNeyf-Pp:bkd22il) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L59)):*

how about BrowserDownloadTrigger
</blockquote></details>

Changed to `browserDownloads`.

---

*[src/io/files.ts, line 66 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSMxTU-v5fsiDG0VB3:-LBXaBZO6AXK6STwbwMo:b9qpnmy) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L66)):*

can this be a string => string object instead of an array, or alternatively have multiple filename arguments?
</blockquote></details>

Done. Per offline discussion. We will allow client only to specify the file-name prefix.

---

*[src/io/files.ts, line 67 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSLpRp0D-roSMSKzZt:-LBXaB7u-gk-Yd8CevU-:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L67)):*

window == null
</blockquote></details>

Done.

---

*[src/io/files.ts, line 164 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOHyf7IMKOd5imWuu:-LBXYwXfAbPeSjFwqPjD:b-pm1isz) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L164)):*

BrowserFileChooser? Something to disambiguate since this will not be used in node.

Also, can you split this file into two files? It's better to have 1:1 classes to files, just like in Java.
</blockquote></details>

per offline discussion, naming this class 'BrowserFiles', as the files don't have to be user-selected files, they can be Files generated from blobs as well.

The two classes and two factory methods will continue to life in `browser_files.ts`, as they are both related to manipulating files in the browser. Doing this also benefits consistency with other files in this folder in the sense that each file defines a save and load method, even though in this case they are distributed between two classes.

---

*[src/io/files.ts, line 167 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSO9uMAiVmagRXMiU1:-LBXaKvM1XoQ7X7al8HA:bxg5bba) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L167)):*

can you just stick a private here?

How many files are you expecting to support here? There's a throw in load() for files with non-2 length.

Also, are you assuming order of the files? We have standardized filenames for everything, so it would be good to support arbitrary ordering.
</blockquote></details>

As discussed in offline discussion, I use File[] here because we need to support FrozenModel artifacts in the future. FrozenModel involves at least 3 files (GraphDef, weights manifest and weights), which is different from tf.Model, which involves 2 files (model.json and weights). File[] is a clean way to capture both.

---

*[src/io/files.ts, line 182 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOlTa0dcMzGzeGmpa:-LBXaxJ41o8fQ7fKOF0o:b-2p0g2r) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L182)):*

can you type this? I think it may be HTMLEvent
</blockquote></details>

I think the type is `Event`. Done.

---

*[src/io/files.ts, line 205 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSOwQADPqx3najVGZK:-LBXdG9j89RQImKEAjt-:bs4oaqp) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L205)):*

why is this? if we're using a file chooser you can easily select many files. I think we should support that use case.
</blockquote></details>

Done adding support for loading from multiple weight files. As discussed offline, an error will be thrown if there are duplicate file basenames (e.g., dir1/weights.bin, dir2/weights.bin) in the weights manifest for some reason.

---

*[src/io/files.ts, line 237 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBSQ9Oa5YnGluKnTG6f:-LBXbEkE3VX3LnYW0RgM:b-dd4zrf) ([raw file](https://github.com/tensorflow/tfjs-core/blob/7e3365ae210b0b6d4ffd0a7442e23d90dd222c27/src/io/files.ts#L237)):*

it seems better for this API to have weights / model separated like this:

```
tf.io.triggerDownloads(
    'myModel.json', ['mymodel.weights.bin1', 'mymodel.weights.bin2']);
```
</blockquote></details>

See my other response about the difference between FrozenModels and tf.Models. Because of the difference in their file counts, doing it as an array is the cleanest approach.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-:-LBXUIs_-3hT-77DAPGA:b-ginj6m)*

nsthorat:
:lgtm_strong:

---

Reviewed 1 of 7 files at r2.
Review status: 1 of 5 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[src/io/browser_files.ts, line 93 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBaatbJCvaT572OROta:-LBaatbJCvaT572OROtb:b3h9ni7) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L93)):*
> ```XML
>       if (modelArtifacts.weightData != null) {
>         const weightDataAnchor = this.weightDataAnchor == null ?
>             document.createElement('a') as HTMLAnchorElement :
> ```

you shouldnt need to cast to HTMLAnchorElement here (magic typescript type lookup for 'a')

---

*[src/io/browser_files.ts, line 209 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayBd71z5tHjKTuONc:-LBayBd71z5tHjKTuONd:b-k62op9) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L209)):*
> ```XML
>         return;
>       };
>       jsonReader.readAsText(this.files[0]);
> ```

use jsonReader variable

---

*[src/io/browser_files.ts, line 250 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayHjf4nBW9kR8gyOY:-LBayHjf4nBW9kR8gyOZ:b-4y42xc) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L250)):*
> ```XML
> 
> /**
>  * Factory method for IOHandler that triggers file downloads.
> ```

How about "Creates a browser file download IOHandler"

similar below

---

*[src/io/browser_files.ts, line 259 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBabFWNASULVfDXTjGD:-LBabFWNASULVfDXTjGE:b-u4whw0) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L259)):*
> ```XML
>  * model.add(tf.layers.dense(
>  *     {units: 1, inputShape: [10], activation: 'sigmoid'}));
>  * const artifactsInfo = await model.save(tf.io.triggerDownloads('mymodel'));
> ```

update this

---

*[src/io/browser_files.ts, line 261 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayQfQ5nexKdR21thA:-LBayQfQ5nexKdR21thB:b-dqvyp2) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L261)):*
> ```XML
>  * const artifactsInfo = await model.save(tf.io.triggerDownloads('mymodel'));
>  * // This will trigger downloading of two files:
>  * //   'mymodel.json' and 'mymodel.weights.bin'.
> ```

qq - should we migrate the other write_weights.py to write ".bin" extensions for consistency?

---

*[src/io/browser_files.ts, line 302 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBabDpG8cye5PjHd2la:-LBabDpG8cye5PjHd2lb:b-u4whw0) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L302)):*
> ```XML
>  * const uploadWeightsInput = document.getElementById('upload-weights');
>  * const model = await tfl.loadModel(
>  *     tfc.io.files([uploadJSONInput.files[0], uploadWeightsInput.files[0]]));
> ```

update this

---

*[src/io/browser_files_test.ts, line 287 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBazJcI9gmLQRgTc4nb:-LBazJcI9gmLQRgTc4nc:bdm9g76) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files_test.ts#L287)):*
> ```XML
>   const reverseOrderValues: boolean[] = [false, true];
>   for (const reverseOrder of reverseOrderValues) {
>     const testTitle = `One group, two paths, reverseOrder=${reverseOrder}`;
> ```

I know you do this in layers, but I prefer we don't do this type of programmatic construction of tests in core. It's much easier to understand / debug if they're static, even if you duplicate code.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-:-LBazejZ21G3tB9cIGeK:b-jbsf8z)*

caisq:
Review status: all files reviewed at latest revision, 21 unresolved discussions, all commit checks successful.

---

*[src/io/browser_files.ts, line 20 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBb2FHQ8rfBUzuEnEEc:-LBbEUwU31uqtgThNEz-:bxnrymq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L20)):*

join the lines above and below this comment into one sentence.
</blockquote></details>

Done. Also removed reference to native files in node.js.

---

*[src/io/browser_files.ts, line 41 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBb2fMV6KVVP6gMs89x:-LBbECef15Sxj7XTm7EU:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L41)):*

In node, the check `if (window == null)` will  throw `ReferenceError: window is not defined`.  Do `if (typeof window === 'undefined')` instead.
</blockquote></details>

Done.

---

*[src/io/browser_files.ts, line 93 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBaatbJCvaT572OROta:-LBbEpNXA4EIUbO52AlA:boyexb3) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L93)):*

you shouldnt need to cast to HTMLAnchorElement here (magic typescript type lookup for 'a')
</blockquote></details>

Done. Thanks!

---

*[src/io/browser_files.ts, line 209 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayBd71z5tHjKTuONc:-LBbFg5v8Br_x8_1fq8h:b-l1x67n) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L209)):*
> jsonReader

This is already using the variable. Maybe I didn't get what  you mean?

---

*[src/io/browser_files.ts, line 250 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayHjf4nBW9kR8gyOY:-LBbF-TB3wnyRiJOfe3o:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L250)):*

How about "Creates a browser file download IOHandler"

similar below
</blockquote></details>

Done.

---

*[src/io/browser_files.ts, line 259 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBabFWNASULVfDXTjGD:-LBbF2ZD72w6J8SCF40u:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L259)):*

update this
</blockquote></details>

Done.

---

*[src/io/browser_files.ts, line 261 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayQfQ5nexKdR21thA:-LBbFAN31_r_iZ2iNyT3:bxt6cr5) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L261)):*

qq - should we migrate the other write_weights.py to write ".bin" extensions for consistency?
</blockquote></details>

Both the python and javascript libraries are agnostic w.r.t. extension names. So I'd say this is not necessary.

---

*[src/io/browser_files.ts, line 302 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBabDpG8cye5PjHd2la:-LBbFND33hya1FxfqJ9z:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L302)):*

update this
</blockquote></details>

Done.

---

*[src/io/browser_files_test.ts, line 287 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBazJcI9gmLQRgTc4nb:-LBbFTRW3N8M1vDudg_9:b-796xgg) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files_test.ts#L287)):*

I know you do this in layers, but I prefer we don't do this type of programmatic construction of tests in core. It's much easier to understand / debug if they're static, even if you duplicate code.
</blockquote></details>

OK. I don't mind removing this kind of parametric tests, as long as you don't mind larger sizes in the test files. Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012)*

caisq:
Review status: 3 of 5 files reviewed at latest revision, 21 unresolved discussions, some commit checks pending.

---

*[src/io/browser_files.ts, line 209 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1012#-LBayBd71z5tHjKTuONc:-LBbUuNO7R1h11vC_cQX:b-tioj6s) ([raw file](https://github.com/tensorflow/tfjs-core/blob/8b2ae3ac442ecad493590c3747488a429d36cbf0/src/io/browser_files.ts#L209)):*
> variable

Using the variable 'jsonFile'.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1012)*

PR #1013, opened 2018-05-03
Avoid disposing intermediate tensor in grad

Lewuathe:
# Purpose

Fix https://github.com/tensorflow/tfjs/issues/227.
As well as other ops, `grad` should not dispose intermediate result. We should be able to call `dispose` explicitly or use `tidy` in such case.


src/tensor_test.ts
```diff
@@ -1089,4 +1089,18 @@ describe('tensor.toString', () => {
...
+    // f(x) = x ^ 3
+    const f = (x: Tensor) => x.pow(tf.scalar(3, 'int32'));
+    // f'(x) = 3x ^ 2
+    const g = tf.grad(f);
+    // f''(x) = 6x
+    const gg = tf.grad(g);
+    const x = tf.tensor1d([2, 3]);
+    const data = gg(x).dataSync();
+
+    expect(data[0]).toEqual(12);
```
bileschi: Please use the same syntax for checking both data[0] and data[1]

nsthorat:

---

*[src/gradients.ts, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1013#-LBaPAhaCA382bhDd8mz:-LBaPAhaCA382bhDd8n-:b-e7yl0h) ([raw file](https://github.com/tensorflow/tfjs-core/blob/73756f8411ae92a0327cf0c40a1e56ea6e5dc070/src/gradients.ts#L146)):*
> ```XML
>                 'shape returned by f([x1,...])');
>       }
>       value.dispose();
> ```

ah this is going to introduce a memory leak, but your obversation is absolutely correct. Can you do this with a tf.tidy() instead of the dispose?

---

*[src/tensor_test.ts, line 1103 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1013#-LB_he8LbOZ5YrT57NC1-r1-1103:-LBaPHBc9nKR8P_lvqhF:b70odga) ([raw file](https://github.com/tensorflow/tfjs-core/blob/604372215f4f5c4a96877c9f73eddd3d98d2dad6/src/tensor_test.ts#L1103)):*
<details><summary><i>Previously, bileschi (Stanley Bileschi) wrote…</i></summary><blockquote>

Please use the same syntax for checking both data[0] and data[1]
</blockquote></details>

You should use expectArraysClose. This method can assert a tensor and a vanilla JS array have the same values.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1013)*

nsthorat:

---

*[src/gradients.ts, line 146 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1013#-LBaPAhaCA382bhDd8mz:-LBavAEI7uFJIjHPvVha:b-3anyun) ([raw file](https://github.com/tensorflow/tfjs-core/blob/73756f8411ae92a0327cf0c40a1e56ea6e5dc070/src/gradients.ts#L146)):*

ah this is going to introduce a memory leak, but your obversation is absolutely correct. Can you do this with a tf.tidy() instead of the dispose?
</blockquote></details>

To be clearer:

```
return tidy(() => {

  const {value, grads} = ENV.engine.gradients(() => f(...args), args, dy);
  // ...
  return grads;
});
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1013)*

dsmilkov:
:lgtm_strong: Thanks! Rerunning travis since we had a strange test failure last time

---

Reviewed 1 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1013#-:-LCG1zjY-lXz3xhjT1t2:b-1j27a5)*

Lewuathe:
Review status: 1 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/tensor_test.ts, line 1093 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1013#-LCG3_y-AQiiCFVHAEfg:-LCG69Od1a2yzJGmdc4F:ban2lem) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1109618ff3e8478ccc361ca77170ff3ec6c4aed8/src/tensor_test.ts#L1093)):*

can you move this test outside of the `describe('tensor.toString')` and inside the `describeWithFlags('tensor', ALL_ENVS, () => {` in the same file above since it does do mathematical operations. This is why the travis fails: https://api.travis-ci.org/v3/job/377450573/log.txt
</blockquote></details>

Oh, good catch. I'm wondering why the test keeps failing. Thanks I'm going to fix it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1013)*

PR #1030, opened 2018-05-09
Add tf.io.browserHTTPRequest

caisq:
* Allows model artifacts to be sent via a multipart/form-data
  HTTP request

Towards: https://github.com/tensorflow/tfjs/issues/13

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1030)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 6 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/io/browser_http.ts, line 102 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFSP04EtY_25ApocWN:-LCFSP04EtY_25ApocWO:btx5vxw) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1d605db7d9207716d4a2e307914065bfa9ed2b64/src/io/browser_http.ts#L102)):*
> ```XML
>   }
> 
>   // TODO(cais): Maybe add load to unify this IOHandler type and the mechanism
> ```

instead of a TODO, file an issue?

---

*[src/io/browser_http.ts, line 128 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFSjQU2QW7s0jesNve:-LCFSjQU2QW7s0jesNvf:bqdm6pt) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1d605db7d9207716d4a2e307914065bfa9ed2b64/src/io/browser_http.ts#L128)):*
> ```XML
>  *
>  * const saveResult = await model.save(tf.io.browserHTTPRequest(
>  *     'http://model-server.domain:5000/upload'));
> ```

nit: domain isn't the right name here

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-:-LCFVQgP1AVTDYP9BEBW:b-jbsf8z)*

caisq:
Review status: 0 of 6 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

*[src/io/browser_http.ts, line 57 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCE_rm0DCeOn3_lFsHq:-LCFVqG8B_sIZ4nsvMPe:bqf201f) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1d605db7d9207716d4a2e307914065bfa9ed2b64/src/io/browser_http.ts#L57)):*

Done. Good point. Thanks.
</blockquote></details>

Now that we don't allow passing in an existing `body`, the `Object.assign` here should be fine, because we don't touch any other fields such as `headers`.

---

*[src/io/browser_http.ts, line 102 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFSP04EtY_25ApocWN:-LCFYh-I4pOFRkhzoYm9:bfmwbv5) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1d605db7d9207716d4a2e307914065bfa9ed2b64/src/io/browser_http.ts#L102)):*

instead of a TODO, file an issue?
</blockquote></details>

Done. https://github.com/tensorflow/tfjs/issues/290

---

*[src/io/browser_http.ts, line 128 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFSjQU2QW7s0jesNve:-LCFWBZ7DSV2eCIKHzoj:b-p3pxnu) ([raw file](https://github.com/tensorflow/tfjs-core/blob/1d605db7d9207716d4a2e307914065bfa9ed2b64/src/io/browser_http.ts#L128)):*

nit: domain isn't the right name here
</blockquote></details>

`localhost.localdomain` used to be the default DNS name of the local machine for some linux distros. I think `domain` is fine.
https://lists.debian.org/debian-devel/2005/10/msg00194.html

But indeed that's not necessary. So I removed it.

---

*[src/io/browser_http.ts, line 50 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFX8aWCo-quQEiL_oD:-LCFXnLhEiSK9_xl0KQl:b3cr4f) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L50)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

should you require the path be of some REST format that take the name of the model?
If not, should the post body contain the name of the model?
</blockquote></details>

The design is to leave metadata information such as name, version etc. of the model up to the client. They can send them in headers.

This keeps the API simple. Users have various constraints and requirements that I don't want to anticipate at this point. Let me know if that sounds good to you.

---

*[src/io/browser_http.ts, line 84 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFW8jSCvrgfgSa261q:-LCFY1Mq9mM0g7XHyEHJ:b-svse8w) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L84)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

this uses browser fetch, it won't be available in node, might be more compatible by using node-fetch directly.
</blockquote></details>

Right, this is why we call the method `browserHTTPReqest`. Unfortunately we can't import node-fetch in tfjs-core.

When we implement HTTP for node, we'll do the following in tfjs-node: use the IOHandler interface from core and import node-fetch to implement an IOHandler. Let me know if that sounds good to you.

---

*[src/io/browser_http.ts, line 128 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFVNC39EuMLw_ePvvm:-LCFYObT3sWA-RRcIgsP:bm9kmzc) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L128)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

I would think a node server example might be more relevant here. 
</blockquote></details>

We can add a node.js server later. The nice thing about a Python server is like this is that you can directly import keras and reconstitute a keras Model object, ready fir `predict` and `fit` calls on the server side. Let me know if that sounds good to you.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1030)*

pyu10055:
:lgtm_strong:

---

Review status: 0 of 6 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[src/io/browser_http.ts, line 50 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFX8aWCo-quQEiL_oD:-LCFnL3rCCulMF0ahQmQ:b1l0fsu) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L50)):*

The design is to leave metadata information such as name, version etc. of the model up to the client. They can send them in headers.

This keeps the API simple. Users have various constraints and requirements that I don't want to anticipate at this point. Let me know if that sounds good to you.
</blockquote></details>

sounds good, still be good to have the metadata (name, version) in the ModelArtifacts, we might need to add a meta file for storage. 

---

*[src/io/browser_http.ts, line 84 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFW8jSCvrgfgSa261q:-LCFqIeT2yBvEgmCNl6H:bb7ohn6) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L84)):*

Right, this is why we call the method `browserHTTPReqest`. Unfortunately we can't import node-fetch in tfjs-core.

When we implement HTTP for node, we'll do the following in tfjs-node: use the IOHandler interface from core and import node-fetch to implement an IOHandler. Let me know if that sounds good to you.
</blockquote></details>

sgtm

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-:-LCGRrVKAVYLbaUcm30g:b-jbsf8z)*

caisq:
Review status: 0 of 6 files reviewed at latest revision, 14 unresolved discussions, some commit checks pending.

---

*[src/io/browser_http.ts, line 50 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1030#-LCFX8aWCo-quQEiL_oD:-LCHG6eb2jafPUos6P_q:b9iy96n) ([raw file](https://github.com/tensorflow/tfjs-core/blob/60438c5328ebb7dd8b7b7c97c71b1b6646ad7676/src/io/browser_http.ts#L50)):*
> metadata

Agreed that metadata is useful. If you look at the existing saving modalities:

1. For downloaded files, the model name is the prefix of the file names.
2. For local storage and indexedDB, the model name is the path key.
3. For HTTP requests (i.e., this PR), it is up to the client of this API to add a name and other metadata for the model, most likely through headers. If we get users asking for it, we can implement it in the library.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1030)*

PR #1034, opened 2018-05-14
Add IS_BROWSER to ENV; add scheme-based URL router to tf.io.* modules

caisq:
The currently supported schemes are:
- localstorage:// for browser local storage
- indexeddb:// for browser IndexedDB
- downloads:// for browser file downloads
- http:// and https:// fro browser HTTP requests

Since browserLocalStorage, browserIndexedDB and browserDownloads
are methods without any optional arguments and can be copmletely
replaced by the URL-like strings, they are removed from the tf.io.*
exports.

Also:
- Export tf.io.getSaveHandler, tf.io.getLoadHandler
- Export tf.io.registerSaveRouter and tf.io.registerLoadRouter

Fixes: https://github.com/tensorflow/tfjs/issues/282
Towards: https://github.com/tensorflow/tfjs/issues/13

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1034)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -31,7 +31,7 @@ export interface Features {
   // Whether to enable debug mode.
   'DEBUG'?: boolean;
   // Whether we are in a node.js (i.e., non-browser) environment.
-  'IS_NODE'?: boolean;
+  'IS_BROWSER'?: boolean;
   // The disjoint_query_timer extension version.
   // 0: disabled, 1: EXT_disjoint_timer_query, 2:
   // EXT_disjoint_timer_query_webgl2.
@@ -53,7 +53,7 @@ export interface Features {
 }
 
 export const URL_PROPERTIES: URLProperty[] = [
-  {name: 'DEBUG', type: Type.BOOLEAN}, {name: 'IS_NODE', type: Type.BOOLEAN},
+  {name: 'DEBUG', type: Type.BOOLEAN}, {name: 'IS_BROWSER', type: Type.BOOLEAN},
   {name: 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_VERSION', type: Type.NUMBER},
   {name: 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_RELIABLE', type: Type.BOOLEAN},
   {name: 'WEBGL_VERSION', type: Type.NUMBER},
@@ -304,8 +304,8 @@ export class Environment {
   private evaluateFeature<K extends keyof Features>(feature: K): Features[K] {
     if (feature === 'DEBUG') {
       return false;
-    } else if (feature === 'IS_NODE') {
-      return typeof window === 'undefined';
+    } else if (feature === 'IS_BROWSER') {
+      return typeof window !== 'undefined';
     } else if (feature === 'BACKEND') {
       return this.getBestBackendType();
     } else if (feature === 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_VERSION') {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -30,7 +30,7 @@ export enum Type {
 export interface Features {
   // Whether to enable debug mode.
   'DEBUG'?: boolean;
-  // Whether we are in a node.js (i.e., non-browser) environment.
+  // Whether we are in a browser (as versus, say, node.js) environment.
   'IS_BROWSER'?: boolean;
   // The disjoint_query_timer extension version.
   // 0: disabled, 1: EXT_disjoint_timer_query, 2:
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -30,6 +30,8 @@ export enum Type {
 export interface Features {
   // Whether to enable debug mode.
   'DEBUG'?: boolean;
+  // Whether we are in a node.js (i.e., non-browser) environment.
+  'IS_NODE'?: boolean;
   // The disjoint_query_timer extension version.
   // 0: disabled, 1: EXT_disjoint_timer_query, 2:
   // EXT_disjoint_timer_query_webgl2.
@@ -51,7 +53,7 @@ export interface Features {
 }
 
 export const URL_PROPERTIES: URLProperty[] = [
-  {name: 'DEBUG', type: Type.BOOLEAN},
+  {name: 'DEBUG', type: Type.BOOLEAN}, {name: 'IS_NODE', type: Type.BOOLEAN},
   {name: 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_VERSION', type: Type.NUMBER},
   {name: 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_RELIABLE', type: Type.BOOLEAN},
   {name: 'WEBGL_VERSION', type: Type.NUMBER},
@@ -302,6 +304,8 @@ export class Environment {
   private evaluateFeature<K extends keyof Features>(feature: K): Features[K] {
     if (feature === 'DEBUG') {
       return false;
+    } else if (feature === 'IS_NODE') {
+      return typeof window === 'undefined';
     } else if (feature === 'BACKEND') {
       return this.getBestBackendType();
     } else if (feature === 'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_VERSION') {
--- a/src/io/browser_files.ts
+++ b/src/io/browser_files.ts
@@ -39,9 +39,10 @@ export class BrowserDownloads implements IOHandler {
   private readonly jsonAnchor: HTMLAnchorElement;
   private readonly weightDataAnchor: HTMLAnchorElement;
 
+  static readonly URL_SCHEME = 'downloads://';
+
   constructor(fileNamePrefix?: string) {
-    // TODO(cais): Use central environment flag when it's available.
-    if (ENV.get('IS_NODE')) {
+    if (!ENV.get('IS_BROWSER')) {
       // TODO(cais): Provide info on what IOHandlers are available under the
       //   current environment.
       throw new Error(
@@ -49,8 +50,8 @@ export class BrowserDownloads implements IOHandler {
           'is not a browser.');
     }
 
-    if (fileNamePrefix.startsWith(URL_SCHEME)) {
-      fileNamePrefix = fileNamePrefix.slice(URL_SCHEME.length);
+    if (fileNamePrefix.startsWith(BrowserDownloads.URL_SCHEME)) {
+      fileNamePrefix = fileNamePrefix.slice(BrowserDownloads.URL_SCHEME.length);
     }
     if (fileNamePrefix == null || fileNamePrefix.length === 0) {
       fileNamePrefix = DEFAULT_FILE_NAME_PREFIX;
@@ -240,13 +241,12 @@ class BrowserFiles implements IOHandler {
   }
 }
 
-const URL_SCHEME = 'downloads://';
 export const browserDownloadsRouter: IORouter = (url: string) => {
-  if (ENV.get('IS_NODE')) {
+  if (!ENV.get('IS_BROWSER')) {
     return null;
   } else {
-    if (url.startsWith(URL_SCHEME)) {
-      return browserDownloads(url.slice(URL_SCHEME.length));
+    if (url.startsWith(BrowserDownloads.URL_SCHEME)) {
+      return browserDownloads(url.slice(BrowserDownloads.URL_SCHEME.length));
     } else {
       return null;
     }
--- a/src/io/browser_files.ts
+++ b/src/io/browser_files.ts
@@ -21,15 +21,19 @@
  */
 
 // tslint:disable:max-line-length
+import {ENV} from '../environment';
+
 import {basename, concatenateArrayBuffers, getModelArtifactsInfoForKerasJSON} from './io_utils';
+import {IORouter, IORouterRegistry} from './router_registry';
 import {IOHandler, ModelArtifacts, SaveResult, WeightsManifestConfig, WeightsManifestEntry} from './types';
+
 // tslint:enable:max-line-length
 
 const DEFAULT_FILE_NAME_PREFIX = 'model';
 const DEFAULT_JSON_EXTENSION_NAME = '.json';
 const DEFAULT_WEIGHT_DATA_EXTENSION_NAME = '.weights.bin';
 
-class BrowserDownloads implements IOHandler {
+export class BrowserDownloads implements IOHandler {
   private readonly modelTopologyFileName: string;
   private readonly weightDataFileName: string;
   private readonly jsonAnchor: HTMLAnchorElement;
@@ -37,7 +41,7 @@ class BrowserDownloads implements IOHandler {
 
   constructor(fileNamePrefix?: string) {
     // TODO(cais): Use central environment flag when it's available.
-    if (typeof window === 'undefined') {
+    if (ENV.get('IS_NODE')) {
       // TODO(cais): Provide info on what IOHandlers are available under the
       //   current environment.
       throw new Error(
@@ -45,7 +49,10 @@ class BrowserDownloads implements IOHandler {
           'is not a browser.');
     }
 
-    if (fileNamePrefix == null) {
+    if (fileNamePrefix.startsWith(URL_SCHEME)) {
+      fileNamePrefix = fileNamePrefix.slice(URL_SCHEME.length);
+    }
+    if (fileNamePrefix == null || fileNamePrefix.length === 0) {
       fileNamePrefix = DEFAULT_FILE_NAME_PREFIX;
     }
 
@@ -233,6 +240,20 @@ class BrowserFiles implements IOHandler {
   }
 }
 
+const URL_SCHEME = 'downloads://';
+export const browserDownloadsRouter: IORouter = (url: string) => {
+  if (ENV.get('IS_NODE')) {
+    return null;
+  } else {
+    if (url.startsWith(URL_SCHEME)) {
+      return browserDownloads(url.slice(URL_SCHEME.length));
+    } else {
+      return null;
+    }
+  }
+};
+IORouterRegistry.registerSaveRouter(browserDownloadsRouter);
+
 /**
  * Creates an IOHandler that triggers file downloads from the browser.
  *
@@ -243,10 +264,17 @@ class BrowserFiles implements IOHandler {
  * const model = tf.sequential();
  * model.add(tf.layers.dense(
  *     {units: 1, inputShape: [10], activation: 'sigmoid'}));
- * const artifactsInfo = await model.save(tf.io.browserDownloads('mymodel'));
+ * const saveResult = await model.save(tf.io.browserDownloads('mymodel'));
  * // This will trigger downloading of two files:
  * //   'mymodel.json' and 'mymodel.weights.bin'.
- * console.log(artifactsInfo);
+ * console.log(saveResult);
+ * ```
+ *
+ * You can also simply pass a string with a 'downloads://' scheme followed by
+ * the file-name prefix to `model.save`:
+ *
+ * ```js
+ * const saveResult = await model.save('downloads://mymodel');
  * ```
  *
  * @param fileNamePrefix Prefix name of the files to be downloaded. For use with
--- a/src/io/browser_http.ts
+++ b/src/io/browser_http.ts
@@ -21,19 +21,26 @@
  * Uses [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
  */
 
+import {ENV} from '../environment';
 import {assert} from '../util';
 
 import {getModelArtifactsInfoForKerasJSON} from './io_utils';
+import {IORouter, IORouterRegistry} from './router_registry';
 // tslint:disable-next-line:max-line-length
 import {IOHandler, ModelArtifacts, SaveResult, WeightsManifestConfig} from './types';
 
-class BrowserHTTPRequest implements IOHandler {
+export class BrowserHTTPRequest implements IOHandler {
   protected readonly path: string;
   protected readonly requestInit: RequestInit;
 
   readonly DEFAULT_METHOD = 'POST';
 
   constructor(path: string, requestInit?: RequestInit) {
+    if (ENV.get('IS_NODE')) {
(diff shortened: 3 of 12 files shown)


src/environment.ts
```diff
@@ -30,6 +30,8 @@ export enum Type {
 export interface Features {
   // Whether to enable debug mode.
   'DEBUG'?: boolean;
+  // Whether we are in a node.js (i.e., non-browser) environment.
+  'IS_NODE'?: boolean;
```
ericdnielsen: This seems a little worrisome to me.   We have IS_NODE here and BACKEND below, shouldn't IS_NODE be derivable from BACKEND?

If this were a "List of features that current backend supports, then I would expect to not see the BACKEND token in this list, and IS_NODE might be ok.

src/environment.ts
```diff
@@ -302,6 +304,8 @@ export class Environment {
   private evaluateFeature<K extends keyof Features>(feature: K): Features[K] {
     if (feature === 'DEBUG') {
       return false;
+    } else if (feature === 'IS_NODE') {
```
ericdnielsen: Given this stanza it seems like it should be IS_BROWSER not IS_NODE.  You're checking if the window object exists, not whether we authoritatively know that we're in node.

src/io/browser_files.ts
```diff
@@ -21,31 +21,38 @@
...
 
-class BrowserDownloads implements IOHandler {
+export class BrowserDownloads implements IOHandler {
   private readonly modelTopologyFileName: string;
   private readonly weightDataFileName: string;
   private readonly jsonAnchor: HTMLAnchorElement;
   private readonly weightDataAnchor: HTMLAnchorElement;
 
   constructor(fileNamePrefix?: string) {
     // TODO(cais): Use central environment flag when it's available.
```
ericdnielsen: remove this todo?

src/io/browser_files.ts
```diff
@@ -233,6 +240,20 @@ class BrowserFiles implements IOHandler {
   }
 }
 
+const URL_SCHEME = 'downloads://';
```
ericdnielsen: Can/should this be a property of the IOHandler?

src/io/browser_http.ts
```diff
@@ -100,6 +107,25 @@ class BrowserHTTPRequest implements IOHandler {
   //   See: https://github.com/tensorflow/tfjs/issues/290
 }
 
+const URL_SCHEMES = ['http://', 'https://'];
```
ericdnielsen: Property of the Handler?

src/io/indexed_db.ts
```diff
@@ -40,7 +44,7 @@ export async function deleteDatabase(): Promise<void> {
 function getIndexedDBFactory(): IDBFactory {
   // TODO(cais): Use central environment flag when it's available.
```
ericdnielsen: Remove this todo.

src/io/io.ts
```diff
@@ -15,27 +15,37 @@
...
 import {IOHandler, LoadHandler, ModelArtifacts, SaveConfig, SaveHandler, SaveResult, WeightsManifestConfig, WeightsManifestEntry} from './types';
 import {loadWeights} from './weights_loader';
 
+const registerSaveRouter = IORouterRegistry.registerSaveRouter;
+const registerLoadRouter = IORouterRegistry.registerLoadRouter;
+const getSaveHandler = IORouterRegistry.getSaveHandler;
+const getLoadHandler = IORouterRegistry.getLoadHandler;
+
 export {
-  browserDownloads,
```
ericdnielsen: I think we might still want to export these for symmetry purpose.  I think its useful to the mental model to expose the non-sugared version that matches the pattern that the non-sugared handlers use.

src/io/router_registry.ts
```diff
@@ -0,0 +1,106 @@
...
+    for (const router of handlerType === 'load' ?
+             this.getInstance().loadRouters :
+             this.getInstance().saveRouters) {
+      const handler = router(url);
+      if (handler !== null) {
+        validHandlers.push(handler);
+      }
+    }
+    if (validHandlers.length === 0) {
+      return null;
```
ericdnielsen: Why null?  This feels like it should also be an error?

nsthorat:
Review status: 0 of 12 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[src/environment.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvD-r1-34:-LCTtF-gEFfKXka-DOQu:b9s7d2b) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/environment.ts#L34)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

This seems a little worrisome to me.   We have IS_NODE here and BACKEND below, shouldn't IS_NODE be derivable from BACKEND?

If this were a "List of features that current backend supports, then I would expect to not see the BACKEND token in this list, and IS_NODE might be ok.
</blockquote></details>

re point 1: It's not, they have different semantics. You could be in Node and use a JS CPU backend.

---

*[src/environment.ts, line 307 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvD-r1-307:-LCTtNuv8IH1W43tXvm3:bnyxnck) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/environment.ts#L307)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Given this stanza it seems like it should be IS_BROWSER not IS_NODE.  You're checking if the window object exists, not whether we authoritatively know that we're in node.
</blockquote></details>

+1 this is a good point. There are other headless settings (react native, etc)

---

*[src/io/browser_files.ts, line 277 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTuSql6olSGbr_U-fY:-LCTuSql6olSGbr_U-fZ:b-3v8lqp) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files.ts#L277)):*
> ```XML
>  *
>  * ```js
>  * const saveResult = await model.save('downloads://mymodel');
> ```

maybe make downloads singular so this is in line with the other prefixes

---

*[src/io/browser_files_test.ts, line 107 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTukmu5qGJBdCQAqBj:-LCTukmu5qGJBdCQAqBk:bo4efqi) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files_test.ts#L107)):*
> ```XML
>   });
> 
>   it('Explicit file name prefix, with existing anchors', async done => {
> ```

dont use async here and below, async tests dont work

---

*[src/io/browser_http.ts, line 110 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvG-r1-110:-LCTurqIE3LGZZ2lX-Ip:b32v57) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_http.ts#L110)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Property of the Handler?
</blockquote></details>

+1

---

*[src/io/io.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvIMN0NhbnNRXLrZm:-LCTvIMN0NhbnNRXLrZn:bser95) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2c9b9a0c09053b09dd04c098ea8179d06dddc06f/src/io/io.ts#L28)):*
> ```XML
> 
> export {
>   browserDownloads,
> ```

what happened to these?

---

*[src/io/local_storage_test.ts, line 111 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvX4q4SOD0uSui253:-LCTvX4q4SOD0uSui254:b-o2dv5f) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2c9b9a0c09053b09dd04c098ea8179d06dddc06f/src/io/local_storage_test.ts#L111)):*
> ```XML
>   it('Save artifacts succeeds', async done => {
>     const testStartDate = new Date();
>     const handler = tf.io.browserLocalStorage('FooModel');
> ```

In general I prefer importing the public facing API from unit tests (it helps you understand what the API looks like) -- can we do that here?

---

*[src/io/router_registry.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTwh1p1dg7n71SPzYd:-LCTwaZ_CUbPXNctbHgQ:b-ieisud) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L22)):*
> ```XML
> export type IORouter = (url: string) => IOHandler;
> 
> export class IORouterRegistry {
> ```

just for my understanding, we need this registry because we want the node IO handlers to not live in this repository?

---

*[src/io/router_registry.ts, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvtZR-SyTY1uWxQ4c:-LCTvtZR-SyTY1uWxQ4d:b-62yhkq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L23)):*
> ```XML
> 
> export class IORouterRegistry {
>   // Singletone instance.
> ```

Singleton

---

*[src/io/router_registry.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTwDna4HQX5RkjFqmC:-LCTwDna4HQX5RkjFqmD:b-ch4a7) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L24)):*
> ```XML
> export class IORouterRegistry {
>   // Singletone instance.
>   private static instance: IORouterRegistry = null;
> ```

no need to initialize to null

---

*[src/io/router_registry.ts, line 88 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCU3Roz9Fsm8pczUZh_:-LCU3Rp-CE9Wwxwm0b6m:bv189d1) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L88)):*
> ```XML
>       IOHandler {
>     const validHandlers: IOHandler[] = [];
>     for (const router of handlerType === 'load' ?
> ```

use forEach for arrays

---

*[src/io/router_registry_test.ts, line 48 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCU441L8mMOiUp1hmP8:-LCU441L8mMOiUp1hmP9:bornzh3) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry_test.ts#L48)):*
> ```XML
>     // tslint:disable:no-any
>     tempRegistryInstance = (IORouterRegistry as any).instance;
>     (IORouterRegistry as any).instance = null;
> ```

reaching into the private variable is a code smell, maybe expose a reset() method

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1034)*

caisq:
Thanks for the review, Eric and Nikhil.

---

Review status: 0 of 12 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[src/environment.ts, line 34 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvD-r1-34:-LCUNQEV-6mkQ2NaTcYh:btv86wg) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/environment.ts#L34)):*

re point 1: It's not, they have different semantics. You could be in Node and use a JS CPU backend.
</blockquote></details>

+1 to Nikhil. So I'll leave it as is. But I'll change it from `iS_NODE` to `IS_BROWSER`. See below.

---

*[src/environment.ts, line 307 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvD-r1-307:-LCUNbVsA1c83sz-7Kqe:b-gk8036) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/environment.ts#L307)):*

+1 this is a good point. There are other headless settings (react native, etc)
</blockquote></details>

Replaced `IS_NODE` with `IS_BROWSER`. Thanks.

---

*[src/io/browser_files.ts, line 43 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvE-r1-43:-LCUQ-oBB4O-hNoC8RhZ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files.ts#L43)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

remove this todo?
</blockquote></details>

Done.

---

*[src/io/browser_files.ts, line 243 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvE-r1-243:-LCUOwMMCvZP_HfpTh4c:bv69w7f) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files.ts#L243)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Can/should this be a property of the IOHandler?
</blockquote></details>

Done throughout indexed_db.ts, local_storage.ts, browser_files.ts and browser_http.ts.

---

*[src/io/browser_files.ts, line 277 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTuSql6olSGbr_U-fY:-LCUP5bu5hMHbKEFuTNt:buaru6w) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files.ts#L277)):*

maybe make downloads singular so this is in line with the other prefixes
</blockquote></details>

I think plural makes more sense, because for both tf.Model and FrozenModel, this will end up being two or three files.

---

*[src/io/browser_files_test.ts, line 107 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTukmu5qGJBdCQAqBj:-LCUQbgJ4l0AIXT8dr9b:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_files_test.ts#L107)):*

dont use async here and below, async tests dont work
</blockquote></details>

Done.

---

*[src/io/browser_http.ts, line 110 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvG-r1-110:-LCUQciw9JC2c_C-Mrx3:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/browser_http.ts#L110)):*

+1
</blockquote></details>

Done.

---

*[src/io/indexed_db.ts, line 45 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvI-r1-45:-LCUQdJeDj4Pd8DHYheo:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/indexed_db.ts#L45)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Remove this todo.
</blockquote></details>

Done.

---

*[src/io/io.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvIMN0NhbnNRXLrZm:-LCUQf-k3QKiqPvM2CDO:bk3jtb3) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2c9b9a0c09053b09dd04c098ea8179d06dddc06f/src/io/io.ts#L28)):*

what happened to these?
</blockquote></details>

See PR description. With these URL schemes, Model.save() and tf.loadModel() can work with only a string argument for browser downloads, local storage and indexedDB. So there is no point keeping these around. They don't have any optional fields, unlike the browserHTTPRequest being kept. I want to follow the principle of "there should be only one way to do it". If we add options to browserDownloads, browserLocalStorage in the future, we can add them back to the exports.

---

*[src/io/local_storage_test.ts, line 111 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvX4q4SOD0uSui253:-LCURFBY4Xm8fmYwFaPd:bm3m58e) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2c9b9a0c09053b09dd04c098ea8179d06dddc06f/src/io/local_storage_test.ts#L111)):*

In general I prefer importing the public facing API from unit tests (it helps you understand what the API looks like) -- can we do that here?
</blockquote></details>

Done using the router registry.

---

*[src/io/router_registry.ts, line 22 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTwh1p1dg7n71SPzYd:-LCUbpv96txLpUVJ5bbq:b9nmlha) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L22)):*

just for my understanding, we need this registry because we want the node IO handlers to not live in this repository?
</blockquote></details>

Yes: it is for tfjs-node. But it can also be used by users to define custom URL schemes.

---

*[src/io/router_registry.ts, line 23 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTvtZR-SyTY1uWxQ4c:-LCUbmIg5Td-9sDvce2O:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L23)):*

Singleton
</blockquote></details>

Done.

---

*[src/io/router_registry.ts, line 24 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCTwDna4HQX5RkjFqmC:-LCUblfg3jGWvYaEyag0:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L24)):*

no need to initialize to null
</blockquote></details>

Done.

---

*[src/io/router_registry.ts, line 97 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCRQp_64e7MBvYDONvN-r1-97:-LCUe83WB2P3hERRLo5-:btskjz2) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry.ts#L97)):*
<details><summary><i>Previously, ericdnielsen wrote…</i></summary><blockquote>

Why null?  This feels like it should also be an error?
</blockquote></details>

I think returning `null` make it a little easier for clients (i.e., tfjs-layers in this case) to use. Clients can just test for null, instead of wrapping things in a try-catch block.

*But* actually coming to think about it, I think in this case it should just return an empty array. And in case there are >1 handlers available, return them all as an array. This was the API is cleaner: it always returns arrays and it delegates the responsibility of deciding what to do when there is 0, 1, or more handlers available to clients. Done.

---

*[src/io/router_registry_test.ts, line 48 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCU441L8mMOiUp1hmP8:-LCUdrZv0H5APJyVZoNC:bmzpczq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/b94fcf6f0b0f330dace756d04873eec3905ac05e/src/io/router_registry_test.ts#L48)):*

reaching into the private variable is a code smell, maybe expose a reset() method
</blockquote></details>

I considered that. But here we not only want to reset the registry, but also want to restore the state back to its original state once the tests are done. For doing that, doing it the way here feels better than adding two methods. This is just a test file.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-:-LCUg7-78ZFNJQNP2ckk:b-1rlk3v)*

nsthorat:
:lgtm_strong:

---

Review status: 0 of 12 files reviewed at latest revision, 16 unresolved discussions.

---

*[src/environment.ts, line 33 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCVEhlTBosp7k9UcUK4:-LCVEhlTBosp7k9UcUK5:bs55twv) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2a2c43a41fd7c2e64f07b83c237d6ba946080eba/src/environment.ts#L33)):*
> ```XML
>   // Whether to enable debug mode.
>   'DEBUG'?: boolean;
>   // Whether we are in a node.js (i.e., non-browser) environment.
> ```

update this comment

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-:-LCVEt2FCh7q333h_WdO:b-jbsf8z)*

caisq:
Review status: 0 of 12 files reviewed at latest revision, 17 unresolved discussions, some commit checks pending.

---

*[src/environment.ts, line 33 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1034#-LCVEhlTBosp7k9UcUK4:-LCVG6AK0jZpcT7dIJHE:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2a2c43a41fd7c2e64f07b83c237d6ba946080eba/src/environment.ts#L33)):*

update this comment
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1034)*

PR #1042, opened 2018-05-16
[WIP] Add optional flag validateShape to Variable.assign

mlajtos:
FEATURE

Validating shape should be optional as in TF:
https://www.tensorflow.org/api_docs/python/tf/assign

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1042)


mlajtos:
This was a bit rushed. Sorry.

Problem is that Tensor.{shape, strides, size, rankType} are ```readonly``` props and they need to be updated when reassigned with tensor that differs in those attributes.

What would be an appropriate change – make them writable in Tensor or override them in Variable?

PR #1049, opened 2018-05-19
Add toPixels() to all tests.

nkreeger:
I have an implementation for the node backend and need to export this test.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1049)


nkreeger:
Hrm after some more research I might need to create a new API since ImageData is only available through Cairo and other third party libraries.

PR #1054, opened 2018-05-23
tf.cumsum should pass the correct axis to backend.cumsum

dsmilkov:
BUG Previously cumsum would pass the original axis to the backend, even if the input was permuted. No tests were failing in core since webgl backend ignores the axis param anyway, but other backends (e.g. node) don't. This fix makes cumsum tests pass in tfjs-node.

Fixes https://github.com/tensorflow/tfjs/issues/326

Also:
 - simplify a bit the cumsum shader

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1054)


nkreeger:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1054#-:-LDCE8Og7k699i_JS2rN:b-jbsf8z)*

PR #1060, opened 2018-05-26
[FEATURE] added op: nonMaxSuppression

justadudewhohacks:
FEATURE added op: nonMaxSuppression

I am looking to convert a mobilenet SSD model using the tfjs converter tool, but as pointed out in [tensorflow/tfjs/issues/188](https://github.com/tensorflow/tfjs/issues/188), there are some OPs missing. As a starting point I added the op for **Non-Maximum Suppression**, which is used in some SSD models such as mobilenet SSD.

Now the question is, there are 3 versions of NonMaxSuppression in tensorflow. 
v1: iouThreshold is assigned as an attr
v2: iouThreshold is now an input argument
v3: scoreThreshold is now an additional argument

This PR basically adds v3, how does tfjs deal with compatibility to models using other op versions? Is it desired to provide an op for every version?

Another question:
How does tfjs deal with node attributes?

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1060)


dsmilkov:
Thank you for the contribution and apologies for the delayed review!

PR #1062, opened 2018-05-28
Check condition type strictly in where op

Lewuathe:
BUG check condition type strictly in where op

# Purpose

TensorFlow [`where`](https://www.tensorflow.org/api_docs/python/tf/where) op only receives `bool` type tensor as a condition. In order to align with TensorFlow, TensorFlow.js implementation also should check condition type strictly regardless of the types of other tensors.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1062)


dsmilkov:
:lgtm_strong: Thank you!

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1062#-:-LDwJ6BQ4sX7B6LO70Wa:b-vn5y6t)*

PR #1064, opened 2018-05-28
Make sure tf.fromPixels() gets called after the DOM is ready

dsmilkov:
BUG Since a newer version of Chrome, you have to create the canvas element when the DOM is ready in order to get a 2d rendering context. This PR postpones the creation of the canvas element until the user calls fromPixels(). This makes the webcam-transfer-learning example work.

Fixes https://github.com/tensorflow/tfjs/issues/313

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1064)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -132,948 +132,958 @@ export class MathBackendWebGL implements KernelBackend {
         }
         this.canvas = document.createElement('canvas');
       }
+      this.canvas.width = pixels.width;
+      this.canvas.height = pixels.height;
+      this.canvas.getContext('2d').drawImage(
+          pixels, 0, 0, pixels.width, pixels.height);
+      pixels = this.canvas;
     }
-    this.canvas.width = pixels.width;
-    this.canvas.height = pixels.height;
-    this.canvas.getContext('2d').drawImage(
-        pixels, 0, 0, pixels.width, pixels.height);
-    pixels = this.canvas;
-  }
-  const tempPixelArray = Tensor.make(texShape, {}, 'int32');
+    const tempPixelArray = Tensor.make(texShape, {}, 'int32');
 
-  // This is a byte texture with pixels.
-  this.texData.get(tempPixelArray.dataId).texType = TextureType.UNSIGNED_BYTE;
-  this.gpgpu.uploadPixelDataToTexture(
-      this.getTexture(tempPixelArray.dataId), pixels);
-  const program = new FromPixelsProgram(outShape);
-  const res = this.compileAndRun(program, [tempPixelArray]);
+    // This is a byte texture with pixels.
+    this.texData.get(tempPixelArray.dataId).texType = TextureType.UNSIGNED_BYTE;
+    this.gpgpu.uploadPixelDataToTexture(
+        this.getTexture(tempPixelArray.dataId), pixels);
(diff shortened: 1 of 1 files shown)


nsthorat:

---

*[src/kernels/backend_webgl.ts, line 306 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1064#-LDbm0NG8cqxEZFOW4yW:-LDbmTBx1YARkac-73KU:b-2ho7iq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/39a02f56dcbfa0016028bef6e3ae839a0a2793f8/src/kernels/backend_webgl.ts#L306)):*

Why are you removing this? I think this is a mistake. We need to create the canvas here.
</blockquote></details>

I see what you're trying to do. fromPixels is not the only place where we will need to initialize the canvas. The WebGLRenderingContext creation will fail since this is created before the DOM is ready, same with the calls to "getCanvas" (which happens from the t-sne code).

Maybe we need to add our own DOMReady listener from the constructor to initialize the canvas, and throw everywhere we try to access an uninitialized gpgpu.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1064)*

nsthorat:
:lgtm_strong:

Giving you an LGTM because I won't be online for a while, but you should make sure this code actually works in a test example since I think not initialize the canvas in the constructor will break a lot of stuff.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1064#-:-LDbnPgP9y1MwqUa180Z:bqg18bx)*

dsmilkov:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/kernels/backend_webgl.ts, line 306 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1064#-LDbm0NG8cqxEZFOW4yW:-LDcCBYk9WoXiEERDe5U:b-je5bnq) ([raw file](https://github.com/tensorflow/tfjs-core/blob/39a02f56dcbfa0016028bef6e3ae839a0a2793f8/src/kernels/backend_webgl.ts#L306)):*

I see what you're trying to do. fromPixels is not the only place where we will need to initialize the canvas. The WebGLRenderingContext creation will fail since this is created before the DOM is ready, same with the calls to "getCanvas" (which happens from the t-sne code).

Maybe we need to add our own DOMReady listener from the constructor to initialize the canvas, and throw everywhere we try to access an uninitialized gpgpu.
</blockquote></details>

Nice catch! You are right, I didn't realize it's the same canvas that we pass to the gpgpu context.

I made a specific fromPixelsCanvas, used only for the fromPixels calls, since according to the actual spec, you shouldn't call both getcontext('2d') and getcontext('webgl') on the same canvas. According to the spec, a call to getcontext(a) followed by a call to getcontext(b) where a != b, should be undefined behavior.

This might be related to why I was getting last 30% failing tests when I was using a global webgl backend, instead of per-describe() backend - call to tf.fromPixels which gets a 2d context on the same canvas might mess up the webgl context.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1064)*

caisq:
Thanks, @dsmilkov !

PR #1067, opened 2018-05-30
Add BrowserHTTPRequest.load()

caisq:
Previously BrowserHTTPRequest only had the save() method. This PR
completes the IOHandler implementation. After this PR is merged
and a new release is put out for tfjs-core, tfjs-layer will be able
to utilize BrowserhTTPRequest.load() for HTTP-based loading.

BrowserHTTPRequest.load() supports requestInit (HTTP request options).

Towards: https://github.com/tensorflow/tfjs/issues/290

FEATURE Add `load()` support to `tf.io.browserHTTPRequest()`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1067)

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+#### Description
+<!--
+Please describe the pull request here.
+Also, if this is an issue/bug fix, please add the issue link for reference here.
+-->
+
+
+---
+<!-- Please do not delete this section -->
+##### For repository owners only:
+
+Please remember to apply all applicable tags to your pull request.
+Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY
+
+For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -36,7 +36,7 @@ module.exports = function(config) {
       accessKey: process.env.BROWSERSTACK_KEY
     },
     reportSlowerThan: 500,
-    browserNoActivityTimeout: 30000,
+    browserNoActivityTimeout: 60000,
     customLaunchers: {
       bs_chrome_mac: {
         base: 'BrowserStack',
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,3 +1,0 @@
-Please remember to apply all applicable tags to your pull request.
-Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY
-For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md 
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -220,7 +220,8 @@ export class Environment {
    * associated with it.  A new backend is initialized, even if it is of the
    * same type as the previous one.
    *
-   * @param backendType The backend type. Currently supports `'webgl'|'cpu'`.
+   * @param backendType The backend type. Currently supports `'webgl'|'cpu'` in
+   *     the browser, and `'tensorflow'` under node.js (requires tfjs-node).
    * @param safeMode Defaults to false. In safe mode, you are forced to
    *     construct tensors and call math operations inside a `tidy()` which
    *     will automatically clean up intermediate tensors.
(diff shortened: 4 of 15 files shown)


caisq:
cc @jameswex

dsmilkov:
:lgtm_strong: Nice work Shanqing. Couple smallish suggestions

---

---

*[src/io/browser_http.ts, line 109 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-LDrVcFc2JDjsd0ULfXy:-LDrVcFc2JDjsd0ULfXz:b-um4q1j) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2333dd072868fa994879098ae7a22878776cd3a7/src/io/browser_http.ts#L109)):*
> ```XML
>   }
> 
>   async load(): Promise<ModelArtifacts> {
> ```

since this is user-facing, would be great to explain what it does in a jsdoc

---

*[src/io/browser_http.ts, line 111 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-LDrUjSHAWx-Pn1Qbzdc:-LDrUjSHAWx-Pn1Qbzdd:bu7gyxo) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2333dd072868fa994879098ae7a22878776cd3a7/src/io/browser_http.ts#L111)):*
> ```XML
>   async load(): Promise<ModelArtifacts> {
>     const modelConfigRequest = await fetch(this.path, this.requestInit);
>     const modelConfig = await modelConfigRequest.json();
> ```

after this line might be a good place to do same input validation (make sure that modelTopology exists as a key, etc)

---

*[src/io/weights_loader.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-LDrVP-OEHgG7Ne0PDOd:-LDrVP-P9fLipsfixUhe:b2jd5ts) ([raw file](https://github.com/tensorflow/tfjs-core/blob/2333dd072868fa994879098ae7a22878776cd3a7/src/io/weights_loader.ts#L33)):*
> ```XML
>  *   length as `fetchURLs`.
>  */
> export async function loadWeightsAsArrayBuffer(
> ```

just double checking: this is exposed to the end users right?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-:-LDrVmCPAb_-ig7afxZk:b-xfm4am)*

caisq:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/io/browser_http.ts, line 111 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-LDrUjSHAWx-Pn1Qbzdc:-LDrs6Uz5mZwAHJwidYT:bgwzc7i) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4dbcc5d5a15a341c24bfba0c22e48fc40fc55cad/src/io/browser_http.ts#L111)):*

after this line might be a good place to do same input validation (make sure that modelTopology exists as a key, etc)
</blockquote></details>

This is actually done implicitly on line 116, which checks to see if there is a weightManifest and takes action accordingly. Also, please note that this load() implementation is written in a way to allow the model JSON to contain weights only or topology only. But I realized that we shouldn't allow *both* weights and topology to be missing. Added a check here. Also added a unit test.

---

*[src/io/weights_loader.ts, line 33 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1067#-LDrVP-OEHgG7Ne0PDOd:-LDrptZ334zZ-S_nkH-b:b-y879ab) ([raw file](https://github.com/tensorflow/tfjs-core/blob/4dbcc5d5a15a341c24bfba0c22e48fc40fc55cad/src/io/weights_loader.ts#L33)):*

just double checking: this is exposed to the end users right?
</blockquote></details>

This is *not* exposed in the public API of tfjs-core. This is just for other modules in tfjs-core to use.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1067)*

meme:
This pull request breaks support in tfjs-node, which requires this module to operate, the offending lines are here:
https://github.com/tensorflow/tfjs-core/pull/1067/files#diff-de1c2d9883e723bf8179aa7e45957a61R118 and here,
https://github.com/tensorflow/tfjs-core/pull/1067/files#diff-5c8893bf31b1cd319675fcf3524961e8R36

I recommend that these are replaced with Node's file system utilities so that loading models from disk can be supported. I am willing to work on this if needed.

PR #1070, opened 2018-05-31
add execute method for inference model to allow retrieve activation for any internal nodes

pyu10055:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1070)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -167,20 +167,13 @@ export interface InferenceModel {
       Tensor|Tensor[]|NamedTensorMap;
 
   /**
-   * Execute the inference for the input tensors and return activation values
-   * for specified output node names.
+   * Single Execute the inference for the input tensors and return activation
+   * values for specified output node names without batching.
    *
    * @param input The input tensors, when there is single input for the model,
    * inputs param should be a Tensor. For models with mutliple inputs, inputs
    * params should be in either Tensor[] if the input order is fixed, or
    * otherwise NamedTensorMap format.
-   * For batch inference execution, the tensors for each input need to be
-   * concatenated together. For example with mobilenet, the required input shape
-   * is [1, 244, 244, 3], which represents the [batch, height, width, channel].
-   * If we are provide a batched data of 100 images, the input tensor should be
-   * in the shape of [100, 244, 244, 3].
-   *
-   * @param config Prediction configuration for specifying the batch size.
    *
    * @param outputs string|string[] Optional. List of output node names to
    * evaluate when running predict(). Defaults to the model's default output.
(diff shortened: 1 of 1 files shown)


caisq:

---

*[src/types.ts, line 167 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1070#-LDr8zis2VtyLRYB2czy:-LDr8zis2VtyLRYB2czz:bnv1bxw) ([raw file](https://github.com/tensorflow/tfjs-core/blob/3a847e1c6b204298151351f4a3d4c7dc4fa3ca33/src/types.ts#L167)):*
> ```
> NamedTensorMap
> ```

Can we remove this option? The outputs will always be returned as a single Tensor or an Array of them. Otherwise it's unclear when the return value will be an array and when it'll be a dict (object).

---

*[src/types.ts, line 193 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1070#-LDr9Hl67WDMAOwS4oiZ:-LDr9Hl67WDMAOwS4oi_:b-w1gr9t) ([raw file](https://github.com/tensorflow/tfjs-core/blob/3a847e1c6b204298151351f4a3d4c7dc4fa3ca33/src/types.ts#L193)):*
> ```
> config: ModelPredictConfig,
> ```

Can we remove this? See my comment in the doc. I think batching is at a level higher than execute().

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1070)*

caisq:
:lgtm_strong:

Thanks!

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/types.ts, line 178 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1070#-LDrIVD164JqNMGeoG0G:-LDrIVD164JqNMGeoG0H:ba6gfpf) ([raw file](https://github.com/tensorflow/tfjs-core/blob/f19d42d75875caf43ec29acba4f4afe452944758/src/types.ts#L178)):*
> ```
>  Optional. 
> ```

This is not optional, right? Also, there is no default, right?

---

*[src/types.ts, line 183 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1070#-LDrIOGM2gRcxW3-mwl6:-LDrIOGM2gRcxW3-mwl7:bxoz9zf) ([raw file](https://github.com/tensorflow/tfjs-core/blob/f19d42d75875caf43ec29acba4f4afe452944758/src/types.ts#L183)):*
> ```
> multiple outputs.
> ```

Add that the order is the same as `outputs`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1070#-:-LDrIKpIBd3ylxRMPOG3:b-afugb9)*

PR #1083, opened 2018-06-06
Generalize the tensor container to arbitrary depth

dsmilkov:
FEATURE

Currently, calling `tf.dispose(obj)` or `tf.tidy(() => obj)` results in walking the `obj` for depth of 1 and disposing of any tensors found.

With this change `obj` is walked to arbitrary depth with cycle-detection to defend against cyclical objects.

This makes `tf.tidy()` and `tf.dispose()` more useful in practice.

Also rename `util.extractTensorsFromContainer` to `util.getTensorsFromContainer` since we are not modifying (extracting from) the container.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1083)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -268,7 +268,7 @@ export class Engine implements TensorManager {
 
     const tensorsToKeep = new Set(this.keepTensors);
 
-    const tensorsToTrackInParent = util.extractTensorsFromContainer(result);
+    const tensorsToTrackInParent = util.getTensorsInContainer(result);
     tensorsToTrackInParent.forEach(tensor => tensorsToKeep.add(tensor.id));
 
     // Dispose the arrays tracked in this scope.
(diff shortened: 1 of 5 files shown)


dsmilkov:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/util.ts, line 459 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1083#-LEKTm1n9QW74gKqeDG4:-LEKXmja9tgFXIa8YFgL:bwto1an) ([raw file](https://github.com/tensorflow/tfjs-core/blob/09c11a2ade97ec540165ce94e7eaf86c22a63ac2/src/util.ts#L459)):*

do you need |void?
</blockquote></details>

yeah because TensorContainer is `void|otherStuff` and void is needed for tidy's that won't return anything.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1083)*

PR #1087, opened 2018-06-07
Upload small tensors that live on CPU as uniforms

dsmilkov:
Tensors that live in GPU and have <= 32 values, are uploaded and sampled as uniforms. 32 is determined empirically. This has no perf impact on sampling, while reducing texture allocation, gpu memory and potential memory leaks since uniforms are cleaned up automatically after the shader finishes.

Fixes https://github.com/tensorflow/tfjs/issues/104

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1087)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -1033,14 +1033,21 @@ export class MathBackendWebGL implements KernelBackend {
     if (output == null) {
       output = this.makeOutputArray(program.outputShape, inputs[0].dtype);
     }
-    const inputsData: Array<TensorData<T>> = inputs.map(input => {
-      this.uploadToGPU(input.dataId);
-      return {tensor: input, texData: this.texData.get(input.dataId)};
+    const inputsData: Array<TensorData<T>> = inputs.map(tensor => {
+      const texData = this.texData.get(tensor.dataId);
+      // Don't upload scalars that live on CPU as texture. We will upload them
+      // as uniforms.
+      if (texData.texture == null && tensor.rank === 0) {
+        return {tensor, texData: null, isUniform: true};
+      }
+      this.uploadToGPU(tensor.dataId);
+      return {tensor, texData, isUniform: false};
     });
     this.uploadToGPU(output.dataId);
     const outputData = {
       tensor: output,
-      texData: this.texData.get(output.dataId)
+      texData: this.texData.get(output.dataId),
+      isUniform: false
     };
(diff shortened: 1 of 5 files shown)


nsthorat:
:lgtm_strong:

Can you add a unit test for explicitly this functionality? I know we indirectly test it, but imagine if someone changes something here and some random test fails, it wouldn't give much information about where to look. Also, if we happen to remove unit tests that indirectly test this, then we lose coverage without knowing. A simple few tests would be fine (also make sure you cover broadcasting since it's a different code path).

---

Reviewed 1 of 5 files at r1, 2 of 2 files at r2.
Review status: 3 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/kernels/webgl/gpgpu_math.ts, line 125 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1087#-LEezAZv-4E98GLYqGeW:-LEezAZv-4E98GLYqGeX:b-tdyxbs) ([raw file](https://github.com/tensorflow/tfjs-core/blob/bc87200fc2918e45c23c8cc9a21ae661c8976c65/src/kernels/webgl/gpgpu_math.ts#L125)):*
> ```XML
>     }
>     // The input is uploaded as uniform.
>     if (texShapeA == null && texShapeB == null) {
> ```

Can you use input.isUniform?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1087#-:-LEezVCF33HuKJZNRJu4:brm13jt)*

dsmilkov:
Ah great point. Added unit tests. It was pretty easy to test using tf.memor().numBytesInGPU (this field is only available when the backend is webgl)

---

Reviewed 3 of 5 files at r1.

---

*[src/kernels/webgl/gpgpu_math.ts, line 125 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1087#-LEezAZv-4E98GLYqGeW:-LEjAmJdEOI5hFIcAUyV:b-7rl0jd) ([raw file](https://github.com/tensorflow/tfjs-core/blob/bc87200fc2918e45c23c8cc9a21ae661c8976c65/src/kernels/webgl/gpgpu_math.ts#L125)):*

Can you use input.isUniform?
</blockquote></details>

done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1087#-:-LEjG8V828XJV8UYnhPE:btyenp2)*

PR #1090, opened 2018-06-08
Migrate environment feature testing to test-if-capable instead of filtering.

nsthorat:
Motivation:
- Allow running tests on multiple devices and test against their native features. We cannot truly emulate iOS (due to precision issues), so to get proper iOS coverage we will run with native features on browserstack.

Goals:
- Migrate to a whitelist of features instead of a constraint filter, with a "default" environment.
- No API changes for `describeWithFlags`

This PR:
- Changes the second param of `describeWithFlags` to take a `Features[]` instead of constraints. This is a list of tests we want to run the given test in. If the real environment doesn't allow for emulation of these flags, we simply skip running it.
- Introduces a native environment. If a test uses `ALL_ENVS`, it turns on default environments. This means the test will try to run against whatever the real environment is (for example, if running on safari on browserstack, will use whatever flags get enabled there). If we detect that the default environment matches an explicit environment, we dedupe. This allows us to run tests against iOS in browser stack using just the default environment, and run all emulations in Chrome.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1090)


nsthorat:

---

*[src/kernels/webgl/gpgpu_context_test.ts, line 90 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEjeMlY935eP_Px2Qaa:-LEjeMlY935eP_Px2Qab:b71ijak) ([raw file](https://github.com/tensorflow/tfjs-core/blob/84d18f2cc049009fe8255613e66c1554de16eb0e/src/kernels/webgl/gpgpu_context_test.ts#L90)):*
> ```XML
> });
> 
> describeWithFlags(
> ```

I'm about to remove this in the next PR so decided to just do it here instead of adding a new flag for iOS in this PR.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1090)*

dsmilkov:
:lgtm_strong: Really nice work! Took some time since I was taking it slow as it is such a sensitive change.

---

Reviewed 8 of 9 files at r1, 1 of 1 files at r2.

---

*[src/environment_test.ts, line 228 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEjqMEa-49_dR55W42Y:-LEjqMEa-49_dR55W42Z:bwce3ha) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/environment_test.ts#L228)):*
> ```XML
>   beforeAll(() => {
>     // Silences backend registration warnings.
>     spyOn(console, 'warn');
> ```

pretty smart way to limit logging to these tests !

---

*[src/jasmine_util.ts, line 17 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk39Gn72fHBgPkEWtV:-LEk39Go6p0ZONxyYjK0:btm3f3a) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L17)):*
> ```XML
>  */
> import {KernelBackend} from '.';
> ```

fix the '.' (i know vscode tends to do this)

---

*[src/jasmine_util.ts, line 21 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk3tLT993Vmcl3TMMk:-LEk3tLT993Vmcl3TMMl:b-z7dx5t) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L21)):*
> ```XML
> import {MathBackendCPU} from './kernels/backend_cpu';
> import {MathBackendWebGL} from './kernels/backend_webgl';
> import {DEFAULT_FEATURES} from './test_util';
> ```

note: jasmine_util didn't depend on test_util previously. I don't see any potential issue here, but wanted to bring it up in case you see anything. (once we had issues with exporting jasmine_util to public api and missing typings to jasmine, but i think this is orthogonal)

---

*[src/jasmine_util.ts, line 31 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk1ivp2Q1vImn9qKMT:-LEk1ivp2Q1vImn9qKMU:bkkyj19) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L31)):*
> ```XML
> 
>   if (feature === 'BACKEND') {
>     for (let i = 0; i < testBackendFactories.length; i++) {
> ```

can be shorter:
```ts
testBackendFactories.map(x => x.name).indexOf(emulatedFeature) >= 0
```

---

*[src/jasmine_util.ts, line 33 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk640D6K1Hz4Kg61O8:-LEk640D6K1Hz4Kg61O9:buuwzra) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L33)):*
> ```XML
>     for (let i = 0; i < testBackendFactories.length; i++) {
>       if (testBackendFactories[i].name === emulatedFeature) {
>         return true;
> ```

hmm, this doesn't check that backend registration actually succeeded. I guess we are assuming whatever is in TEST_BACKENDS will not throw an error during backend-registration time, so checking if the key exists in TEST_BACKEND is enough. I think this is ok for now.

---

*[src/jasmine_util_test.ts, line 3 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk1EgKBBlKF0MygUE6:-LEk1EgKBBlKF0MygUE7:bn3g4qa) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util_test.ts#L3)):*
> ```XML
> /**
>  * @license
>  * Copyright 2017 Google Inc. All Rights Reserved.
> ```

2018 Google LLC

---

*[src/test_util.ts, line 18 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk3Gre9poDs4qQE0LM:-LEk3Gre9poDs4qQE0LN:btdytvl) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/test_util.ts#L18)):*
> ```XML
>  */
> 
> import {Features} from '.';
> ```

fix the '.'

---

*[src/test_util.ts, line 39 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk-bLnBcPaLrQ3KsRP:-LEk-bLnBcPaLrQ3KsRQ:b-12ml48) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/test_util.ts#L39)):*
> ```XML
> // Emulates the current device.
> export const DEFAULT_FEATURES: Features = {};
> export const BROWSER_ENVS: Features[] = [].concat(WEBGL_ENVS).concat(CPU_ENVS);
> ```

just do WEBGL_ENV.concat(CPU_ENVS) (no need for [].concat)

---

*[src/kernels/backend_webgl_test.ts, line 21 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEjtuH6A0yP0jlN6xDm:-LEjtuH6A0yP0jlN6xDn:b-kv8zyx) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/kernels/backend_webgl_test.ts#L21)):*
> ```XML
> import * as tf from '../index';
> import {describeWithFlags} from '../jasmine_util';
> // tslint:disable-next-line:max-line-length
> ```

no need for this line's tslint

---

*[src/ops/lrn_test.ts, line 34 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEjtbTO0GB1KJNp7Klo:-LEjtbTO0GB1KJNp7Klp:b4tp8sc) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/ops/lrn_test.ts#L34)):*
> ```XML
> };
> 
> describeWithFlags(
> ```

strange this shifted, since its 80 columns (no need to wrap)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-:-LEk6sa89a_AhK2FCC7H:be6o9g8)*

nsthorat:

---

*[src/jasmine_util.ts, line 21 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk3tLT993Vmcl3TMMk:-LEk9Kn47_S5PrMkPGum:be8x2kp) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L21)):*

note: jasmine_util didn't depend on test_util previously. I don't see any potential issue here, but wanted to bring it up in case you see anything. (once we had issues with exporting jasmine_util to public api and missing typings to jasmine, but i think this is orthogonal)
</blockquote></details>

Yep, I was aware of this -- this is fine. As long as jasmine_util is a leaf it doesn't matter (if it's included by any of the source code it will pull in jasmine).

---

*[src/jasmine_util.ts, line 33 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk640D6K1Hz4Kg61O8:-LEk9bKQ-DJ7j3jXYmWs:bs69tjz) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L33)):*

hmm, this doesn't check that backend registration actually succeeded. I guess we are assuming whatever is in TEST_BACKENDS will not throw an error during backend-registration time, so checking if the key exists in TEST_BACKEND is enough. I think this is ok for now.
</blockquote></details>

Yep, your observation is absolutely correct. The problem is that the backend registration doesnt happen until beforeAll is called, which is way later than when the describeWithFlags itself is called. This is related to what I said about 'BACKEND' being special cased -- if we move the backend registration to before all of the unit tests, instead of beforeAll, we can check whether the real backend is there.

---

*[src/ops/lrn_test.ts, line 34 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEjtbTO0GB1KJNp7Klo:-LEkAWx2BlYE7uu_8129:bjzxpwv) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/ops/lrn_test.ts#L34)):*

strange this shifted, since its 80 columns (no need to wrap)
</blockquote></details>

Hm.. when I manually removed the line it formatted correctly, thanks!

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1090)*

nsthorat:

---

*[src/jasmine_util.ts, line 33 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1090#-LEk640D6K1Hz4Kg61O8:-LEkY72g2zckWeXVd_rU:bcnztlx) ([raw file](https://github.com/tensorflow/tfjs-core/blob/657c9db3ad9a7e401ad3443be9bb34ce2699a721/src/jasmine_util.ts#L33)):*

Yep, your observation is absolutely correct. The problem is that the backend registration doesnt happen until beforeAll is called, which is way later than when the describeWithFlags itself is called. This is related to what I said about 'BACKEND' being special cased -- if we move the backend registration to before all of the unit tests, instead of beforeAll, we can check whether the real backend is there.
</blockquote></details>

Changed this so we call findBackend and register backends immediately.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1090)*

PR #1092, opened 2018-06-09
Don’t require("crypto") in browsers.

mbostock:
Same fix as tensorflow/tfjs#355.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1092)

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -44,6 +44,7 @@ export default {
     node(),
     // Polyfill require() from dependencies.
     commonjs({
+      ignore: ["crypto"],
       include: 'node_modules/**',
       namedExports: {
         './node_modules/seedrandom/index.js': ['alea']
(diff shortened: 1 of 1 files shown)


dsmilkov:
:lgtm_strong: Thanks! If `crypto` is ignored here,  then users of tfjs-core (e.g. tfjs) won't have to explicitly ignore it, right?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1092#-:-LEeJd_N5GTWsqeGnhae:b9ly0ar)*

nsthorat:
Awesome, thanks Mike! I think you're right, since we ignore it we don't have to do it for all dependencies of core.

PR #1100, opened 2018-06-12
Check whether uniform locations are null before trying to upload them.

nsthorat:
The compiler can be too smart when we're not using a uniform, so we should guard against that case.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1100)


dsmilkov:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1100#-:-LEoxpPU-8-b78JXOlMA:b-jbsf8z)*

PR #1105, opened 2018-06-15
fix index for min

jbencook:
#### Description
This PR fixes a bug in `tf.min` for the cpu backend where if you compute the min along an axis it returns the global min for each element, e.g.
```
tf.setBackend('cpu');
tf.tensor([[1, 2], [3, 4]]).min(0).print();
```
returns
```
Tensor
    [1, 1]
```
I'm not super familiar with these operations but the index of the flat typed array was fixed at 0 instead of using the offset that the other reduction operations use. Changing that seems to fix the behavior I was seeing.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1105)

### DIFF
--- a/src/kernels/backend_cpu.ts
+++ b/src/kernels/backend_cpu.ts
@@ -560,7 +560,7 @@ export class MathBackendCPU implements KernelBackend {
     const aVals = x.dataSync();
     for (let i = 0; i < vals.length; ++i) {
       const offset = i * reduceSize;
-      let min = aVals[0];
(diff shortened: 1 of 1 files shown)


nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1105#-:-LFNNfnm9dHut3S6S6Mj:b-jbsf8z)*

PR #1106, opened 2018-06-15
Optimizers: Change epsilon value as per environment

ManrajGrover:
#### Description
This PR changes epsilon value as per environment for optimizers. Not sure if this will introduce any inconsistencies in different environments.

Attempts to fix: https://github.com/tensorflow/tfjs/issues/425.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1106)

### DIFF
--- a/src/optimizers/adadelta_optimizer.ts
+++ b/src/optimizers/adadelta_optimizer.ts
@@ -26,8 +26,6 @@ import {NamedVariableMap} from '../types';
 import {Optimizer} from './optimizer';
 import * as optimizer_utils from './optimizer_utils';
 
-const DEFAULT_EPSILON = optimizer_utils.getOptimizerDefaultEpsilonValue();
-
 /** @doclink Optimizer */
 export class AdadeltaOptimizer extends Optimizer {
   static className = 'AdadeltaOptimizer';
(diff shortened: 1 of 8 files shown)


nsthorat:
Can you also update this test here to make sure it passes on browerstack?

https://github.com/tensorflow/tfjs-core/blob/master/src/optimizers/adadelta_optimizer_test.ts#L24

You should use "ALL_ENVS" for the environment -- if your change works it will pass on iOS!

---

---

*[src/optimizers/adadelta_optimizer.ts, line 29 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1106#-LFIQl_G-Y1ZP_k4p2sH:-LFIQl_G-Y1ZP_k4p2sI:b-p8yq8n) ([raw file](https://github.com/tensorflow/tfjs-core/blob/58eaaabf34eda1d7dba3f1fb5f8dfeb8eac3255c/src/optimizers/adadelta_optimizer.ts#L29)):*
> ```XML
> import * as optimizer_utils from './optimizer_utils';
> 
> const DEFAULT_EPSILON = optimizer_utils.getOptimizerDefaultEpsilonValue();
> ```

Unfortunately this isn't going to work in unit tests, when the page loads this value will be computed immediately, and the unit tests will not be able to override the environment to get a new epsilon.

You should call this method inside the constructor of the optimizer so we can check the environment flags lazily.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1106#-:-LFIRPS5-BFyJwhamxYu:b8n6c8)*

nsthorat:
Can you use null instead of undefined for defaults? Then LGTM

ManrajGrover:
@nsthorat Yup, fixed it. Had to change default epsilon value to `1e-4` for the tests to pass. Should be good to go now.

nsthorat:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1106#-:-LFIyZtz3AEmRkmSDo67:b-jbsf8z)*

nsthorat:
Thanks Manraj!

PR #1108, opened 2018-06-15
update raw byte calculation to use byteLength

kumarde:
#### Description
This fixes https://github.com/tensorflow/tfjs/issues/432.

Instead of computing byte length by magic number, use `buffer.byteLength`.
---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1108)

### DIFF
--- a/src/io/io_utils.ts
+++ b/src/io/io_utils.ts
@@ -122,10 +122,10 @@ export function concatenateTypedArrays(xs: TypedArray[]): ArrayBuffer {
   xs.forEach(x => {
     // tslint:disable-next-line:no-any
     if (x as any instanceof Float32Array || x as any instanceof Int32Array) {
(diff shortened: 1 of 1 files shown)


caisq:
:lgtm_strong:

Thanks for the PR, @kumarde 

I'm curious: under what circumstance can the byte length calculated with the magic number based on the data type be different from the value of `buffer.byteLength`?

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1108#-:-LFFtY_06t5K-UAnSwq9:b3m5c2k)*

kumarde:
@caisq I think in theory, this should never happen. However, there's a workflow (that I mentioned offline) and a model specifically  where the `.buffer` didn't seem to match the higher level representation of the `TypedArray`.

PR #1116, opened 2018-06-21
Fixing various nan related issues impacted by floating point precision and NaN

reighlan:
This is fixing issue: https://github.com/tensorflow/tfjs/issues/291

On the newer macs we had issues computing pow, mod, acosh, and atanh due to various issues related to NaN not being properly handled at the GPU level. 

In the case of **acosh** and **mod** the problem stemmed from Div not successfully dividing 3/3 (or 6/6, 7/7, etc). In those cases the result would be .99999 and when floor was applied we got 0 instead of 1. 

In the case of **pow** the problem stemmed from **log** not catching NaN for negative numbers.

Lastly, **atanh** failed due to not checking the inputs for values outside of the range [-1, 1]. For those values **atanh** needed to just return NaN directly.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1116)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -15,71 +15,71 @@
  * =============================================================================
  */
 
-import {MemoryInfo, TimingInfo} from '../engine';
-import {ENV} from '../environment';
+import { MemoryInfo, TimingInfo } from '../engine';
+import { ENV } from '../environment';
 import * as axis_util from '../ops/axis_util';
-import {Conv2DInfo} from '../ops/conv_util';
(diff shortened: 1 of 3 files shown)


dsmilkov:
:lgtm_strong: Wohoo! Thanks for figuring this out. I didn't realize it's that involved !

---

Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1116#-:-LFctOsQ0Mmsh9I5raN3:b-u7kl18)*

PR #1111, opened 2018-06-17
Improve error message in tensorXd

Lewuathe:
#### Description

MISC

Throw an error message saying "The input must be non-null value" instead of saying "values is not a Flat/TypedArray" in case of null input value.

see: https://github.com/tensorflow/tfjs/issues/424

---

##### For repository owners only:

Please remember to apply all applicable tags to your pull request.
Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY

For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1111)

PR #1124, opened 2018-06-25
Improve the error messages in Tensor.make()

Lewuathe:
#### Description

MISC

Improve the error message in Tensor.make in case of the input shape is null.

see: https://github.com/tensorflow/tfjs/issues/424

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -15,6 +15,27 @@
  * =============================================================================
  */
 
+const karmaTypescriptConfig = {
+  tsconfig: 'tsconfig.json',
+  // Disable coverage reports and instrumentation by default for tests
+  coverageOptions: {
+    instrumentation: false
+  },
+  reports: {
+  }
+};
+
+// Enable coverage reports and instrumentation under KARMA_COVERAGE=1 env
+const coverageEnabled = !!process.env.KARMA_COVERAGE;
+if (coverageEnabled) {
+  karmaTypescriptConfig.coverageOptions.instrumentation = true;
+  karmaTypescriptConfig.coverageOptions.exclude = /_test\.ts$/;
+  karmaTypescriptConfig.reports = {
+    html: 'coverage',
+    'text-summary': ''
+  };
+}
+
 module.exports = function(config) {
   config.set({
     frameworks: ['jasmine', 'karma-typescript'],
@@ -22,10 +43,7 @@ module.exports = function(config) {
     preprocessors: {
       '**/*.ts': ['karma-typescript'],  // *.tsx for React Jsx
     },
-    karmaTypescriptConfig: {
-      tsconfig: 'tsconfig.json',
-      reports: {} // Do not produce coverage html.
-    },
+    karmaTypescriptConfig,
     reporters: ['progress', 'karma-typescript'],
     browsers: ['Chrome'],
     browserStack: {
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "link-local": "yalc link",
     "publish-local": "yarn build-npm && yalc push",
     "lint": "tslint -p . -t verbose",
+    "coverage": "KARMA_COVERAGE=1 karma start --browsers='Chrome' --singleRun",
     "test": "karma start",
     "test-travis": "./scripts/test-travis.sh"
   },
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -21,7 +21,7 @@ import {Profiler} from './profiler';
 // tslint:disable-next-line:max-line-length
 import {backpropagateGradients, getFilteredNodesXToY, NamedGradientMap, TapeNode} from './tape';
 // tslint:disable-next-line:max-line-length
-import {DataId, setTensorTracker, Tensor, Tensor3D, Variable} from './tensor';
+import {DataId, Tensor, Tensor3D, Variable} from './tensor';
 // tslint:disable-next-line:max-line-length
 import {NamedTensorMap, NamedVariableMap, TensorContainer} from './tensor_types';
 import {getTensorsInContainer, isTensorInList} from './tensor_util';
@@ -94,7 +94,6 @@ export class Engine implements TensorManager {
     this.activeScope = {track: []};
     this.scopeStack = [this.activeScope];
     this.profiler = new Profiler(backend);
-    setTensorTracker(this);
   }
 
   tidy<T extends TensorContainer>(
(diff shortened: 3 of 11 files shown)


bileschi:

---

*[src/tensor_test.ts, line 280 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LFrJ7qs4eTZAbVmcbxb:-LFrJ7qs4eTZAbVmcbxc:b-wsga61) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9461714f529e331986cc8ee800f368f3566c16de/src/tensor_test.ts#L280)):*
> ```
> input value
> ```
nit: "null shape" here

---

*[src/tensor_test.ts, line 281 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LFrJDRh0OkhVIvULWpZ:-LFrJDRh0OkhVIvULWp_:b-vrr1lb) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9461714f529e331986cc8ee800f368f3566c16de/src/tensor_test.ts#L281)):*
> ```
> {values: null}
> ```
Probably best to verify the null shape error when values is non-null.  When both are null it's not clear which error should take precedence. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124)*

nsthorat:

---

*[src/tensor.ts, line 400 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LFrzSrnB12RrXWNPaPq:-LFrzSrnB12RrXWNPaPr:bvn3pvm) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9461714f529e331986cc8ee800f368f3566c16de/src/tensor.ts#L400)):*
> ```XML
>                                              R extends Rank = Rank>(
>       shape: ShapeMap[R], data: TensorData, dtype?: D): T {
>     util.assert(shape !== null, 'shape is null');
> ```

How about

'The shape field passed to the tensor constructor is null. Please pass a non-null shape.'

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124)*

dsmilkov:
This PR is small enough that we should also check for values being null, not just shape. This way this PR can fully fix tensorflow/tfjs#424

---

---

*[src/tensor.ts, line 400 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LFrzSrnB12RrXWNPaPq:-LG5j92hEAxUKLllmQqi:btlqbb9) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9461714f529e331986cc8ee800f368f3566c16de/src/tensor.ts#L400)):*

How about

'The shape field passed to the tensor constructor is null. Please pass a non-null shape.'
</blockquote></details>

+1. Also let's do the validation inside the Tensor constructor and leave the static make method to be just a proxy for the private c-tor.

---

*[src/tensor_test.ts, line 282 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LG5jOYm5v395z6uuQ3r:-LG5jOYn203aRxyXaGff:b-h38az0) ([raw file](https://github.com/tensorflow/tfjs-core/blob/e993abee5c5dee803794048856674755038faf0a/src/tensor_test.ts#L282)):*
> ```XML
>   it('Tensor.make throw error with null shape', () => {
>     const values = new Float32Array([1, 2, 3, 4]);
>     expect(() => Tensor.make(null, {values}))
> ```

Instead of calling Tensor.make(), which is an undocumented method, and used only internally, call the public tf.tensor() via the unit tests.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-:-LG5jdw-CH8Dnmi53qMT:b-fg9e62)*

Lewuathe:

---

*[src/tensor_test.ts, line 282 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LG5jOYm5v395z6uuQ3r:-LG9_P029ss-YLENjhTK:b-7mlwg) ([raw file](https://github.com/tensorflow/tfjs-core/blob/e993abee5c5dee803794048856674755038faf0a/src/tensor_test.ts#L282)):*

Instead of calling Tensor.make(), which is an undocumented method, and used only internally, call the public tf.tensor() via the unit tests.
</blockquote></details>

tf.tensor already have a guard for null shape. It infers the shape in such case. So I think testing the `tf.tensor` here is meaningless. Don't we need to take care of the case when user call `Tensor.make` directly? 
That is the case [the issue](https://github.com/tensorflow/tfjs/issues/424) cares.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124)*

dsmilkov:

---

*[src/tensor_test.ts, line 282 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1124#-LG5jOYm5v395z6uuQ3r:-LGAWgr8BZ471UytZJPr:bdrn2vu) ([raw file](https://github.com/tensorflow/tfjs-core/blob/e993abee5c5dee803794048856674755038faf0a/src/tensor_test.ts#L282)):*

tf.tensor already have a guard for null shape. It infers the shape in such case. So I think testing the `tf.tensor` here is meaningless. Don't we need to take care of the case when user call `Tensor.make` directly? 
That is the case [the issue](https://github.com/tensorflow/tfjs/issues/424) cares.
</blockquote></details>

I'm asking for this file specially to unit-test public API, namely tensor(), tensor1/2/3/4/5D(), instead of unit testing Tensor.make().

Also I just edited the issue you linked to say: "When someone calls tensor()" , instead of "When someone calls Tensor.make()" since they are equivalent - one calls the other directly.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1124)*

dsmilkov:
Hi Kai,

I just saw you wrote https://github.com/tensorflow/tfjs-core/pull/1111 which addressed https://github.com/tensorflow/tfjs/issues/424. With that I don't think there is a need to improve err msg in `Tensor.make()` or add a null check for values in the Tensor c-tor since null is reserved for un-initialized tensors. In that, I think we can drop this PR. Thanks for the contribution!

Lewuathe:
@dsmilkov Thanks for checking. I'm going to close this.

PR #1129, opened 2018-06-28
Ensure that tf.exp() uses non-int32 tensors.

nkreeger:
Fixes https://github.com/tensorflow/tfjs/issues/445#event-1687297929

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1129)

### DIFF
--- a/src/ops/unary_ops.ts
+++ b/src/ops/unary_ops.ts
@@ -151,6 +151,7 @@ export class UnaryOps {
   @operation
   static exp<T extends Tensor>(x: T|TensorLike): T {
     const $x = convertToTensor(x, 'x', 'exp');
+    util.assert($x.dtype !== 'int32', 'Input dtype must not be `int32`.');
 
     const bck = (dy: T, saved: Tensor[]) => {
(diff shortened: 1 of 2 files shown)


nsthorat:
Can we actually solve this by relaxing the requirement in node by casting to a float32 if it's an integer?

nkreeger:
We could - but then we differentiate from the TF api:
https://www.tensorflow.org/api_docs/python/tf/exp

nsthorat:
Yep, I know, but it's a bit of a strange requirement to have tf.exp not allow integers.

dsmilkov:
Based on a few reports on github and the mailing list, there is a need to relax tfjs API to allow mixing ints and floats (e.g. allowing `mul(int, float)`, `add(int, float)` etc). This makes the API different from TF, but not in a way that will hurt the user, yet improve usability of the library.

nkreeger:
Closing and fixing in https://github.com/tensorflow/tfjs-node/pull/107

PR #1130, opened 2018-06-28
When backend is already registered, reuse existing backend

dsmilkov:
Fixes https://github.com/tensorflow/tfjs/issues/109.

When the backend is already registered, re-use the existing backend. Also tested that double imports in the browser works:

```html
<script src="tf-core.min.js"></script>
<script src="tf-core.min.js"></script>
<script>tf.square(3).print()</script>
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1130)


dsmilkov:
This is ready for another look

PR #1133, opened 2018-06-29
Loss Ops: Adds Softmax Cross Entropy Loss

ManrajGrover:
#### Description
This PR adds `tf.losses.softmaxCrossEntropy` with weights support.

1. Renames current `tf.losses.softmaxCrossEntropy` -> `tf.softmaxCrossEntropyWithLogits` (since loss function depends on it, can be moved to loss function itself)
2. Adds support for `tf.losses.softmaxCrossEntropy` with weighted loss.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1133)


ManrajGrover:
Ah, not sure what to do here. Since softmax makes use of `log`, for Safari any value less than `1e-4` gets converted to `0` which `log` returns as `NaN`. And if values are clipped at `1e-4`, test cases fail.

ManrajGrover:
@dsmilkov Made the changes

PR #1145, opened 2018-07-09
use setImmediate instead of requestAnimationFrame on Node.js

ChALkeR:
#### Description

This makes `await tf.nextFrame();` work on Node.js for better code interoperability.

It also does the expected thing of unblocking the event loop and giving the engine time to perform other tasks, though the name is a bit confusing.

The order of the check is important: IE and Edge also implement `setImmediate`, but all versions that do that also have `requestAnimationFrame`, so `requestAnimationFrame` is used on those.

Refs: [setImmediate](https://nodejs.org/api/timers.html#timers_setimmediate_callback_args), [Node.js Event Loop](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/).

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1145)

### DIFF
--- a/src/browser_util.ts
+++ b/src/browser_util.ts
@@ -16,15 +16,21 @@
  */
 import {doc} from './doc';
 
+const delayCallback = typeof requestAnimationFrame !== 'undefined'
+  ? requestAnimationFrame // Browsers
+  : setImmediate; // Node.js
+
 export class BrowserUtil {
   /**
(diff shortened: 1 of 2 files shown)


ChALkeR:
@dsmilkov I added tests, but using `describeWithFlags('nextFrame'` — that looks to be more in line with the other tests.

PR #1150, opened 2018-07-10
Allow tf.dispose(any), not just tf.dispose(TensorContainer)

davidsoergel:
The disposal code does not actually require the TensorContainer constraint.  It is useful to be able  tf.dispose(...) any object without type constraints (particularly in dependent libraries where TensorContainer is not a central concept).

This fixes a (perhaps inadvertent) breaking API change in #1083.

Happy to discuss, of course, whether the tighter API is actually what we want!

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1150)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -164,7 +164,8 @@ export class Environment {
    *     `Promise`s are not supported.
    */
   @doc({heading: 'Performance', subheading: 'Memory'})
-  static dispose(container: TensorContainer) {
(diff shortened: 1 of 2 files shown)


davidsoergel:
Not needed: per offline discussion, I can do `tf.dispose(myObject as {})`, so long as I'm confident that `myObject` is not a Promise.

PR #1151, opened 2018-07-10
engine: name all scopes by default

ChALkeR:
#### Description
TypeScript doesn't catch that — `.name` is expected to be a string in `runKernel` -> `profileKernel` -> `logKernelProfile`.

Without this, `test-node` prints unhandled rejections due to `name` being undefined in that code path.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1151)


ChALkeR:
https://github.com/tensorflow/tfjs-core/blob/2fe1567ed7ad94cc441c6876412993d77eb22b4a/src/jasmine_util.ts#L133-L135 starts unnamed scopes and causes the abovementioned problem.

Alternatively, scope name could be required in `startScope`, but that is an incompatible change.

PR #1158, opened 2018-07-11
Remove the doc decorator in favor of a jsdoc annotation.

nsthorat:
- Remove the doc decorator in favor of a jsdoc annotation. This removes complexity and stack traces.
- Remove classes for ops as they can now be floating functions.
- `op` now takes a dictionary mapping function name to the method, allowing op callers to use the sugar `const pad = op({pad_});`, attaching a name to the returned op function as well as naming the scope.

Verified no change in number of symbols exported:
```
Found: 
  7 headings
  29 subheadings
  173 methods
```

Corresponding doc generator change: https://github.com/tensorflow/tfjs-website/pull/180
Corresponding layers change: https://github.com/tensorflow/tfjs-layers/pull/267

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1158)


ChALkeR:
Yay, this is great! I wanted to suggest that myself when I noticed those.
The doc decorator needlessly polluted transplited code with `__decorate` function in each file.

nsthorat:
Yup +1, we were being too fancy. Learned our lesson :)

PR #1184, opened 2018-07-25
Allow `tf.loadModel(url)` in node.js if fetch exists

aman-tiwari:
#### Description
Check for `fetch` instead of checking for `ENV.get('IS_BROWSER')`. This lets people use `tf.loadModel('http://....')` if they import a fetch polyfill, e.g:

```ts
import fetch from 'node-fetch'
global.fetch = fetch
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1184)

### DIFF
--- a/src/io/browser_http.ts
+++ b/src/io/browser_http.ts
@@ -21,7 +21,6 @@
  * Uses [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
  */
 
-import {ENV} from '../environment';
 import {assert} from '../util';
 
 // tslint:disable:max-line-length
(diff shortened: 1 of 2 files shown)


caisq:
Thanks for the PR @aman-tiwari . This step is necessary, but not sufficient to get http:// loading working in tfjs-node. I will take a detailed look at that feature soon.

aman-tiwari:
Hey @caisq, not sure what you mean. If one imports the fetch polyfill, this seems to work. One could include the fetch polyfill in tensorflow.js to get it to work without requiring the client to import it, but this seems like the clients responsibility to appropriately prepare the environment.

aman-tiwari:
Added unit tests as per our in-person conversation.

PR #1190, opened 2018-07-26
Add tf.addN(), allow bool weights and improve tslint

dsmilkov:
- Add tf.addN() following [tf.add_n](https://www.tensorflow.org/api_docs/python/tf/add_n) in Python. This is needed when converting the `ssd_mobilenetv2_coco` model.
- Add ability to load weights of dtype `bool`
- Improve tslint rule so `imports {}` and `exports {}` are not subject to 80 width line length
- Also fix circular dep which was disabled on some files due to `//tslint:disable`

FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1190)


dsmilkov:
Will submit since the Travis failure is unrelated (one of the browsers got disconnected)

PR #1196, opened 2018-07-31
Add support for tensors with 0 in shape

dsmilkov:
Fixes https://github.com/tensorflow/tfjs/issues/567 (see the issue for details)

FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1196)


dsmilkov:
Let me know if you have ideas for other ops to test (in addition to matmul, add and concat). Asking since as you can see, the fix sometimes is op-specific, like in the concat case.

PR #1215, opened 2018-08-07
Expose describeWithFlags as tf.test_util.describeWithFlags

caisq:
DEV

Towards: https://github.com/tensorflow/tfjs/issues/575

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1215)

### DIFF
--- a/src/test_util.ts
+++ b/src/test_util.ts
@@ -39,6 +39,8 @@ export const CPU_ENVS: Features = {
 };
 export const ALL_ENVS: Features = {};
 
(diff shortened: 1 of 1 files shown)


nsthorat:
I think I left a comment on the other PR about why we can't do this. `test_util` currently is exported by index, which means it's part of our bundle. Adding this introduces a dependency on jasmine, which bloats the binary.

`jasmine_util` is the only exception to the policy about public exports. You can import this as a file directly.

nsthorat:
No worries :)

PR #1233, opened 2018-08-18
allow backend engine to be cached by backend name

pyu10055:
#### Description
This would allow multiple backends used in the same session, and it will prevent
tensor ref count being cleared when switching between backends.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1233)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -403,6 +403,10 @@ export class Environment {
     }
     this.registry[name].backend.dispose();
     delete this.registry[name];
+
+    if (name in this.engines) {
+      delete this.engines[name];
+    }
   }
 
   get engine(): Engine {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -31,6 +31,7 @@ const TEST_EPSILON_FLOAT32 = 1e-3;
 
 export class Environment {
   private features: Features = {};
+  private engines: {[id: string]: Engine} = {};
   private globalEngine: Engine;
   private registry:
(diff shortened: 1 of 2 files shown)


nkreeger:
This seems bad - keeping Tensors in different backends will cause problems when trying to use them in Ops executed by different backends...

nsthorat:
Nick is right -- I don't think we can go forward with this change because Tensors fundamentally cannot be shared between backends because the backend owns the mapping between DataId => data. Let's set up a VC at some point to talk about the problem and a better solution :)

dsmilkov:
I think caching the engine is a good idea and doesn't change anything about previous behavior - backends still can't share tensors. What this change makes possible is the following:

```ts
// Some work on webgl.
tf.setBackend('webgl');
const squareOnWebGL = tf.square(2);

// Then a bit on cpu.
tf.setBackend('cpu');
const sqrtOnCpu = tf.sqrt(squareOnWebGl.dataSync());

// Then back to webgl
tf.setBackend('webgl');
// Note that squareOnWebGL is not lost.
const addOnWebGL = tf.add(squareOnWebGL, sqrtOnCpu.dataSync());
```

nkreeger:
Just curious what the use case is here - FWIW TensorFlow keeps Tensors stored on "devices" and automatically handles copying them over if they need to run on different host memory. I think exposing a way to jump between backends for an end user might be confusing and lead to misuse/bugs/etc.

Issue #467, opened 2017-12-26
Add unit tests for gradientScope memory cleanup.


nsthorat:
When using a `math.gradientScope`, we should test that the intermittent gradients from backprop / inner `math.scope()`s have not been cleaned up.

This depends on `math.getNumArrays` being fixed.

dsmilkov:
This got fixed, right? Should we close the issue?

PR #1241, opened 2018-08-21
solve linear equation (#467)

kedevked:
#### Description

Using Jordan-Gauss elimination to solve linear equation
Compute the determinant of a matrix
Compute the invert of a matrix
Compute the adjoint of a matrix

---

FEATURE solve equation AX=B https://github.com/tensorflow/tfjs/issues/467

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1241)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-core",
-  "version": "0.12.11",
+  "version": "0.12.12",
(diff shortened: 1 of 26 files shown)


nsthorat:
Since this isn't part of TensorFlow core we won't have this as part of core, so closing this out. If folks want this they can use this as a reference implementation.

PR #1244, opened 2018-08-22
Change epsilon for 32 bit float to 1e-7 and add couple more unit tests

dsmilkov:
Epsilon for 32 bit floats is changed to 1e-7 so that `tf.clip([0, 1], eps, 1-eps)` returns `[eps, 1-eps]` instead of `[0, 1]`.

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1244)

### DIFF
--- a/src/environment_test.ts
+++ b/src/environment_test.ts
@@ -18,6 +18,7 @@
 import * as device_util from './device_util';
 import {ENV, Environment} from './environment';
 import {Features, getQueryParams} from './environment_util';
+import * as tf from './index';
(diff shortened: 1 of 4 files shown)


src/environment_test.ts
```diff
@@ -288,4 +289,8 @@ describeWithFlags('epsilon', {}, () => {
     const epsilonValue = ENV.backend.floatPrecision() === 32 ? 1e-8 : 1e-4;
```
manrajgrover: Seems like value in this test should also be changed.

PR #1245, opened 2018-08-23
Update test commands

happyhj:
#### Description
Currently, running `$ yarn test --browsers=Chrome --grep='multinomial'` don't filters tests.
 
The syntax to use to pass arguments to the package script running with `yarn (run)` has changed.

Because of this, the arguments that are passed to the `yarn test` which is posted in DEVELOPMENT.md are not applied to the `karma start` command.

So I updated the command to match the new syntax.

Please refer to the issue below
https://github.com/yarnpkg/yarn/issues/1667

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1245)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -33,7 +33,7 @@ $ yarn lint
 To run a subset of tests and/or on a specific browser:
 
 ```bash
-$ yarn test -- --browsers=Chrome --grep='multinomial'
+$ yarn test --browsers=Chrome --grep='multinomial'
  
 > ...
 > Chrome 62.0.3202 (Mac OS X 10.12.6): Executed 28 of 1891 (skipped 1863) SUCCESS (6.914 secs / 0.634 secs)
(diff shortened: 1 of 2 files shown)


nsthorat:
Which version of yarn are you using?

With `yarn -v` I get `1.5.1`.

When I run this I get the following message:
```
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```

In the issue that you linked, the user is using yarn version `0.16.1`.

tafsiri:
One fix could be to specify a minimum yarn version in the 'engines' entry of our package.json https://yarnpkg.com/en/docs/package-json#toc-engines. We could try `>= 1.4.0` as a start.

happyhj:
@nsthorat You're right. My yarn version was outdated. I updated yarn to v1.5.1 and not it works like a magic. So I got the test commands back.

@tafsiri I think that is good idea to specify minimum yarn version. So I done that on last commit.

happyhj:
The travis CI's yarn version is v1.3.2 so I lower the minimum yarn version in the 'engines' entry to fix the error on CI.

PR #1246, opened 2018-08-23
serialization.SerializationMap.register -> serialization.registerClass

caisq:
This will simplify the public API as well.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1246)

### DIFF
--- a/src/optimizers/adadelta_optimizer.ts
+++ b/src/optimizers/adadelta_optimizer.ts
@@ -18,7 +18,7 @@
 import {ENV} from '../environment';
 import {keep, tidy} from '../globals';
 import {scalar, zerosLike} from '../ops/ops';
-import {ConfigDict, Serializable, SerializableConstructor, SerializationMap} from '../serialization';
+import {ConfigDict, registerClass, Serializable, SerializableConstructor} from '../serialization';
 import {Scalar} from '../tensor';
 import {NamedVariableMap} from '../tensor_types';
 import {Optimizer} from './optimizer';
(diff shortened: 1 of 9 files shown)


src/serialization.ts
```diff
@@ -132,3 +132,8 @@ export class SerializationMap {
         [cls, cls.fromConfig];
   }
 }
+
+export function registerClass<T extends Serializable>(
```
ericdnielsen: Add some public facing docstring.

src/optimizers/adadelta_optimizer.ts
```diff
@@ -119,4 +119,4 @@ export class AdadeltaOptimizer extends Optimizer {
     return new cls(config.learningRate, config.rho, config.epsilon);
   }
 }
-SerializationMap.register(AdadeltaOptimizer);
+registerClass(AdadeltaOptimizer);
```
ericdnielsen: Lets make this serialization.registerClass.

I think its important that we don't lose the "serialization" is why we are registering this class. and not registering a class is not doing something larger in the context of the whole sytem.

I think this means we want to double import.
import serialization from ....
import {ConfigDict, Serializable, SerializableConstructor} from ...

We directly import the three base classes/interfaces because their name describes their purpose, and they are used so commonly that qualifiying their names gets messy/verbose very quickly.

I think the other option for "registerClass" is call it "registerSerializableClass" and then we could avoid the non-qualified import for it.

caisq: I prefer to keep it called "registerClass" for succinctness. I re-revised the code so that they say `serialization.registerClass(...)`.

PR #1248, opened 2018-08-26
Svd

kedevked:
#### Description
Compute the SVD for a matrix m

Use the QR decomposition to compute eingen values and vectors
Use the solve function to compute v as the result of (us)X = m
Use a fix number of iterations that can be improved

---
FEATURE: Singular-value decomposition (SVD) #110

### DIFF
--- a/src/ops/linalg_ops.ts
+++ b/src/ops/linalg_ops.ts
@@ -19,10 +19,12 @@
  * Linear algebra ops.
  */
 
+import {concat, linalg} from '..';
 import {ENV} from '../environment';
 import {dispose} from '../globals';
 import {Tensor, Tensor1D, Tensor2D} from '../tensor';
 import {assert} from '../util';
+
 import {eye, split, squeeze, stack, unstack} from './array_ops';
 import {norm} from './norm';
 import {op} from './operation';
@@ -227,5 +229,29 @@ function qr2d(x: Tensor2D, fullMatrices = false): [Tensor2D, Tensor2D] {
   }) as [Tensor2D, Tensor2D];
 }
 
+/* function svd_(m: Tensor2D) {
+  const mT = m.dot(transpose(m));
+  const tM = transpose(m).dot(m);
+} */
+
+function eingenValues_(m: Tensor): Tensor1D {
+  for (let i = 0; i < 10; i++) {
+    const [x, y] = linalg.qr(m);
+    x.print();
+    y.print();
+    m = y.mul(x);
+  }
+  return diagonalElements(m);
+}
+
+function diagonalElements(m: Tensor): Tensor1D {
+  const ei: Tensor1D[] = [];
+  for (let i = 0; i < m.shape[0]; i++) {
+    ei.push(m.slice([i, i], [1, 1]).as1D());
+  }
+  return concat(ei);
+}
+
 export const gramSchmidt = op({gramSchmidt_});
 export const qr = op({qr_});
+export const eingenValues = op({eingenValues_})
--- a/src/ops/linalg_ops.ts
+++ b/src/ops/linalg_ops.ts
@@ -19,7 +19,7 @@
  * Linear algebra ops.
  */
 
-import {concat, linalg} from '..';
+import {concat, linalg, sqrt, transpose} from '..';
 import {ENV} from '../environment';
 import {dispose} from '../globals';
 import {Tensor, Tensor1D, Tensor2D} from '../tensor';
@@ -29,7 +29,8 @@ import {eye, split, squeeze, stack, unstack} from './array_ops';
 import {norm} from './norm';
 import {op} from './operation';
 import {sum} from './reduction_ops';
-import {tensor2d} from './tensor_ops';
+import {tensor1d, tensor2d} from './tensor_ops';
+
 
 /**
  * Gram-Schmidt orthogonalization.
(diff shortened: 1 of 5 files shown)


DirkToewe:
Hi @kedevked,

It's awesome that You've started the implementation of some of the most crucial Linear Algebra operations. I'm afraid, however, that the Implementation of the QR Algorithm has to become a lot more complicated before it becomes fast enough and converges reliably. Some observations:

  * The Schur Decomposition of a real-valued matrix is usually not real-valued. There is a more lenient version of the Schur Decompostion: The [Real Schur Decomposition](http://people.inf.ethz.ch/arbenz/ewp/Lnotes/chapter2.pdf#page=7), which only guarantess a block-triangular matrix and is real-valued for every real-valued input.

  * The unshifted QR-Algorithm likely never converges to the Real Schur decomposition

  * The QR Decomposition takes O(n³), so the QR Algorithm - as implemented - will take O(n<sup>4</sup>). This can be improved upon by performing a Hessenberg Decomposition first, as a Hessenberg matrix can be QR-Decomposed in O(n²).

  * It is usually recommended to [balance](https://www.mathworks.com/help/matlab/ref/balance.html) the matrix before the Schur Decomposition. [Aside from a few exeptions](http://www.sci.wsu.edu/math/faculty/watkins/pdfiles/balbad.pdf), it is supposed to improve the numeric stability.

There is only one viable Schur Decomposition Algorithm for the non-symmetric Eigenvalue Problem that I could find: The Doubly-Shifted Francis QR Algorithm. There are two JavaScript implementations of the Francis Algorithm that i know of: [Numeric.js](https://github.com/sloisel/numeric) and [NDJS](https://github.com/DirkToewe/ndjs/blob/master/nd.js). The latter contains some literature references that I found very helpful while implementing it. Maybe they will help You as well.

Alternatively, I can offer to try and incorporate  my implementation from NDJS into TFJS.

kedevked:
Hi @DirkToewe ,
tfjs already supports [QR decomposition](https://js.tensorflow.org/api/0.13.3/#linalg.qr) using householder transformation. I used that QR decomposition and implemented the algorithm described here: http://www.math.tamu.edu/~dallen/linear_algebra/chpt6.pdf . I don't really think that in top of that we need the schur decomposition before computing the svd. 
Maybe @caisq could give some pointers in the direction to go.

kedevked:
The PR is about creating a new operator tf.svd that will return the SVD decomposition of a given matrix. Therefore, I am not exporting eigen values and vectors only for testing.  
The third point holds if I were to implement the QR decomposition. That's not the case.

JasonShin:
@kedevked hey, thanks for SVD implementation! I see there is a TSLint error and I made a PR to your fork that fixes it. I wish to use this API soon in my project!

kedevked:
That's great @JasonShin. Hopefully @caisq could review it or give us glimpse about how to proceed to make it part of the API

DirkToewe:
Sorry to bring this up again but I am still wondering about a few points:

1. I am not sure that the QR *iteration* - even for symmetric positive semidefinite matrices - is guaranteed to converge to a diagonal matrix without a [shift](https://www.igpm.rwth-aachen.de/Download/ss13/EWP/ex6_new.pdf).
2. Currently the QR iteration in `eingen_()` uses a fixed number of 5 steps. The Francis QR Algorithm on average takes O(n) steps (literature says ~2n, for random matrices it was up to 3n in my experiments). That should also be the order of steps necessary for symmetric positive semidefinite matrices. Without a shift I expect this number to be even larger.
3. For checking convergence, You could check the off-diagonal norm every n steps.
4. Calling `dispose()` manually makes backpropagation over the SVD impossible. Maybe there is a way to check wether or not backpropagation is desired?
5. *(Minor point)* Computing the SV by computing the EV of **A**<sup>T</sup>**A** or **AA**<sup>T</sup> computes the squares of the SV, which would create a minor risk of underflow (`Math.fround(1e-23**2) == 0`). SVD is often used to determine the rank or to compute rank-deficient linear least square problem. In both cases underflow matters.
6. *(Moot point)* The QR Decomposition is already not the fastest in TFJS. Building a ~2n operation on top of it would mean ~5 minutes for a [100,100] matrix and ~1.2 hours for a [200,200] matrix according to my estimations.

JasonShin:
`The PR is about creating a new operator tf.svd that will return the SVD decomposition of a given matrix. ` I support this

DirkToewe:
QR decomposition, QR iteration and (Francis implictly shifted) QR method, are three distinct algorithms in my understanding (see: Linear Algebra Handbook, chap. 42 "Symmetric Matrix Eigenvalue Techniques"). If I understand @kedevked's code correctly, the QR iteration is used to solve the symmetric positive semidefinite eigenvalue problem **AA**<sup>T</sup> to then solve the SV problem of **A** with it. Only one of my points was addressed to the QR decomposition (the one about performance). The other points are all directed at the implementation of `svd_()` and `eingen_()` in the PR.

Here are the lines of code that make me believe, QR iteration is used:
```js
// ...
function svd_(m: Tensor2D): {u: Tensor, s: Tensor, v: Tensor} {
  const mT = m.dot(transpose(m));
  const u = eingen(mT).vectors;
// ...
```

```js
// ...
function eingen_(m: Tensor): {values: Tensor1D, vectors: Tensor} {
  let z;
  for (let i = 0; i < 5; i++) {
    const [x, y] = linalg.qr(m);
    m = y.dot(x);
    z = z ? z.dot(x) : x;
    y.dispose();
  }
// ...
```
`eingen_` looks like QR iteration without shift.

A few more points that came to my attention looking at the PR for the fourth time:
  * `eingen(mT)` is called twice in `svd_()`.
  * In an SVD **U** and **V** are both orthonormal. So `solve_()` is not necessary since **U**<sup>-1</sup>=**U**<sup>T</sup>
  * `m.dot(transpose(m))` could be replaced by `tf.matMul(m,m, false, true)`

As always, I may very well be mistaken in all of this.

P.S.: To my knowledge, nobody is opposed to having SVD in TFJS afaik, certainly not me. I was hoping that my first comment made that clear.

kedevked:
I will take the time to go through the algorithms for implementation suggested so as to come with a sound implementation. Thanks !

DirkToewe:
For symmetric eigenvalue problems (like the SVD) there is also another, simpler algorithm: The (classic) [Jacobi eigenvalue algorithm](https://en.wikipedia.org/wiki/Jacobi_eigenvalue_algorithm). It also takes O(n⁴) operations and in my opinion it is much simpler to understand and implement than Francis QR or Golub-Reinsch. There's also [O(n³) implementations](http://www.netlib.org/lapack/lawnspdf/lawn170.pdf) of the Jacobi SVD but they are really complicated.

stken2050:
I have a question, for node.

For Python version, svd is already there;
https://www.tensorflow.org/api_docs/python/tf/linalg/svd
and I guess these Python apis are based on C binaries.

tfjs for node is also based on the same C binary, I've heard. So I think it should be trivial to implement all of the apis that are already implemented in Python version, for tfjs for node.

I understand all of the apis of  tfjs on browser is pure js and it seems a long way development to cover basic apis. 

So, here's my question. 

Is there any development policies that apis for both browser and node are synchlonized? If so, even there are exsiting binaries to bind from node, we must wait every api for browser pure js implementation.

If I miss something, plese advice. Thanks.

PR #1256, opened 2018-08-28
add tf.diag

kedevked:
mimic python tf.diag

---
FEATURE https://github.com/tensorflow/tfjs/issues/639

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1256)


kedevked:
I made the changes as suggested. But instead of returning a tensor2d the gpu backend is returning a tensor1d that is reshaped in `diag.ts`.
I have added more tests with tensor3d and tensor4d

kedevked:
With output of 2D, we cannot compute the diagonal of tensor whose dimension is greater than 1. The issue that rises is the following: if the output is 2D, we only have to compare 2 coordinates, but when the tensor that is returned should be a 4D, 6D, 8D, ... we respectively have to compare 4, 6 and 8 coordinates. We cannot simply do it with getOutputCoords on a 2d output. That is why I wrote two helper functions `indextoCoords` and `coordstoIndex`.
`indextoCoords` will take the index of the 1D input and give his coordinates given the shape of the initial tensor and `coordstoIndex ` is useful to getX at a coordinate.

kedevked:
Everything is fixed now, I think.

dsmilkov:
CLA is ok because the other commits are from me (merging with master)

kedevked:
Thanks  @dsmilkov for merging this code. There is also the related operator [tensorflow/tfjs#655](tensorflow/tfjs#655)

PR #1259, opened 2018-08-30
Batch matmul

annxingyuan:
RE: https://github.com/tensorflow/tfjs/issues/539

This PR changes matmul so that it accepts any two tensors of rank >= 2. This mimics the behavior of Tensorflow in other languages. 

The `matmul` op now passes its inputs to the kernels as 3D tensors, collapsing the batch dimensions together and creating a batch dimension of 1 if necessary. The kernels perform straightforward matmul on each batch element and return a 3D tensor, which is reshaped to its original dimensions before being returned.

#### Benchmark results 

I ran `matmul` and downloaded the result 200 times for matrices of sizes 1, 100, 400, and 1000 across this branch and `master`. As we would expect, the results were very similar:

`master`: 7, 7, 12, 96 (miliseconds per operation)
`batch_matmul`: 7, 7, 10, 95

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1259)


nsthorat:
About to review, but just wanted to show you the travis build. If you look at the bottom of the PR you'll see this (build is failing):

image

If you click on "details" then click on whichever node version is failing (in this case both), you'll see the problem.

You can run the linter locally with `yarn lint`.

nsthorat:
Also I think you can set up Sublime with clang-format so that some of these things like line length are done automatically.

The clang-format style we use is "Google", which you can see here: https://github.com/tensorflow/tfjs-core/blob/master/.vscode/settings.json#L24

PR #1265, opened 2018-08-31
Share tensors across backends

dsmilkov:
Internal change that allows tensors to be shared seamlessly across backends, and allow users to switch backends in the middle of an execution.

There are no public APIs in this PR.

Details:

Before this change, switching a backend would re-create a new engine and lose all the previously created tensors. Now, we have only 1 global engine, that is responsible to track in which backend each tensor lives. When backends come across a `dataId` that they don't own, they notify the engine by calling `engine.moveData(dataId)`, which will trigger the engine to fetch the tensor from its current backend, remove it, and register it with the new backend.

- Moving tensors between backends is synchronous (uses `readSync`), which simplifies the world.
- Errors of the type `No data found for this tensor. Did you change your backend in the middle of the program?` are not longer possible.

FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1265)


dsmilkov:
Removed the public API from this PR making this an internal-only change. PTAL.

PR #1269, opened 2018-09-05
Check Ops: Adds assertEqual to the API

manrajgrover:
#### Description

This PR adds `tf.assertEqual` to the API. Fixes https://github.com/tensorflow/tfjs/issues/638.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1269)

### DIFF
--- a/src/ops/check_ops.ts
+++ b/src/ops/check_ops.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


nsthorat:
Apologies for this review -- I just saw the TensorFlow docs which do this. I'm a little skeptical at all that we need this, or that the use case that would be used would fit the task (it's quite easy to check this yourself with a `.data()`).

Also it looks like the python version doesn't support eager anyways, so I'm inclined to close this out if that's okay with you!

nsthorat:
@pyu10055 would we ever support assert ops in the converter? It seems not very useful.

dsmilkov:
Hi Manraj,

After some discussion we decided not to add tf.assertEqual() for now. See https://github.com/tensorflow/tfjs/issues/638 for more info.

Thank you for jumping on this and apologies for making you do this work!

PR #1271, opened 2018-09-06
Fix ArgMinMaxProgram bestIndex initialization

pvaneck:
#### Description
Currently, bestIndex is initialized to zero, but should be set to inOffset to match how bestValue is initialized.

This fixes issues when an array is split into multiple windows, and the first value of a non-first window happens to be the largest value.

Fixes https://github.com/tensorflow/tfjs/issues/665

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1271)

### DIFF
--- a/src/kernels/webgl/argminmax_gpu.ts
+++ b/src/kernels/webgl/argminmax_gpu.ts
@@ -44,7 +44,7 @@ export class ArgMinMaxProgram implements GPGPUProgram {
         int outIdx = coords[1];
         int inOffset = outIdx * ${windowSize};
 
-        int bestIndex = 0;
+        int bestIndex = inOffset;
         float bestValue = getA(batch, inOffset);
 
         for (int i = 0; i < ${windowSize}; i++) {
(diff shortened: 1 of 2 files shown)


chrisdonahue:
Could anyone review?

src/kernels/webgl/argminmax_gpu.ts
```diff
@@ -44,7 +44,7 @@ export class ArgMinMaxProgram implements GPGPUProgram {
         int outIdx = coords[1];
         int inOffset = outIdx * ${windowSize};
 
-        int bestIndex = 0;
+        int bestIndex = inOffset;
```
manrajgrover: This would need a test which you can add [here](https://github.com/tensorflow/tfjs-core/blob/master/src/ops/reduction_ops_test.ts#L415)

pvaneck: Hey, thanks for the suggestion. I went ahead an added a test case for both argMin and argMax which this change affects.

pvaneck:
@tafsiri Hey, a couple tests were added to cover the areas where the bug is hit.

pvaneck:
@tafsiri Good suggestion. I went ahead and added the change.

PR #1278, opened 2018-09-17
Specify dtypes of both arguments in multiply

Lewuathe:
#### Description

MISC

As indicated in [the multiply kernel in `backend_cpu`](https://github.com/tensorflow/tfjs-core/blob/master/src/kernels/backend_cpu.ts#L443), we should also specify the dtype of both argument in `backend_gpu.`

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1278)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -617,7 +617,7 @@ export class MathBackendWebGL implements KernelBackend {
   }
 
   multiply(a: Tensor, b: Tensor): Tensor {
-    if (a.dtype === 'complex64') {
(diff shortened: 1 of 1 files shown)


nsthorat:
We already assert strict types matching here, which is why I didn't do this: https://github.com/tensorflow/tfjs-core/blob/master/src/ops/binary_ops.ts#L318

I'm curious why you were seeing this

Lewuathe:
@nsthorat Thanks so I'm going to close this. I took a look which ops now supports `complex64` dtype.

PR #1279, opened 2018-09-17
added support prod reduce op

pyu10055:
#### Description

implement python [reduce_prod](https://www.tensorflow.org/api_docs/python/tf/reduce_prod) op
ref https://github.com/tensorflow/tfjs/issues/692

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1279)

### DIFF
--- a/src/kernels/webgl/reduce_gpu.ts
+++ b/src/kernels/webgl/reduce_gpu.ts
@@ -65,7 +65,8 @@ export class ReduceProgram implements GPGPUProgram {
       if (${reduceType === 'sum'}) {
         sumValue += dot(values, ones);
       } else if (${reduceType === 'prod'}) {
-        prodValue *= values[0] * values[1] * values[2] * values[3];
+        vec2 tmp = vec2(values[0], values[1]) * vec2(values[2], values[3]);
+        prodValue *= tmp[0] * tmp[1];
       } else {
         minMaxValue = ${compareOp}(values, minMaxValue);
       }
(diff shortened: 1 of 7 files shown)


src/kernels/webgl/reduce_gpu.ts
```diff
@@ -59,6 +64,8 @@ export class ReduceProgram implements GPGPUProgram {
     let updateSnippet = `
       if (${reduceType === 'sum'}) {
         sumValue += dot(values, ones);
+      } else if (${reduceType === 'prod'}) {
+        prodValue *= values[0] * values[1] * values[2] * values[3];
```
yhwang: I happened to trace the code here recently. since the `updateSnippet` would be part of the program code, directly performing the necessary calculation without if/else condition may improve the performance, especially `updateSnippet` is inside the for loop.

so the GPGPUProgram would look like this:
```
        for (int i = 0; i < 0; i += 4) {                                        
          int inIdx = inOffset + i;                                             
          vec4 values = vec4(                                                   
            getValue(batch, inIdx),                                             
            getValue(batch, inIdx + 1),                                         
            getValue(batch, inIdx + 2),                                         
            getValue(batch, inIdx + 3)                                          
          );                                                                    
                                                                                
          prodValue *= values[0] * values[1] * values[2] * values[3];  // only the calculation but no if/else         
        } 
```

pyu10055: @yhwang thanks for the suggestion, curious, have you done any measure on this improvement?

PR #1281, opened 2018-09-20
Prevent TypeError of 'arraysEqual()' for null arguments

cimfalab:
#### Description
Prevent possible errors in 'arraysEqual()' when arguments are null.
For example, in the src/kernels/webgl/gpgpu_math.ts, 'texShapeB' is null in case of 'input.isUniform' is true:
```
    const texShapeB = input.isUniform ? null : input.texData.texShape;
    if (!util.arraysEqual(texShapeA, texShapeB)) {
```
Then 'arraysEqual()' fails when accessing 'n2.length'.

* So I added some checks prior to length check.
* I found from this. Please see it: https://deepscan.io/dashboard/#view=project&pid=3218&bid=27161&subview=issues&status=%5B%22fixed%22%5D

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1281)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -163,6 +163,13 @@ export function isScalarShape(shape: number[]): boolean {
 }
 
 export function arraysEqual(n1: FlatVector, n2: FlatVector) {
+  if (n1 === n2) {
+    return true;
+  }
+  if (n1 == null || n2 == null) {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


nsthorat:
Thanks for the PR!

PR #1284, opened 2018-09-21
Add benchmark suite in benchmark test

Lewuathe:
#### Description

MISC

It added the following benchmark suites. 
- Batch normalization
- Pooling operation
- Reduction operation
- Unary operation

You can only run the specific benchmark suite with `--grep`.

```
$ yarn benchmark --grep='unary(avg)'
```

The default benchmark suites without any `--grep` option was unchanged. In short, `matmul`, `conv2d` and `mobilenet_v1` will run.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1284)

### DIFF
--- a/integration_tests/benchmarks/README.md
+++ b/integration_tests/benchmarks/README.md
@@ -10,6 +10,12 @@ When running benchmarks locally, use:
 
 `yarn benchmark`
 
+You can only run the specific benchmark suite with `--grep`.
+
+`yarn benchmark --grep='pool(avg)'`
+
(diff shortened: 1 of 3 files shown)


Lewuathe:
Okay. I just want to compare the performance of other ops (batchnormalization etc) because it must be useful when we try to improve these corresponding ops. And also the benchmark implementations for these ops are already available. I thought it is better to make use of them in benchmark tool.

PR #1293, opened 2018-10-03
feat(NativeScript) Add support for NativeScript

shiv19:
#### Description
These changes will allow TFJS to be used inside a NativeScript App

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1293)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -28,7 +28,8 @@ declare namespace java {
 declare var CACurrentMediaTime: () => number;
 interface NativeScriptGlobal {
   android?: {};
-  ios?: {};
+  NSObject?: {};
+  NSString?: {};
 }
 
 /** Shuffles the array using Fisher-Yates algorithm. */
@@ -526,7 +527,7 @@ export function now(): number {
     const g = global as NativeScriptGlobal;
     if (g.android) {
       return java.lang.System.nanoTime() / 1000000;
-    } else if (g.ios) {
+    } else if (g.NSObject && g.NSString) {
       return CACurrentMediaTime();
     }
   }
--- a/src/util.ts
+++ b/src/util.ts
@@ -17,6 +17,20 @@
 
 import {ArrayData, DataType, DataTypeMap, FlatVector, RecursiveArray, RegularArray, TensorLike, TypedArray} from './types';
 
+/** Declaring variables for NativeScript support */
+declare namespace java {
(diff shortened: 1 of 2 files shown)


shiv19:
Here is the POC
https://github.com/shiv19/nativescript-tfjs-poc

jlooper:
So exciting! Hope to see this sweet PR merged so we can tinker with TF.js!

src/util.ts
```diff
@@ -508,6 +508,10 @@ export function now(): number {
   } else if (typeof process !== 'undefined') {
     const time = process.hrtime();
     return time[0] * 1000 + time[1] / 1000000;
+  } else if (global && global.android !== 'undefined') {
+    return java.lang.System.nanoTime() / 1000000;
+  } else if (global && global.ios !== 'undefined') {
```
NathanaelA: There is no `global.ios` -- I've tried to get NS team to add one to make this simpler for us plugin developers -- but it doesn't exist; you would have to do something like `if (global && global.NSObject  && global.NSString) {` to determine if you are on iOS.

shiv19: Ah! Let me update that. `global.android` is present, right?

nsthorat:
It's great to see this, though I don't think we can maintain this in core since it's a little specific to a single execution environment which we don't have expertise in.

I have a suggestion, though, is this something we could do outside of the library? You can easily override `util.now()`, or `performance.now()` and patch in global as window.

shiv19:
Hi @nsthorat 
Thanks for your suggestion,
like you said, I was able to get it working with changes made outside of the library.
Now I can just create a NativeScript plugin to do the same.
I'll close this PR now. Thanks for the idea :)

I wrote this code to get it working outside of Tfjs

```
global.setImmediate = setTimeout;
global.clearImmediate = clearTimeout;
global.performance = {
  now() {
    if (global.android) {
      return java.lang.System.nanoTime() / 1000000;
    } else {
      return CACurrentMediaTime();
    }
  }
};
global.process = {
  versions: {
    node: '10.0'
  }
};
```

It works fine, but there could be unintended consequences later. I'll update this thread if I find any such thing.

nsthorat:
Great! Yes definitely let me know. If we try to understand what hooks we need for other things like react native we may be able to provide an API to set some of this stuff more explicitly.

PR #1295, opened 2018-10-03
always return from uploadtogpu if texture is already on gpu

annxingyuan:
This fixes a bug where we failed to return out of uploadToGpu if `WEBGL_PAGING_ENABLED` was true. 

BUG

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1295)

### DIFF
--- a/src/debug_mode_test.ts
+++ b/src/debug_mode_test.ts
@@ -21,7 +21,7 @@ import {convertToTensor} from './tensor_util_env';
 import {ALL_ENVS, expectArraysClose} from './test_util';
 
 describeWithFlags('debug on', ALL_ENVS, () => {
-  beforeEach(() => {
+  beforeAll(() => {
     tf.ENV.set('DEBUG', true);
   });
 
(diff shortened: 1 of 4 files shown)


annxingyuan:
I think WebGL 1 was failing because when we call `ENV.reset` (e.g. in `environment_test`) in a `describeWithFlags` block, we don't preserve the features of the test environment - they only get revived by `executeTests` (in `jasmine_util`). So when the environment was reset after the first test in the `WEBGL_PAGING_ENABLED` test set, the WebGL version was reset to 2 instead of 1. I think this caused the matMul failure - our pipeline might have had different ideas about which WebGL version it was using at different points, so it may have created `RGBA` textures but then tried to read them as `RGBA32F` or something. 

To fix this I'm passing in the test environment to the `describeWithFlags` block, so individual tests can reset to the test environment features if they want. Is there a better way?

annxingyuan:
@dsmilkov 

I left one `ENV.reset` call in `environment_test`, in the `Backend` test set towards the end of the file. These tests are in a `describe` block so they don't get the `beforeEach`/`afterEach` environment reset treatment. Let me know if you want me to handle this differently.

PR #1301, opened 2018-10-08
Add a PROD environment flag

dsmilkov:
Add a PROD environment flag to enable "production" environment.

Production flags are widespread among JS libraries. Enabling this flag will turn off various safety checks that can cause a slowdown in performance. The user can enable this flag once the application/model is correctly working and there are no visible bugs. Enabling this flag should result in a performance boost.

To set the flag, the user calls `ENV.set(‘PROD’, true);`. Internally, we guard computation behind specific flags. e.g. `if(ENV.get(‘WEBGL_PAGING_ENABLED’)) { … }`, which are evaluated based on the PROD flag.

Initially, when the flag is enabled, the following functionalities will be turned off:
- Deep shape validation when creating a tensor from a native array.
- Paging in the WebGL backend.

We will gradually add more functionalities behind this flag as we profile more real models.

Also:
- Replace the condition `util.sizeFromShape(input.shape) <= SIZE_UPLOAD_UNIFORM && ENV.get('WEBGL_RENDER_FLOAT32_ENABLED')` with a `WEBGL_SIZE_UPLOAD_UNIFORM` flag in order to make it easier to predict when that code path gets executed since it's behind an explicit flag.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1301)


dsmilkov:
Ah, good question. Adding the `WEBGL_SIZE_UPLOAD_UNIFORM` flag in unrelated to the `PROD` flag. (haven't seen any perf impact of uploading textures vs small uniforms). This is a cleanup to replace the condition `util.sizeFromShape(input.shape) <= SIZE_UPLOAD_UNIFORM && ENV.get('WEBGL_RENDER_FLOAT32_ENABLED')` with a specific flag in order to make it easier to predict when that code path gets hit.

I updated the PR description accordingly.

PR #1307, opened 2018-10-10
support SetDiff1D op

pyu10055:
This is same op as ListDiff, since the output shape of the op is variable, and due to its
simplicity, it is implemented as an async op with cpu implementation only.
[tf python api](https://www.tensorflow.org/api_docs/python/tf/setdiff1d)

fixes https://github.com/tensorflow/tfjs/issues/692

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1307)


nsthorat:
async kernels turned out to be super complex so let's drop it in favor of what you're doing here (not calling the kernel for async functions). Once you fix the comments addressed and respond we can go ahead and merge this.

PR #1328, opened 2018-10-17
Preserve TextureUsage when caching on the CPU if it was previously set to PACK

annxingyuan:
This PR changes `cacheOnCPU` so that the texture usage is preserved if it was previously `TextureUsage.PACK`. 

This will be useful for implementing lazy packing / unpacking as part of packing batchNormalization (https://github.com/tensorflow/tfjs/issues/774). When `DEBUG` is on, we call `cacheOnCPU` for every texture in order to check for NaN's. This sets the `TextureUsage` of the texture to `UPLOAD`, regardless of its original `TextureUsage`. This makes it difficult to lazily unpack while debugging. 

BUG

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1328)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -1803,7 +1803,8 @@ export class MathBackendWebGL implements KernelBackend {
       texData.texture = null;
       texData.texShape = null;
     }
-    texData.usage = TextureUsage.UPLOAD;
+    texData.usage =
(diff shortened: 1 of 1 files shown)


annxingyuan:
This is no longer an issue since https://github.com/tensorflow/tfjs-core/pull/1329 separated `usage` from `isPacked`

PR #1329, opened 2018-10-17
Lazily pack and unpack textures in WebGL kernels

annxingyuan:
### Changes
- Add `packedInput` boolean property to GPGPUProgram to identify WebGL kernels that require packed inputs (`batchMatMul`, `conv2DWithIm2Col`).
- Such kernels no longer unpack their outputs, and they only pack their inputs if they are not already packed.
- In `compileAndRun`, check whether the GPGPUProgram requires packed inputs. If it does not, and an input texture is packed, unpack it before passing it to the program.

PERF

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1329)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -1611,7 +1611,7 @@ export class MathBackendWebGL implements KernelBackend {
   }
 
   private packTensor<T extends Tensor>(x: T): T {
-    if (this.texData.get(x.dataId).usage === TextureUsage.PACK) {
+    if (this.texData.get(x.dataId).packed) {
       return x;
     }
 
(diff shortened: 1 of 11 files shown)


annxingyuan:
@nsthorat 

If I dispose the input tensors to a kernel immediately after packing / unpacking them, gradient operations stop working because the inputs to `engine.runKernel` are disposed by the time they get passed to the tape node. 

Maybe it would make sense to have the gradient scope dispose of the tensors instead? But I'm sure how to approach implementing that - is there a straightforward way that takes advantage of existing infrastructure that you know of?

annxingyuan:
Note - this PR now depends on https://github.com/tensorflow/tfjs-core/pull/1327 because the batchNorm tests need to be able to print 4D packed outputs

PR #1331, opened 2018-10-18
Add complex64 support for tf.abs().

nsthorat:
We hide this behind tf.abs.

`y = sqrt(real ^ 2 + imag ^ 2)`

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1331)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-core",
-  "version": "0.13.4",
+  "version": "0.13.6",
   "description": "Hardware-accelerated JavaScript library for machine intelligence",
   "private": false,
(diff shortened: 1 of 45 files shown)


DirkToewe:
There's also a [length()](https://www.khronos.org/files/opengles_shading_language.pdf#page=74) function in GLSL ES 1.00. It might be slightly faster and also avoid some fringe cases of float32 underflow ( `float32( (1e-23)² ) = 0` ).

dsmilkov:
That's a great point @DirkToewe. Want to send a PR?

DirkToewe:
[Here You go](https://github.com/tensorflow/tfjs-core/pull/1391)

PR #1332, opened 2018-10-19
Add tf.math.confusionMatrix

caisq:
FEATURE

Fixes https://github.com/tensorflow/tfjs/issues/771

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1332)


nsthorat:
+1 to usability over performance (making it sync is better than async & a minor performance win).

PR #1336, opened 2018-10-23
Pin dependencies to exact versions.

nsthorat:
Pin depenencies to the ones that work (webgl-ext is not working at 0.0.30).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1336)


nsthorat:
Dropping this in favor of https://github.com/tensorflow/tfjs-core/pull/1337

PR #1342, opened 2018-10-25
Support LogSoftmax op

yhwang:
Althought LogSoftmax is log(Softmax) but the implementation is not.
Add the LogSoftmax op and use the implementation from tensorflow.

fixes [issue 684](https://github.com/tensorflow/tfjs/issues/684)
Signed-off-by: Yihong Wang <yh.wang@ibm.com>

#### Description

---


yhwang:
I used the log_softmax implementation in tensorflow [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/softmax_op_functor.h#L67-L81) and use the gradient function [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/nn_grad.py#L247-L262)

I haven't added test cases yet. Any suggestion for the test cases? Can I duplicate the test cases for softmax and update the expect results accordingly?

I directly put the logSoftmax() with softmax in `softmax.ts`. Should I separate them?

yhwang:
@pyu10055 thanks for the comments. I followed the `softmax` and came up this PR. In logSoftmax it needs `exp().sum().log()` operations, same as softmax in tfjs-core. I guess the reason that original softmax impl doesn't put those calculation into WebGL is the `sum()` calculation in the middle. Therefore, I didn't try to move everything to WebGL. but that's my guess. Also the LogSoftmax implementation in tensorflow [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/softmax_op_functor.h#L67-L81) seems has no specialization for GPU. Any suggestion to move this to WebGL and enhance the performance?

nsthorat:
@pyu10055 lets wait till the questions are all addressed before merging next time :)

yhwang:
@nsthorat @pyu10055 thanks for the reviews again. One question, may I also work on `tfjs-converter` to put `logSoftmax` into normalization ops?

nsthorat:
@yhwang go for it!

PR #1345, opened 2018-10-25
Unpack eagerly within packed ops that reshape their outputs

annxingyuan:
This PR is a quick fix to https://github.com/tensorflow/tfjs/issues/834

### Changes
- Eagerly unpack outputs of `matMul` and `im2Row` so the final reshape occurs on the unpacked tensor
- Add a test that catches the issue

BUG

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1345)


annxingyuan:
@caisq We need to figure out a way to unpack lazily so adjacent packed ops don't have to unpack / pack their outputs / inputs unnecessarily. But there are no performance implications for now because only matMul is packed.

PR #1347, opened 2018-10-25
Make loadWeights independent of fetch

justadudewhohacks:
#### Description

I want to suggest using dependency injection to inject the actual weight fetching function into loadWeights. This should solve the issues discussed in https://github.com/tensorflow/tfjs/issues/479 and https://github.com/tensorflow/tfjs/issues/489.

This way nodejs users could load their models from the local filesystem:

``` javascript
const fetchWeightsFromDisk = async (filePaths: string[]) => {
  return filePaths.map(filePath => fs.readFileSync(filePath).buffer)
}
const loadWeights = tf.io.loadWeightsFactory(fetchWeightsFromDisk)
const manifest = JSON.parse(fs.readFileSync('./my_model-weights_manifest').toString())
const weightMap = loadWeights(manifest, './')
```

Also people who don't use the node-fetch polyfill to inject their own fetching logic, for example using axios:

``` javascript
const fetchWeightsAxios = async (fetchUris: string[]) => {
  return Promise.all(fetchUris.map(uri=> axios.get(uri, { responseType: 'arraybuffer' })))
}
```

I hope this change is reasonable to you.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1347)


justadudewhohacks:
Just a small update: I had to explicitly set the return type of the factory function due to a bug in typescript 2.9, which emits an import statement with local filepath into the typings file, when the return type should be resolved automatically for some reason.

Didn't notice it earlier since I am using typescript 2.8.4.

justadudewhohacks:
Thanks for the review @tafsiri, I made the changes regarding your and @caisq's comments. For now I left the @doc decorator and only added a tsdoc description with example. Can be added later to the docs, if you want to hold this for now.


Regarding your comment:
>  so we want to hold off on making this public until we have a chance to do that

Are you referring to merging the change in general or to making the documentation public? Would be good to know for me, in case you are planning to refactor the function before publishing the change, since I am looking to bring nodejs support to face-api.js soon and would need this in order to easily load models with quantization from disk.

If you still need some time on this just let me know, no problem. In this case I will simply implement a temporary solution, using decodeWeights and later patch it.

tafsiri:
@justadudewhohacks i meant holding off on adding it to the public docs (which is what the @doc annotation does). We definitely agree that this is needed, but we may refactor it before adding to the public docs. In the mean time this can get in and be part of the next release imo since we don't have an eta on that. 

My comment was more that this api may change change in future and that would affect face-api.js when upgrading down the line. I think we should ultimately bring this functionality to the public api but need to consolidate between this and the IOhandlers and some of the other code we have around loading weights and models.

justadudewhohacks:
Alright got, thanks for clarification. :)

I am aware that this is not part of the public API yet, so if the API is going to change, that shouldn't be a problem and I will make the required upgrades to face-api.js.

PR #1355, opened 2018-10-26
WIP Change KernelBackend from an interface to a class

dsmilkov:
This allows all Backends to not break whenever we add a new method in core. Instead all backends get a default "Not implemented" implementation.

This allows us to run Travis nightly on `tfjs-node` and get a log with all the failing unit tests without being prematurely terminated by a compiler build error.

DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1355)

### DIFF
--- a/src/kernels/backend.ts
+++ b/src/kernels/backend.ts
@@ -83,231 +83,496 @@ export interface BackendTimer {
  * methods, this can be done gradually (throw an error for unimplemented
  * methods).
  */
-export interface KernelBackend extends TensorStorage, BackendTimer {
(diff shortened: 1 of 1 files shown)


nsthorat:
I'm not completely sure I follow, but you can create a class that forces you to meet the exact interface of the backend, which is what we'll do for the WebGL backend and the CPU backend:

```
public class NikhilBackend implements KernelBackend {
}
```

or the following for Node, which doesn't force you to implement everything:

```
public class NickBackend extends KernelBackend {
}
```

PR #1371, opened 2018-11-01
WebGL backend should hand off kernel execution to CPU if inputs are small and on the CPU

annxingyuan:
This PR addresses https://github.com/tensorflow/tfjs/issues/826

### Changes
- Added a `shouldExecuteOnCPU` method to the `MathBackendWebGL` that checks whether inputs are on the CPU and within a size threshold
- Singled out certain ops (`stridedSlice`, `slice`, `concat`, `min`, `max`, `subtract`, `multiply`, `greater`, `less`) for passing kernel execution to the CPU if conditions are met. These are the ops that get called most frequently in COCO SSD with tiny inputs. 

This PR is not productionized (no tests or environment flag) - I wasn't sure whether we wanted to just experiment with the approach first.

### COCO SSD benchmark

on `master`: 277ms
on `webgl_cpu_handoff`: 184ms *(+50.5%)*

These are prediction times across 20 runs.

#### Note

I think there's a better way to write this so we don't have to repeat the checking logic in so many places. If we just create an array of kernel names we want to apply this treatment to, then we could loop through them and forward to the CPU using *arguments*. And if we could get a list of all the kernels then we could apply the treatment across the webGL backend. But it might require some TypeScript overrides...

---

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -645,10 +645,6 @@ export class MathBackendWebGL implements KernelBackend {
   }
 
   multiply(a: Tensor, b: Tensor): Tensor {
-    if (this.shouldExecuteOnCPU([a, b])) {
-      return this.cpuBackend.multiply(a, b);
-    }
-
     if (a.dtype === 'complex64') {
       const aData = this.texData.get(a.dataId);
       const bData = this.texData.get(b.dataId);
@@ -671,6 +667,10 @@ export class MathBackendWebGL implements KernelBackend {
       real.dispose();
       imag.dispose();
       return complex;
+    }
+
+    if (this.shouldExecuteOnCPU([a, b])) {
+      return this.cpuBackend.multiply(a, b);
     }
 
     const program = new BinaryOpProgram(binaryop_gpu.MUL, a.shape, b.shape);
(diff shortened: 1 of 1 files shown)


src/kernels/backend_webgl.ts
```diff
@@ -128,6 +129,7 @@ export interface TensorHandle {
   dtype: DataType;
 }
 
+const CPU_HANDOFF_SIZE_THRESHOLD = 10;
```
pyu10055: add explanation on this hard-coded number.

nsthorat:
@pyu10055 I thought about this even more, and I actually believe we won't be able to move this out of the WebGL backend code, even if we do some smart backend allocation down the road (which I still think we should, though it's lower priority because there isn't a concrete use-case).

The WebGL-backend knows more information than we can ever pass up to the engine, namely where the data actually is stored physically -- the tensor data can be in the WebGL backend, but on the CPU as a Float32Array or the GPU as a WebGLTexture. The actual physical location is not available outside the backend because that complexity is hidden by the WebGL backend. For example, we wouldn't want the engine to know about paging logic which can move data between the CPU and the GPU within the WebGL backend.

pyu10055:
> @pyu10055 I thought about this even more, and I actually believe we won't be able to move this out of the WebGL backend code, even if we do some smart backend allocation down the road (which I still think we should, though it's lower priority because there isn't a concrete use-case).
> 
> The WebGL-backend knows more information than we can ever pass up to the engine, namely where the data actually is stored physically -- the tensor data can be in the WebGL backend, but on the CPU as a Float32Array or the GPU as a WebGLTexture. The actual physical location is not available outside the backend because that complexity is hidden by the WebGL backend. For example, we wouldn't want the engine to know about paging logic which can move data between the CPU and the GPU within the WebGL backend, but the WebGL backend should be able to make decisions about where to run operations that depend on how paging logic has moved data around.

I agree what you said that WebGL has more detail information about location of the tensors, but it does not conflict with the decision to give up the execution of the op. The engine does not need to know the details, it just need to push the execution to the next available backend.

The coupling makes more sense to me if the cpu backend is a child of webgl backend and hidden from the engine. Since we have the new syntax of run on a specific backend now, this type of forwarding can be surprising if the user specifically want to run the op on a particular backend.

nsthorat:
I think we can move forward with this change.

PR #1372, opened 2018-11-01
Add integration test with layers/node/converter when we modify `version.ts`

dsmilkov:
Run a 3rd travis job (integration) when the `src/version.ts` file is modified.

The integration job clones `layers`/`node`/`converter` and tests those repos against `core@master`, giving the PR author a heads up before they make a new release of `core`.

https://travis-ci.org/tensorflow/tfjs-core/builds/449534806 is an example of a Travis run. See the 3rd job (TO_TEST=integration).

status

Details:
 - rename `yarn test-travis` to `yarn run-browserstack` to better reflect the task
 - move all travis-related scripts under `yarn test-travis` to align core with the other repos (layers & node).

Towards tensorflow/tfjs#848

DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1372)


dsmilkov:
cc @tafsiri @annxingyuan

PR #1377, opened 2018-11-05
Remove node 8 from travis since 10 is LTS

dsmilkov:
Even though we officially support node 8 as well (package.json stays the same), we only test the latest LTS on Travis.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1377)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
-node_js:
-  - '8'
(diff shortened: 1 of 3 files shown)


tafsiri:
Just FYI node 8 is still an active LTS (they overlap) https://github.com/nodejs/Release

PR #1379, opened 2018-11-05
Speedup matmul when the matrices are vectors

dsmilkov:
In the WebGL backend, we can parallelize dot products by calling `tf.mul(x,y).sum()` -- this is due to the fact that the WebGL's sum() uses divide-and-conquer and is O(sqrt(N)).

Have matmul() in WebGL call `tf.mul(x, y).sum()` when we have matrix x vector, vector x matrix, or vector x vector and the shared dimension is longer than 1000 (see benchmark below).

Fixes https://github.com/tensorflow/tfjs/issues/881

Benchmark of `[1,K]x[K,1]` for different `K`s:
=== OLD ===
LOG: 'K = 1, Took 0.72 ms'
LOG: 'K = 10, Took 0.67 ms'
LOG: 'K = 100, Took 0.59 ms'
LOG: 'K = 1000, Took 0.64 ms' --- slower after K>1000.
LOG: 'K = 10000, Took 1.18 ms'
LOG: 'K = 100000, Took 10.66 ms'
LOG: 'K = 1000000, Took 110.90 ms'

=== NEW ===
LOG: 'K = 1, Took 0.62 ms'
LOG: 'K = 10, Took 0.62 ms'
LOG: 'K = 100, Took 0.67 ms'
LOG: 'K = 1000, Took 0.67 ms' --- faster after K>1000.
LOG: 'K = 10000, Took 1.12 ms'
LOG: 'K = 100000, Took 0.88 ms'
LOG: 'K = 1000000, Took 0.94 ms'

Benchmarked mobilenet, coco-ssd, and posenet --  no perf impact there.

PERF 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1379)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -133,6 +133,10 @@ export interface TensorHandle {
 const BEFORE_PAGING_CONSTANT = 300;
 // Tensors with size <= than this will be uploaded as uniforms, not textures.
 export const SIZE_UPLOAD_UNIFORM = 4;
+// Empirically determined minimal shared dimension in matmul before we forward
+// to a.mul(b).sum() in order to take advantage of GPU parallelism. See
+// https://github.com/tensorflow/tfjs-core/pull/1379 for benchmarks.
+const MATMUL_SHARED_DIM_THRESHOLD = 1000;
 
 export class MathBackendWebGL implements KernelBackend {
   private texData: DataStorage<TextureData>;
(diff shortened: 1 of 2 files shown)


dsmilkov:
Hold on for a bit, I'm going to generalize this further for any matrix-vector, vector-matrix product based on the excellent observation of @DirkToewe in https://github.com/tensorflow/tfjs/issues/881#issuecomment-436045897

dsmilkov:
This is ready for review

PR #1381, opened 2018-11-07
Fix failing test in travis

dsmilkov:
https://github.com/tensorflow/tfjs-core/pull/1379 added a unit test that failed in travis due to numerical precision: https://travis-ci.org/tensorflow/tfjs-core/jobs/451654174#L801

This PR:
- fixes that test
- Also fixes `test-travis.sh` to make sure we don't forward webgl --> cpu except for the last browser run -- probably caused during conflict resolving of https://github.com/tensorflow/tfjs-core/pull/1371 and https://github.com/tensorflow/tfjs-core/pull/1372 (https://reviewable.io/reviews/tensorflow/tfjs-core/1371#-LQ_a2y92YYLRMiYWqhC)

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1381)

### DIFF
--- a/scripts/test-travis.sh
+++ b/scripts/test-travis.sh
@@ -29,11 +29,11 @@ if [[ $(node -v) = *v10* ]]; then
 
   # Run the rest of the karma tests in parallel. These runs will reuse the
   # already downloaded binary.
-  npm-run-all -p -c --aggregate-output \
-    "run-browserstack --browsers=bs_safari_mac --features '{\"HAS_WEBGL\": false}' --backend cpu" \
-    "run-browserstack --browsers=bs_ios_11 --backend webgl --features '{\"WEBGL_CPU_FORWARD\": false}'" \
-    "run-browserstack --browsers=bs_ios_11 --features '{\"HAS_WEBGL\": false}' --backend cpu" \
-    "run-browserstack --browsers=bs_firefox_mac" \ # Implicitly has WEBGL_CPU_FORWARD: false
-    "run-browserstack --browsers=bs_chrome_mac" \ # Implicitly has WEBGL_CPU_FORWARD: false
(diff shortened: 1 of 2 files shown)


nsthorat:
Looks like those bash comments are screwing things up, my guess is because of the `\` happening before the comment.

DirkToewe:
Out of curiosity: Why is the Safari JSVM behaving differently? The results seem to be way of. Even if summed up naively:
```js
Array.from({ length: 10*1000 }, () => Math.fround(0.1**2) )
     .reduce( (x,y) => Math.fround(x+y), /*initVal=*/0.0 )
```
The result on my machine is `100.0029525756836`. But if I am not mistaken the CPU matrix multiplication in TFJS is blocked up. Inside the blocks, a float64 value is used to sum the values up. So the result should be something along the lines of:
```js
const blockSize = 50;
Array.from(
  { length: 10*1000/blockSize },
  () => Math.fround(
    Array.from({ length: blockSize }, () => Math.fround(0.1**2) )
         .reduce( (x,y) => x+y, /*initVal=*/0.0 )
  )
).reduce( (x,y) => Math.fround(x+y), /*initVal=*/0.0 )
```
That result even perfectly rounds to 100 on my machine.

In the Travis log the result seems to be `99.29019927978516` which would be way off.

dsmilkov:
I noticed the same - Didn't give much thought since that test was not failing on my local laptop (Macbook Pro). The only diff between my laptop and Travis was the Safari version: mine is 12.0.1, Travis used 11.1.2

DirkToewe:
I too had some issued with the Safari tests on Travis. My guess was that it was a WebGL thing, but maybe there is an issue with `Float32Array` in that Webkit/Safari version?

PR #895, opened 2018-03-25
UnaryOps: Add dl.round

Lewuathe:
# Purpose

Add `dl.round`. As [tf.round](https://www.tensorflow.org/api_docs/python/tf/round) does, the algorithm is banker's rounding. 

see: https://github.com/PAIR-code/deeplearnjs/issues/894 and https://github.com/tensorflow/tfjs/issues/70

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/895)

PR #1391, opened 2018-11-14
Made tf.abs underflow-safe on CPU & GPU and added test for it.

DirkToewe:
#### Description

Fixes part of [#895](https://github.com/tensorflow/tfjs/issues/895). The CPU backend now uses `Math.hypot` and the WebGL backend uses `length` albeit with some workaround to ensure underflow-safety (which is otherwise not given, at least on Intel GPUs).

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1391)

### DIFF
--- a/src/kernels/backend_cpu.ts
+++ b/src/kernels/backend_cpu.ts
@@ -1122,7 +1122,7 @@ export class MathBackendCPU implements KernelBackend {
     for (let i = 0; i < x.size; ++i) {
       const real = values[i * 2];
       const imag = values[i * 2 + 1];
-      resultValues[i] = Math.sqrt(real * real + imag * imag);
+      resultValues[i] = Math.hypot(real, imag);
     }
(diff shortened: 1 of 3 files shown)


DirkToewe:
Done, hope I didn't forget anything.

DirkToewe:
`eps²` will not underflow, but I tried to improvise...

nsthorat:
Hm, I know what the issue is (we're uploading this as a uniform) so I'll merge this change and then fix it :)

nsthorat:
Really nice work!

DirkToewe:
Thank You!
What issue are You referring to? The 0/0 was in `complex_abs_gpu` and I hope I fixed it. My other statement about `eps²` generally not rounding down to zero is not a bug. What would be necessary is another constant like `ENV.get('MIN_VALUE')` to determine a value that close to underflow.

PR #1393, opened 2018-11-15
Disable uniform uploads when testing on Travis

dsmilkov:
- Disable uniform uploads when testing on travis.
- Change the default epsilon for 16bit float devices to 1e-4 instead of 1e-3 and change the clip test to stop testing `1-epsilon < 1` since the IEEE 754 buckets around 1 are different than around 0.
- Bring back node 8 testing since the LTS is for another 14 months. And test cpu backend in both node envs.

DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1393)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
-node_js: '10'
+node_js:
+  - '8'
+  - '10'
 script:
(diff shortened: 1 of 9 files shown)


annxingyuan:
I see - that makes sense - thanks!

PR #1396, opened 2018-11-15
Add trendline to visualize model performance over time, call dataSync for every run

annxingyuan:
Related issue: https://github.com/tensorflow/tfjs/issues/713

### Changes
- Added a trendline to visualize the speed of each inference so we can get a more granular sense of model performance. This is helpful in the case that, for example, an optimization introduces overhead for the first several inferences, but leads to better model performance in the long run. Average prediction time itself can vary significantly depending on `numRuns`.
- Changed `measureAveragePredictTime` to call `dataSync` for every run, rather than only once at the end. 

Here's what the trendline looks like for MobileNet v2:
screen shot 2018-11-15 at 10 37 13 am
As you can see it takes about 10 runs to stabilize. The average inference time is 23ms, best inference time is 20ms, worst inference time is 36 ms.

DEV, BUG

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1396)

### DIFF
--- a/integration_tests/benchmarks/benchmark.html
+++ b/integration_tests/benchmarks/benchmark.html
@@ -40,8 +40,14 @@
       font-size: 10px;
     }
 
-    #trendline-container {
+    #trendline-container svg {
+      overflow: visible;
+      margin-top: 20px;
+    }
 
+    #trendline-container path {
+      fill: none;
+      stroke: #222;
     }
 
     #modal-msg {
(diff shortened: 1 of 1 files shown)


integration_tests/benchmarks/benchmark.html
```diff
@@ -193,19 +246,20 @@ <h2>TensorFlow.js Model Benchmark</h2>
       tbody.appendChild(tr);
     }
 
+    async function awaitResult(res) {
+      if (res instanceof Promise) {
+        return await res;
```
pyu10055: if the returen type is Promise<Tensor>, you would need call await (await res).dataSync() ?

nsthorat:
Just wanted to put a comment on this since we spoke on chat: it would be good to understand why we're seeing a backoff here. The first pass through we do shader compilation / texture upload, there shouldn't be a speed up after the 3rd run (unless we're competing for resources, or there's some other optimization the browser is doing we've never seen before).

PR #1399, opened 2018-11-17
Add BrowserFiles.save method to save model to File objects

dsanders11:
#### Description

I don't love the way this implementation panned out (so very open to suggestions), but it was the simplest way I could add this functionality with minimal changes.

Goal is to provide a way to save the model as `File` objects rather than triggering a browser download, allowing for further control. For example, I have metadata about the model (author, description, etc) that I'd like to download as well. Downloading a bunch of different files begins to be a bad user experience, so I bundle them into a ZIP file first and then download the single ZIP file. With this PR it becomes:

```javascript
const { files } = await model.save(tf.io.BrowserFiles)

// Add files to ZIP, trigger download of ZIP
```

There's several changes in this PR I thought I'd highlight individually:
* Factor out common code for creating `File` objects from `ModelArtifacts` into `getModelArtifactsAsFiles` util function
* Refactor `SaveResult` to use sub-interfaces for handling the different return values for different `IOHandler` implementations
* Add `BrowserFiles.save` method and expose `BrowserFiles` class so that it can be used with `model.save`. Note that this implementation of `save` is a *static* method as the constructor of `BrowserFiles` expects files for loading.

It's the last change I don't love, however I wasn't able to come up with a cleaner implementation. It could possibly be implemented as an entirely different `IOHandler` which implements the save functionality and registers a new scheme (`browser-files://`?), but that seemed to be straying from the existing implementation.

Thoughts and suggestions welcome.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1399)

### DIFF
--- a/src/io/browser_files.ts
+++ b/src/io/browser_files.ts
@@ -21,9 +21,10 @@
  */
 
 import {ENV} from '../environment';
-import {basename, concatenateArrayBuffers, getModelArtifactsInfoForJSON} from './io_utils';
+
+import {basename, concatenateArrayBuffers, getModelArtifactsAsFiles, getModelArtifactsInfoForJSON} from './io_utils';
 import {IORouter, IORouterRegistry} from './router_registry';
-import {IOHandler, ModelArtifacts, SaveResult, WeightsManifestConfig, WeightsManifestEntry} from './types';
+import {BrowserFilesSaveResult, IOHandler, ModelArtifacts, SaveResult, WeightsManifestConfig, WeightsManifestEntry} from './types';
 
 const DEFAULT_FILE_NAME_PREFIX = 'model';
 const DEFAULT_JSON_EXTENSION_NAME = '.json';
@@ -59,33 +60,20 @@ export class BrowserDownloads implements IOHandler {
   }
 
   async save(modelArtifacts: ModelArtifacts): Promise<SaveResult> {
-    const weightsURL = window.URL.createObjectURL(new Blob(
-        [modelArtifacts.weightData], {type: 'application/octet-stream'}));
-
     if (modelArtifacts.modelTopology instanceof ArrayBuffer) {
       throw new Error(
           'DownloadTrigger.save() does not support saving model topology ' +
           'in binary formats yet.');
     } else {
-      const weightsManifest: WeightsManifestConfig = [{
-        paths: ['./' + this.weightDataFileName],
-        weights: modelArtifacts.weightSpecs
-      }];
-      const modelTopologyAndWeightManifest = {
-        modelTopology: modelArtifacts.modelTopology,
-        weightsManifest
-      };
-      const modelTopologyAndWeightManifestURL =
-          window.URL.createObjectURL(new Blob(
-              [JSON.stringify(modelTopologyAndWeightManifest)],
-              {type: 'application/json'}));
+      const [modelTopologyFile, weightDataFile] = getModelArtifactsAsFiles(
+          modelArtifacts, this.modelTopologyFileName, this.weightDataFileName);
 
       // If anchor elements are not provided, create them without attaching them
       // to parents, so that the downloaded file names can be controlled.
       const jsonAnchor = this.jsonAnchor == null ? document.createElement('a') :
                                                    this.jsonAnchor;
-      jsonAnchor.download = this.modelTopologyFileName;
-      jsonAnchor.href = modelTopologyAndWeightManifestURL;
+      jsonAnchor.download = modelTopologyFile.name;
+      jsonAnchor.href = window.URL.createObjectURL(modelTopologyFile);
       // Trigger downloads by calling the `click` methods on the download
       // anchors.
       jsonAnchor.click();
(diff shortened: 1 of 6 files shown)


dsanders11:
> You mentioned adding a browser-files:// URL scheme. But I don't see how that might work, given that you can't point to a File object from a string?

I was referring to using that as a 'fake' URL scheme in the same way as 'downloads://' is currently used, as a way to tell `model.save` to trigger a download. So from my original example, it would look like:

```javascript
const { files } = await model.save('browser-files://my-model-1')
```

This has the benefit that you can provide the filename.

> Do you know of a way?

You can actually create URLs to File objects (well, blobs, which they are) using `window.URL.createObjectURL`, but the resulting URL is generated (starts with `blob:`) and you have no control over it. Here's a [JSFiddle](https://jsfiddle.net/tqa5dvym/) showing a quick demo. I'm not in any way suggesting TFJS use this (I haven't thought through how it could be easily used), just pointing it out for completeness.

> This new method should live in browser_files.ts, as it uses the browser-specific File class, which is not available in Node.js. Also, once it is moved there, you won't need to export it anymore.

It's needed in `browser_http.ts` as well, so export would still be necessary.

> Is this related to the browser File objects change? If not, can you open a separate PR for this change?

It's related in the sense that I needed to add `files` as a return from `model.save`, and the options were to make it an optional key on `SaveResult` like `responses` or to change the pattern to sub-interfaces of `SaveResult` rather than cramming more optional keys into it. I'll make it a separate PR for the moment so its merits can be discussed separately.

> Sorry I didn't include this in the comments earlier. But do you foresee adding support for loading a model from browser File objects in the future?

Er, doesn't that already exist?

```javascript
tf.loadModel(tf.io.browserFiles([modelJSONFile, weightsFile]))
```

That's why I mentioned the ugly asymmetry between loading and saving.

@caisq, I think I'm going to hold off on making changes for the moment and try to ponder the API some more and how to make it cleaner. The nice thing about `model.save` is it can take an `IOHandler` so it's easy to roll your own on the spot, which is what I did for my own usage at the moment. If `getModelArtifactsAsFiles` was part of the public API it would be less fragile to roll your own handler to save to `File`.

dsanders11:
@caisq, what if the public API looked like this:

```javascript
// Save to files
const { files } = model.save(tf.io.browserFiles('my-model'));

// Then load from files
tf.loadModel(tf.io.browserFiles(files));
```

Or, I'd prefer there weren't a wrapping function and just exposing the `BrowserFiles` class itself, as it makes it more clear that any `IOHandler` can be provided.

```javascript
// Save to files
const { files } = model.save(new tf.io.BrowserFiles('my-model'));

// Then load from files
tf.loadModel(new tf.io.BrowserFiles(files));
```

Or if the different arguments for the constructor isn't clean enough:

```javascript
// Save to files
const { files } = model.save(new tf.io.SaveBrowserFiles('my-model'));

// Then load from files
tf.loadModel(new tf.io.LoadBrowserFiles(files));
```

Could also ditch the `tf.io.browserHTTPRequest` helper and `downloads://` scheme in favor of:

```javascript
const { files } = model.save(new tf.io.BrowserFiles('my-model'));

// Send the HTTP request yourself, using whatever method you
// want, and whatever extra data you want. For example, you may
// want to provide a metadata file.

// Generic download files method, allowing extra files to be added
// to the download process, like a metadata file.
tf.io.browserDownload(files);
```

dsanders11:
> So I thought about it a little more. Looking at the API you proposed above (such as const { files } = model.save(tf.io.browserFiles('my-model'));), I don't particularly like the changing the return type of Model.save() to make it a union between SaveResult and File[]. This return type should be agnostic about the browser vs. Node.js environment.

It wouldn't be a union. It's a derived interface of `SaveResult`, `BrowserFilesSaveResult`. That was the reason for #1407 which you had me pull out into a separate PR.

> The API I think I'm starting to like is something like this:

I'm not really a fan, as now there are two different return values, the result of `model.save` and the separate `getFiles`. I think it adds further complexity to the public API, which IMO is already a bit confusing with the use of factory functions. As I mentioned at the end of my previous comment, I think it would make the API cleaner to drop the factory functions, and opt for giving examples of how users can achieve advanced used cases (downloading extra files, extra HTTP requirements, send over RTCDataChannel to a peer, etc). There just needs to be *some* API for getting saved model as a JavaScript object that can be worked with.

If greater parity with Node.js is desired, perhaps an `IOHandler` which saves to/loads from `ArrayBuffer`. It's trivial to turn that into a `File` object (it's one of the types accepted by the constructor) and I believe Node.js has native support for `ArrayBuffer`. The tricky thing with this is there needs to be emphasis put on filename, since the name of the weights file is written into the JSON file, which makes it brittle if you don't save the weights file with the expected name.

dsanders11:
> Do you plan to make it accept both File[] and string as the input argument

Yes, that seems like a reasonable way to go. Each method, `load` and `save` can throw an error if the object was not the constructed with the correct type.

caisq:
This PR is stale and has quite a few conflicts. I'll close it for now. @dsanders11, we are migrating the tfjs-core to a the repo at https://github.com/tensorflow/tfjs, where it'll be consolidated with other TensorFlow.js repos such as tfjs-layers and tfjs-converter. If you'd like, please reopen a PR after the migration is done. Thanks.

PR #1401, opened 2018-11-19
Use tslib and es2017 target when possible

alippai:
#### Description
This improves the usage of tfjs a little bit:
- Allows more control over the build chain in the apps depending on tfjs.
- Helps debugging by not shipping ES5 code for NodeJS (as NodeJS 8 supports ES2017)

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1401)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@types/seedrandom": "2.4.27",
     "@types/webgl-ext": "0.0.30",
     "@types/webgl2": "0.0.4",
-    "seedrandom": "2.4.3"
+    "seedrandom": "2.4.3",
+    "tslib": "^1.9.3"
   }
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -41,11 +41,11 @@ function minify() {
   return uglify({output: {preamble: PREAMBLE}});
 }
 
-function config({plugins = [], output = {}, external = []}) {
+function config({plugins = [], output = {}, external = [], target = 'es5'}) {
   return {
     input: 'src/index.ts',
     plugins: [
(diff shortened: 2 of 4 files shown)


nsthorat:
Hi @alippai, thanks for the PR!

Just curious about how this helps you. Are you using a Node debugger, or just looking at the standard Node stack trace? if you use this the `source-map-support` package as mentioned [here](http://2ality.com/2015/04/node-es6-transpiled.html) you can see the TypeScript stack trace in our code.

When you say "more control", what exactly do you mean? I see a small benefit of this change which is bundle size (targeting newer JavaScript versions typically results in less code), but not much of a change to build tooling.

alippai:
Using the sourcemap support module definitely solves that issue.
The code size improves by a marginal amount, in a real world app 1-2% at most.

For in-browser development&debug I usually use rollup watch without sourcemaps, so I noticed the space for improvement there. 

It's just feels cleaner to ship and use more modern JS when possible. If the change breaks anything it may not worth merging it. Think of this as a code style / nit picking grade PR.

nsthorat:
What this would potentially change is consumers of the library that rely on a bundler. If for some reason they don't have a transpiler (writing es5 but bundling) then we'd be generating es2017 code for them without transpilation, potentially breaking a real es5 target.

I definitely agree it feels cleaner, but right now we've decided to stick to ES5 targets as to target as many devices as possible out of the box. I think we may come back to this at some point when we focus deeply on bundle size.

We really appreciate the PR and thinking through the problem :)

alippai:
Thanks for the detailed explanation. I understand the decision, maybe I'd call the same :)

nsthorat:
:) going to close this out!

PR #1404, opened 2018-11-19
Use specific config types for (de)serialization, not just ConfigDict

davidsoergel:
BREAKING
FEATURE

This is for discussion only at this point.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1404)


bileschi:
I'm guessing this is no longer relevant, given the serialization changes in layers (@tensorflow/tfjs-layers#419)

nsthorat:
What did we decide on this PR? Can we close this out? If we want to do this in the future let's file an issue and then point to the closed out issue.

davidsoergel:
I think the approach in this PR is valuable but I don't have bandwidth to revisit it currently.  Closing for now & mentioning on tensorflow/tfjs#1082.  I'll revive the effort when possible.

PR #1408, opened 2018-11-20
Add string dtype to Tensor

dsmilkov:
Add `string` dtype to `Tensor`.

This opens the door for adding Python's [`string_ops`](https://www.tensorflow.org/api_docs/python/tf/strings) to TensorFlow.js, which are used for text-based models, and for adding pre-processing layers that operate on strings.

Details:
- dtype was not added as a generic to the Tensor class in order to keep compiler errors simple and code backwards compatible.
- dataSync() can be optionally typed to cast its result. E.g. `t.dataSync<'string'>()` returns `string[]` while `t.dataSync()` returns `TypedArray` for backwards compatibility.
- `layers` and `converter` pass with this build. `node` has 30ish failed tests since `string` is an unknown dtype.
- Only `clone`, `reshape` and `cast` work with strings at this point to keep this PR small. Other ops will get the functionality in a follow-up PR.
- Added unit tests to assert that numeric ops throw on string tensors.
- Backends now should support dtype `string` in their `register/write/read` methods.
- Added a vscode config to do debugging directly from vscode

FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1408)


src/engine_test.ts
```diff
@@ -361,6 +361,18 @@ describeWithFlags('memory', ALL_ENVS, () => {
     expect(sum.dtype).toBe('int32');
     expectArraysClose(sum, [1 + 1 + 0 + 1]);
   });
+
+  it('string tensor', () => {
+    const a = tf.tensor([['a', 'bb'], ['c', 'd']]);
+
+    expect(tf.memory().numTensors).toBe(1);
+    expect(tf.memory().numBytes).toBe(10);  // 10 letters, each 2 bytes.
```
pyu10055: 5 letters? Do we always store string type as UTF-16, do we support UTF-8?

src/kernels/backend_cpu.ts
```diff
@@ -678,7 +678,7 @@ export class MathBackendCPU implements KernelBackend {
     this.assertNotComplex(x, 'logicalNot');
 
     const values = x.dataSync();
-    const newValues = new Int32Array(values.length);
+    const newValues = new Uint8Array(values.length);
```
pyu10055: does it support int32 logicalNot?

PR #1412, opened 2018-11-27
Implement packed clipping op.

annxingyuan:
This PR addresses https://github.com/tensorflow/tfjs/issues/803

### Changes
- Added `ClipPackedProgram` that implements clipping for packed textures
- Added ENV flag `WEBGL_PACK_CLIP`, defaults to `false`
- Added ENV flag `WEBGL_PACK`, defaults to `false`, that turns all packing-related flags on or off
- Turn off NaN checking if `PROD` flag is on
- Support scalar sampling in packed ops
- Partially support packed `get${Var}AtOutCoords` and broadcasting - broadcasting only occurs when the innermost dimensions are the same and the vector to be broadcast is rank 0 or 1. 

PERF

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1412)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -311,6 +311,10 @@ export class Environment {
       return true;
     } else if (feature === 'WEBGL_PACK_BATCHNORMALIZATION') {
       return false;
+    } else if (feature === 'WEBGL_PACK_CLIP') {
+      return false;
+    } else if (feature === 'WEBGL_PACK_DEPTHWISECONV') {
+      return false;
     } else if (feature === 'WEBGL_LAZILY_UNPACK') {
       return false;
     } else if (feature === 'WEBGL_CONV_IM2COL') {
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -309,16 +309,16 @@ export class Environment {
       return isChrome();
     } else if (feature === 'WEBGL_CPU_FORWARD') {
       return true;
+    } else if (feature === 'WEBGL_PACK') {
+      return false;
     } else if (feature === 'WEBGL_PACK_BATCHNORMALIZATION') {
-      return false;
+      return this.get('WEBGL_PACK');
     } else if (feature === 'WEBGL_PACK_CLIP') {
-      return false;
-    } else if (feature === 'WEBGL_PACK_DEPTHWISECONV') {
-      return false;
+      return this.get('WEBGL_PACK');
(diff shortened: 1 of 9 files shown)


src/environment.ts
```diff
@@ -308,14 +308,18 @@ export class Environment {
...
+    } else if (feature === 'WEBGL_PACK_BATCHNORMALIZATION') {
+      return this.get('WEBGL_PACK');
+    } else if (feature === 'WEBGL_PACK_CLIP') {
+      return this.get('WEBGL_PACK');
     } else if (feature === 'WEBGL_PACK_DEPTHWISECONV') {
-      return false;
+      return this.get('WEBGL_PACK');
     } else if (feature === 'WEBGL_LAZILY_UNPACK') {
-      return false;
+      return this.get('WEBGL_PACK');
```
astojilj: Minor issue: might be useful to note that this won't work if trying to turn on packing by calling set('WEBGL_PACK', true) from application code; it is still required to call set('WEBGL_LAZILY_UNPACK', true) since tf-layers initialization evaluates ENV.get('EPSILON') before application code, and causes evaluating and setting WEBGL_LAZILY_UNPACK to false.   
I'm aware that this should be done, properly, using URL query string parameters.  

Call stack:
Environment.evaluateFeature (environment.ts:299)
Environment.get (environment.ts:270)
MathBackendWebGL.compileAndRun (backend_webgl.ts:1988)
MathBackendWebGL.abs (backend_webgl.ts:1396)
(anonymous) (backend_webgl.ts:2029)
(anonymous) (engine.ts:156)
Engine.scopedRun (engine.ts:167)
Engine.tidy (engine.ts:153)
Environment.tidy (environment.ts:186)
MathBackendWebGL.floatPrecision (backend_webgl.ts:2028)
Environment.evaluateFeature (environment.ts:368)
Environment.get (environment.ts:270)
(anonymous) (common.ts:15)
extendStatics (tf-layers.js:20)
(anonymous) (tf-layers.js:21)

annxingyuan: Hey @astojilj - that's a good point - could you create a separate PR for this change?

astojilj: > Hey @astojilj - that's a good point - could you create a separate PR for this change?

I reported this so that you are aware of it when testing. Otherwise, I'm not sure what would be the desired behavior - setting master switch WEBGL_PACK would override all previously set sub-flags?

PR #1416, opened 2018-11-28
Update core version to 0.14.0

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1416)


nsthorat:
cc @nkreeger FYI the new release will break Node.js tests because of the string dtype change. Other than that the integration tests all pass.

PR #1420, opened 2018-11-30
getTexureShapeFromLogicalShape should return even numbers for packed …

astojilj:
…texures

#### Description
Ensure that even width and height is returned, even though it could exceed
maxTexSize - e.g. 6x12 shape is sizeToSquarishShape-ed to 8x9.

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1420)

### DIFF
--- a/src/kernels/webgl/webgl_util.ts
+++ b/src/kernels/webgl/webgl_util.ts
@@ -375,7 +375,14 @@ export function getTextureShapeFromLogicalShape(
       logShape[1] * logShape[2] * logShape[3] <= maxTexSize) {
     return [logShape[0], logShape[1] * logShape[2] * logShape[3]];
   } else {
-    return util.sizeToSquarishShape(size);
+    const texShape = util.sizeToSquarishShape(size);
+    if (isPacked) {
+      // Ensure that packed texture has even width and height, even it could
+      // exceed maxTexSize - e.g. 6x12 shape is sizeToSquarishShape-ed to 8x9.
+      texShape[0] = util.nearestLargerEven(texShape[0]);
+      texShape[1] = util.nearestLargerEven(texShape[1]);
+    }
+    return texShape;
   }
 }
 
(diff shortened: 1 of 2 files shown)


astojilj:
@annxingyuan, PTAL

astojilj:
@annxingyuan 
Thanks.
>The texShape returned by getTextureShapeFromLogicalShape actually refers to the number of rows and columns that the texture will need to accommodate, not its physical texture width and height

I found the logic that packed tex shape dims are half of computed texShape width and height logical and intuitive. It is obvious from the code and documentation.

Understand your point and see how this cannot get triggered. 

Let's check the [`uploadToGPU`:](https://github.com/tensorflow/tfjs-core/blob/596ca2575b7310ace0ba0e56e1caa347203e46a3/src/kernels/backend_webgl.ts#L1980) 
```

    const texShape =
        webgl_util.getTextureShapeFromLogicalShape(shape, isPacked);
    texData.texShape = texShape;
    const newTexture = this.acquireTexture(dataId, texShape, usage, isPacked);
    texData.texture = newTexture;
    if (values != null) {
      // TODO(smilkov): Propagate the original typed array to gpgpu.
      if (isPacked) {
        const batch = this.getBatchDim(shape);
        let rows = 1, cols = 1;
        if (shape.length) {
          [rows, cols] = this.getRowsCols(shape);
        }
        this.gpgpu.uploadMatrixToPackedTexture(
            newTexture, batch, rows, cols, texShape[0], texShape[1],
            typedArrayToFloat32(values as Float32Array));
      } else {
```

getPackedMatrixTextureShapeWidthHeight is called within acquireTexture and uploadMatrixToPackedTexture call tree. Replacing replace ceil(width/2) by width/2 in getPackedMatrixShapeWidthAndHeight and make the sizes even in getTextureShapeFromLogicalShape would also affect getKeyFromTextureShape. Closing the PR as unnecessary.

PR #1422, opened 2018-12-01
Enable passing custom fetch function to BrowserHTTPRequest constructor

caisq:
Towards fixing https://github.com/tensorflow/tfjs/issues/410

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1422)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -57,5 +57,4 @@ To install it locally, run `yarn add ./tensorflow-tf-core-VERSION.tgz`.
 
 > On Windows, use bash (available through git) to use the scripts above.
 
-Looking to contribute, and don't know where to start? Check out our "help wanted"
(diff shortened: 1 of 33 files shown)


caisq:
@nsthorat For context: this is the first of two PRs that will fix the issue in which tfjs-node users have to override `global.fetch` in order to load models via http/https.

nsthorat:
My 2c is that passing a function is more canonical JS than inheritance (and inheritance usually is more confusing anyways). I don't care that much though.

caisq:
@nsthorat @tafsiri I have updated the PR to add an additional argument to the constructor of `BrowserHTTPRequest` and the factory function `browserHTTPRequest`. This has the additional nice property that we need to export one fewer symbol.

Thanks for the suggestion. PTAL.

PR #1425, opened 2018-12-03
Remove the global outer scope.

nsthorat:
This causes a memory leak in backend_cpu and Node.js because we hold a pointer to all Tensors.

I verified with htop that memory usage does not explode linearly and that GC kicks in.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1425)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -97,7 +97,7 @@ export class Engine implements TensorManager, DataMover {
 
   // Keep Tensors that parallel the tapes.
   private activeScope: ScopeState;
-  private scopeStack: ScopeState[];
+  private scopeStack: ScopeState[] = [];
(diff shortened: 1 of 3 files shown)


nsthorat:
There isn't really anything I can write that the existing unit tests wouldn't catch.

nsthorat:
The bug is that the program holds pointers to all the Float32Arrays, we cannot access the page memory to understand that it is blowing up. `tf.memory()` does not reflect the change here because that tracks explicit memory management with `dispose()` and `tidy()`.

nsthorat:
Added a test that is outside our testing mechanism to test that allocation of tensors succeeds outside a scope (does not test anything about memory management).

PR #1430, opened 2018-12-05
Allow squeeze to use negative index

VisibleMarkov:
Issue in focus: tensorflow/tfjs#940

#### Description

`squeeze()` didn't support the use of negative index which is supported in tensorflow for Python. Added tests for the same in `util_test` and `tensor_test`.
The previous implementation didn't allow for random arrangement of values of axis. For example if axes is set to `[2, 3]`, it would remove them both but if you pass it as `[3, 2]` it would only remove 3 and ignore 2 entirely. I tried the same with tensorflow for python in this [Google Colab Notebook](https://colab.research.google.com/drive/1jzyuRFXfIVrK50jI7vDrVuW5dYpiYw0V) and this too was inconsistent with tfjs as tensorflow python allows for removal of axis irrespective of order of input if it's allowed. This inconsistency was removed when adding the support for negative index ( it was totally unintentional but resulted from my implementation to allow negative index ).

Clarification required for following topics:
* I am not sure whether the test messages are satisfactory for the new tests added - Please let me know if this can be improved and I'll implement the same as soon as possible.
* The random input order support was a result of my judgement where i thought `[-1, -2]` was intuitively much better for a programmer but if we just replaced it with the positive index of same, it would have been `[5, 4]` which due to it's lack of ascending order would have ignored 4 completely. If this needs to changed, please let me know and I'll come up with another implementation to change this behavior.

Tests Ran:
* [x] yarn test
* [x] yarn lint

BUG

---

Tags: BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1430)

### DIFF
--- a/src/tensor_test.ts
+++ b/src/tensor_test.ts
@@ -1372,6 +1372,16 @@ describeWithFlags('tensor', ALL_ENVS, () => {
     expect(() => a.squeeze([-3, -2])).toThrowError();
   });
 
+  it('squeeze axis out of range', () => {
+    const a = tf.tensor3d([4, 2, 1], [3, 1, 1], 'bool');
+    expect(() => a.squeeze([10, 11])).toThrowError();
+  });
+
+  it('squeeze negative axis out of range', () => {
+    const a = tf.tensor3d([4, 2, 1], [3, 1, 1], 'bool');
+    expect(() => a.squeeze([-13, -12])).toThrowError();
+  });
+
   it('squeeze throws when passed a non-tensor', () => {
     expect(() => tf.squeeze({} as tf.Tensor))
         .toThrowError(/Argument 'x' passed to 'squeeze' must be a Tensor/);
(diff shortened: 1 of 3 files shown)


VisibleMarkov:
I signed it!

VisibleMarkov:
* Added an exception case in case the given value of axis doesn't lie in `[ -shape.length, shape.length)` which further makes it consistent with tensorflow for python.
The behavior of Python in the case can be found in this [Google Colab Notebook](https://colab.research.google.com/drive/1jzyuRFXfIVrK50jI7vDrVuW5dYpiYw0V)

Seeking Clarification on following issue:
* The error message was more than 80 characters long and I suppressed TSist using `disable-line`. Please let me know if this can be replaced with a shorted message, or another approach altogether

Tests Ran:
* [x] yarn test
* [x] yarn lint

VisibleMarkov:
@caisq Addressed the changes requested in 02d2344
If anything needs to be changed, please let me know and I'll implement it as soon as I can (^_^)

VisibleMarkov:
Thanks for the green flag (^_^)
Just curious: Should I wait for another review by someone else?

caisq:
@VisibleMarkov PR is merged. Thank you for your contribution!

PR #1434, opened 2018-12-07
Simplify logic for allocating array for packed textures. 

annxingyuan:
### Changes

- Replaced an inappropriate use of `getUnpackedArraySizeFromMatrixSize` with simpler logic for computing the size of a packed texture array.
- Removed an error check in `encodeMatrixToPackedRGBA` that ensures the output array is big enough for the data because we use the same error checking logic in the construction of the output array. This cuts down the number of use cases for `getPackedMatrixTextureShapeWidthHeight`.
- Modified `getTextureShapeFromLogicalShape` to return 2 as channel height for 1D inputs, and to change the packed `size` input to `sizeToSquarish` such that channel dimensions remain even in squarification.
- Fixed the packed broadcasting logic to account for squarified textures and added a unit test.
- Modified im2col unit test to account for new squarification logic.

DEV, BUG, PERF

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1434)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -1806,10 +1806,13 @@ export class MathBackendWebGL implements KernelBackend {
 
   private packedReshape<R extends Rank>(input: Tensor, afterShape: ShapeMap[R]):
       Tensor<R> {
-    const inputAs3D = input.reshape(
-        [webgl_util.getBatchDim(input.shape), ...webgl_util.getRowsCols(input.shape)]);
-    const afterShapeAs3D =
-        [webgl_util.getBatchDim(afterShape), ...webgl_util.getRowsCols(afterShape)];
+    const inputAs3D = input.reshape([
+      webgl_util.getBatchDim(input.shape),
+      ...webgl_util.getRowsCols(input.shape)
+    ]);
+    const afterShapeAs3D = [
+      webgl_util.getBatchDim(afterShape), ...webgl_util.getRowsCols(afterShape)
+    ];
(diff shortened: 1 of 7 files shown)


annxingyuan:
Hey @dsmilkov, I think this is ready for another look.

- I did not include an error check in `getPackedMatrixTextureShapeWidthHeight` for even inputs because there are unit tests (`gpgpu_context_test.ts:181`) that directly call the method with odd inputs, and I felt those should remain valid use cases. Happy to change the unit tests though - let me know.
- Removed an error check in `encodeMatrixToPackedRGBA` that ensures the output array is big enough for the data because we use the same error checking logic in the construction of the output array. This cuts down the number of use cases for `getPackedMatrixTextureShapeWidthHeight`.
- Modified `getTextureShapeFromLogicalShape` to return 2 as channel height for 1D inputs, and to change the packed `size` input to `sizeToSquarish` such that channel dimensions remain even in squarification.
- Fixed the packed broadcasting logic to account for squarified textures and added a unit test.
- Modified im2col unit test to account for new squarification logic.

I tested squarifying unconditionally in `webgl_util` (except for 1D textures) and all unit tests pass (except for tests of `getTextureShapeFromLogicalShape`).

By the way, the test I was referring to yesterday that ensures packed data is properly encoded even if different batches must occupy the same physical row is in `gpgpu_util_test.ts:180`.

annxingyuan:
@dsmilkov Made a small change - now unconditional squarification works regardless of rank.

PR #1435, opened 2018-12-07
Fix context for window.fetch; Remove unnecessary done() function in test

caisq:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1435)

### DIFF
--- a/src/io/browser_http.ts
+++ b/src/io/browser_http.ts
@@ -46,7 +46,6 @@ export class BrowserHTTPRequest implements IOHandler {
             'browserHTTPRequest is not supported outside the web browser ' +
             'without a fetch polyfill.');
       }
-      this.fetchFunc = fetch;
     } else {
       assert(
           typeof fetchFunc === 'function',
@@ -110,7 +109,7 @@ export class BrowserHTTPRequest implements IOHandler {
           'model.weights.bin');
     }
 
-    const response = await this.fetchFunc(this.path as string, init);
+    const response = await (this.fetchFunc || fetch)(this.path as string, init);
 
     if (response.ok) {
       return {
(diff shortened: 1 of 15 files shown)


src/io/browser_http.ts
```diff
@@ -230,9 +232,20 @@ export class BrowserHTTPRequest implements IOHandler {
...
+
+  /**
+   * Helper method to get the `fetch`-like function set for this instance.
+   *
+   * This is mainly for avoiding confusion with regard to what context
+   * the `fetch`-like function is bound to. In the default (browser) case,
+   * the function will be bound to `window`, instead of `this`.
+   */
+  private getFetchFunc() {
+    return this.fetchFunc || fetch;
```
pyu10055: looks like this.fetchFunc is always set, not the getFetchFunc method is needed.

src/io/browser_http_test.ts
```diff
@@ -59,7 +59,7 @@ const modelTopology1: {} = {
   'backend': 'tensorflow'
 };
 
-describeWithFlags('browserHTTPRequest-load fetch-polyfill', NODE_ENVS, () => {
+describeWithFlags('browserHTTPRequest-load fetch', BROWSER_ENVS, () => {
   let requestInits: RequestInit[];
 
   // simulate a fetch polyfill, this needs to be non-null for spyOn to work
```
pyu10055: this does not match the title of the test.

src/io/browser_http.ts
tafsiri: yep. since you are using the getFetchFunc approach you can just skip binding fetch when assigning it to this.fetchFunc in the constructor. Unless you are also trying to protect against accidental direct usage of this.fetchFunc? In any case with the current setup the || part isn't needed.

caisq:
Comments are addressed. All tests now pass. I tested integration_tests/benchmarks also and confirmed it works, @annxingyuan

annxingyuan:
Thanks @caisq :)

PR #1440, opened 2018-12-09
Replace 'new ImageData' with 'createImageData' in tf.toPixels

teruhisa:
ImageData is not defined in NodeJS.
Instead, call createImageData on CanvasRenderingContext2D to create the
object to set the pixel bytes.

#### Description

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1440)

### DIFF
--- a/src/ops/array_ops.ts
+++ b/src/ops/array_ops.ts
@@ -429,7 +429,8 @@ async function toPixels(
     canvas.width = width;
     canvas.height = height;
     const ctx = canvas.getContext('2d');
-    const imageData = new ImageData(bytes, width, height);
+    const imageData = ctx.createImageData(width, height);
+    imageData.data.set(bytes);
     ctx.putImageData(imageData, 0, 0);
   }
   if ($img !== img) {


nkreeger:
Hi @teruhisa :-)

Did you see the tfjs-node bindings: https://github.com/tensorflow/tfjs-node

We're working on splitting up these browser-specific ops soon. Core TensorFlow already has an API we need to surface for that in the tfjs-node package.

This is an interesting fix - let's discuss @dsmilkov @nsthorat

nsthorat:
toPixels is now under tf.browser.toPixels so this won't be the method to use in Node.js. I think we should close this PR. We can add another nodejs specific API separately (and likely bind to C++).

Nick, should we file an issue to track that?

nsthorat:
Either way, I verified this works by checking out the branch (so seems fine to merge). @teruhisa, can you update this to master and then sign the CLA and we can merge?

teruhisa:
Sure, I'll make the update when I'm back.

teruhisa:
Looks like the method was removed by #1590.
I might check out the node variant of tfjs library as nkreeger@ mentioned when I get the chance.
Given that this change is obsolete, closing the PR.

PR #1441, opened 2018-12-10
Fix condition for expensive reshape with even columns

astojilj:
Packed reshape when both, source and destination shape, have even
columns is not free - with current 2x2 packing, pixel's .ba values
are the next (logical) row of .rg values. If row length changes,
different values are expected in .ba values.

e.g. 2 -> 4 columns reshape:

```
0 1 4 5    -->  0 1 2 3
2 3 6 7         4 5 6 7
```

Added test that exposes the issue and the fix.

BUG

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1441)


astojilj:
Reduced the size from 32 to 16. 
I would like to keep the texture layout with four logical rows packed to one texture row in this test.

PR #1446, opened 2018-12-12
Align pow gradient with TF

jgartman:
#### Description

Multiple problems with the gradient for pow were brought up in https://github.com/tensorflow/tfjs/issues/346.  One was fixed by #1376, the other is that the gradient with respect to the exponent is nan when the base is negative.  This differs from TF which returns 0 in this case: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py#L1046

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1446)


jgartman:
There was a test I added that was failing because of numerical precision issues and I decided to just remove it.  The failure was because tf.greater is being used to compare the base of an exponent to 0 and on Safari when the base is 0 this comparison was returning true.  I tried comparing the base to a small non-zero values like 1e-4 but the test still failed.  Let me know if there are other suggestions, not sure how useful this PR is if it's not going to work as expected on all platforms.

nsthorat:
Hmm -- unfortunately the merge broke this :/ could you look into this if you have a sec?

jgartman:
Just needed to update a recently added test to match the new behavior

PR #1447, opened 2018-12-12
Fix invalid isPacked state when texture is released.

astojilj:
When the data is read back (unpacked) to CPU and texture is released, keeping
the isPacked flag leads to state when unpacked data is understood as packed.
Added a test that exposes the issue.

BUG 

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1447)


astojilj:
@annxingyuan PTAL

astojilj:
Tests pass but there are issues:
13 12 2018 13:44:57.503:ERROR [reporter.browserlabs]: ✖ Could not update BrowserStack status
Error: 403 Forbidden (Rate Limit Exceeded)

astojilj:
> Tests pass but there are issues:
> 13 12 2018 13:44:57.503:ERROR [reporter.browserlabs]: ✖ Could not update BrowserStack status
> Error: 403 Forbidden (Rate Limit Exceeded)

CI issue fixed.

annxingyuan:
@astojilj 

Ah - nice catch. Although I think it might make more sense to change the logic in `compileAndRun` - `backend_webgl.ts:1879` - to the following: if `texData.texture == null`, then we want to set the `isPacked` bit to equal `program.usesPackedTextures` (no need for the if statement). Does that make sense?

Also I believe this unit test passes at head because the input tensor will be uploaded as a uniform. Could you also set the `WEBGL_SIZE_UPLOAD_UNIFORM` flag to 0 as part of the test setup?

astojilj:
>Ah - nice catch. Although I think it might make more sense to change the logic in compileAndRun - backend_webgl.ts:1879 - to the following: if texData.texture == null, then we want to set the isPacked bit to equal program.usesPackedTextures (no need for the if statement). Does that make sense?

It is a matter of preference - because isPacked sounds like data is packed, and on the other places we don't check (isPacked && texData.texture != null) it looks like a loose end to me that is better to be fixed at the origin: where texture is released.

> Also I believe this unit test passes at head because the input tensor will be uploaded as a uniform. Could you also set the WEBGL_SIZE_UPLOAD_UNIFORM flag to 0 as part of the test setup?

WEBGL_SIZE_UPLOAD_UNIFORM [is set to 0](https://github.com/tensorflow/tfjs-core/blob/a71700bedf5a65b79203a88523e4c27d3e9b9ae8/src/jasmine_util.ts#L110) for all tests now so it fails without the patch. Good point - I'll set to 0 in this test.

astojilj:
>The bad news is that it appears that one or more commits were authored or co-authored by someone other than the pull request submitter. 

Is it because you pressed update branch and I merged upstream/master with that commit?
Is it OK to force push only squashed commit to fix this or you'd prefer other way?

PR #1449, opened 2018-12-12
Optimize concat for the WebGL backend

dsmilkov:
Speedup WebGL concat by 5x without warmup and 2x with shader warmup.

Benchmark used:
```js
const tensors = [];
for (let i = 0; i < 400; i++) {
  tensors.push(tf.ones([1, 696]));
}
const res = tf.concat(tensors, 0 /* axis */);
res.dataSync();
console.log('Took', performance.now() - start);
```

Without warmup: 626ms (this PR) vs 3066ms (master)
With warmup: 34ms (this PR) vs 65ms (master)

The benchmark above reflects a real workflow of preparing training data (stacking 400 examples, 696 values each), taken from the Audio recognition codelab. Measuring both with warmup and without warmup is important since in the use-case of collecting examples, the number of examples is dynamic and causes recompilation of the shaders.

**Details**
- Add a new WebGL flag `WEBGL_MAX_TEXTURES_IN_SHADER` that gives the maximum number of textures we can have as uniform samples in a single shader.
- Change concat_gpu to take arbitrary number of inputs.
- When concating a large number of tensors, do binary concatenation (divide-and-conquer) to reduce number of concat calls.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1449)

### DIFF
--- a/src/environment.ts
+++ b/src/environment.ts
@@ -17,7 +17,7 @@
 
 import * as device_util from './device_util';
 import {Engine, MemoryInfo, ProfileInfo, ScopeFn, TimingInfo} from './engine';
-import {Features, getFeaturesFromURL, getWebGLDisjointQueryTimerVersion, getWebGLMaxTextureSize, isChrome, isDownloadFloatTextureEnabled, isRenderToFloatTextureEnabled, isWebGLFenceEnabled, isWebGLVersionEnabled} from './environment_util';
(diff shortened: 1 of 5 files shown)


dsmilkov:
Good q. A quick search for (`Tensor[]` or `T[]`) in backend.ts gives `addN()` as another candidate.

PR #1455, opened 2018-12-13
add try catch to all await calls for http browser io handler

pyu10055:
fixes https://github.com/tensorflow/tfjs/issues/992
This change would propagate the promise errors and fixes the unhandled promise rejection warnings.

---

### DIFF
--- a/src/io/browser_http.ts
+++ b/src/io/browser_http.ts
@@ -159,58 +159,68 @@ export class BrowserHTTPRequest implements IOHandler {
   }
 
   protected async loadBinaryModel(): Promise<ModelArtifacts> {
-    const graphPromise = this.loadBinaryTopology();
-    const manifestPromise =
-        await this.getFetchFunc()(this.path[1], this.requestInit);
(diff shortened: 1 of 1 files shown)


nsthorat:
Hey can we prioritize this? A few people are asking about it.

nsthorat:
One more thing we should do here is when the server gives an HTML response for shards (e.g. not octet-stream). We get these errors about the number of bytes of the response being off, but we should error with the status code of the server.

nsthorat:
Here is an example:

image

nsthorat:
Here are some proposals for error messages:

- Request to http://.... failed with status code 500 (this would happen for each type of request)
- Request to http://... expected a mime type octet-stream but got ... (this would happen for all requests)
- Request to http:// returned xyz bytes but expected abc bytes.

Here are some proposals from Daniel: https://github.com/tensorflow/tfjs/issues/924#issuecomment-440322136

Thanks Ping!

nsthorat:
Can you also update the PR description / title?

PR #1466, opened 2018-12-20
In DEBUG mode check downloads for Infinity and check uploads for NaN / Infinity / underflow / overflow

annxingyuan:
This PR addresses https://github.com/tensorflow/tfjs/issues/1043

### Changes
- In `DEBUG`, check downloads for Infinity
- In `DEBUG`, check uploads (regardless of type) for NaN, Infinity, underflow / overflow

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1466)

### DIFF
--- a/src/debug_mode_test.ts
+++ b/src/debug_mode_test.ts
@@ -60,7 +60,7 @@ describeWithFlags('debug on', ALL_ENVS, () => {
     expect(a).toThrowError();
   });
 
-  it('debug mode errors when trying to upload values that would underflow', () => {
+  it('debug mode errors when uploading values that would underflow', () => {
     const savedFlag = tf.ENV.get('WEBGL_RENDER_FLOAT32_ENABLED');
     tf.ENV.set('WEBGL_RENDER_FLOAT32_ENABLED', false);
 
--- a/src/debug_mode_test.ts
+++ b/src/debug_mode_test.ts
@@ -45,6 +45,20 @@ describeWithFlags('debug on', ALL_ENVS, () => {
     expect(a).toThrowError();
   });
 
+  it('debug mode errors when infinities in op output', () => {
+    const a = tf.tensor1d([1, 2, 3, 4]);
+    const b = tf.tensor1d([2, -1, 0, 3]);
+    const c = () => a.div(b);
+    expect(c).toThrowError();
+  });
+
+  it('debug mode errors when nans in op output', () => {
+    const a = tf.tensor1d([-1, 2]);
+    const b = tf.tensor1d([0.5, 1]);
+    const c = () => a.pow(b);
+    expect(c).toThrowError();
+  });
+
   it('debug mode errors when nans in oneHot op (tensorlike), int32', () => {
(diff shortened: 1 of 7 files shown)


src/ops/tensor_ops.ts
```diff
@@ -93,6 +106,16 @@ function tensor<R extends Rank>(
   if (!isTypedArray(values) && !Array.isArray(values)) {
     values = [values] as number[];
   }
+
+  if(ENV.get('DEBUG')) {
```
dsmilkov: I'm not comfortable having this logic outside WebGL yet since it only affects WebGL for now. Can you move this to backend_webgl and check values in backend.write()

src/ops/tensor_ops.ts
```diff
@@ -25,6 +25,19 @@ import {assert, assertNonNull, flatten, getArrayFromDType, inferDtype, isTypedAr
 import {complex} from './complex_ops';
 import {op} from './operation';
 
+// https://en.wikipedia.org/wiki/Half-precision_floating-point_format
+const MIN_FLOAT16 = 5.96e-8;
+const MAX_FLOAT16 = 65504;
+
+function canBeRepresented(num: number): boolean {
```
dsmilkov: move this to webgl_util or similar since it's webgl specific

src/debug_mode_test.ts
```diff
@@ -35,10 +35,9 @@ describeWithFlags('debug on', ALL_ENVS, () => {
...
   });
 
-  it('debug mode errors when there are nans, float32', () => {
-    const a = tf.tensor1d([2, NaN]);
-    const f = () => tf.relu(a);
-    expect(f).toThrowError();
+  it('debug mode errors when nans in tensor construction, float32', () => {
+    const a = () => tf.tensor1d([2, NaN], 'float32');
+    expect(a).toThrowError();
   });
```
dsmilkov: add a unit test making sure that even though the user didn't upload any NaNs, but a kernel produced those, it still errors. Another similar test for Infinities. For example do tf.div(a, b) where b has 0. Likewise tf.log(a) where a has zeros

src/util.ts
```diff
@@ -498,18 +495,18 @@ export function toTypedArray(
   if (dtype === 'string') {
     throw new Error('Cannot convert a string[] to a TypedArray');
   }
-  if (noConversionNeeded(a, dtype)) {
-    return a as TypedArray;
-  }
   if (Array.isArray(a)) {
     a = flatten(a as number[]);
   }
+  if (debugMode) {
```
dsmilkov: Seems like you fixed a bug where by moving this check sooner. Can you add a unit test to reproduce this (create a tensor directly from a TypedArray with NaNs in debug mode and assert that it throws)

src/debug_mode_test.ts
annxingyuan: Done

src/util.ts
annxingyuan: Done

PR #1468, opened 2018-12-25
Add tf.util.range()

caisq:
Compatible with `np.arange()`.

Fixes https://github.com/tensorflow/tfjs/issues/1049

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1468)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -40,9 +40,22 @@ export function shuffle(array: any[]|Uint32Array|Int32Array|
 /**
  * Create an array of equally spaced numbers.
  *
+ * Example:
+ * ```
+ * console.log(tf.util.range(0, 6));
+ * // Gives [0, 1, 2, 3, 4, 5]
+ *
+ * console.log(tf.util.range(0, 6, 2));
+ * // Gives [0, 2, 4]
+ *
+ * console.log(tf.util.range(0, 6, -2));
+ * // Gives [0, -2, -4]
+ * ```
+ *
  * @param start The starting value (inclusive).
  * @param stop The ending value (exclusive).
  * @param step The spacing between adjacent values in the array.
+ *   Defaults to `1`.
  * @returns Output array of equally spaced numbers.
  */
 export function range(start: number, stop: number, step?: number): number[] {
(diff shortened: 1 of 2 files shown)


caisq:
@nsthorat Yes. As I commented in the issue https://github.com/tensorflow/tfjs/issues/1049, I think a method like this will still be useful, because `tf.range()` returns a tensor, not an array of numbers. It's cumbersome to get an uniformly spaced array of numbers by using a tensor method, because the it'd be necessary to call data() or dataSync() and do memory management.

DirkToewe:
Maybe a Generator instead of an Array output might be worth considering:
```js
function* range( start, end, step ) {
  // ...
  for( let i=0; i < end; i+=step )
    yield i
}
```
This way, less memory is used and it can still easily be converted into an Array using `[...tf.range(start,end,step) ]`.

caisq:
@DirkToewe I like that idea. I think we should support both an array return type and a generator return type The former is easier to use and just works for smallish problems. The latter scales up well. How about I make the signature of the function like  `function range(start, end, step, isGenerator = false)`?

nsthorat:
To be honest I wouldn't worry about memory consumption too much on the CPU. Let's cross that bridge if we find this to be a problem in practice. At the end of the day you're likely to store that as a TypedArray anyways.

Also, one more question, what are you going to use this for? Maybe return a TypedArray so that this is much faster?

DirkToewe:
Not sure if the question is directed at me or @caisq. Personally,I'm probably not gonna use it since I would rather use either [itertools](https://www.npmjs.com/package/itertools) or something like `Float32Array.from({length: 1337}, (_,i) => 10*i)`. In Python, range is used as way more idiomatic replacement of the awful C-style `for(;;)` loop which is prone to terrible and sometimes hilarious mistakes.
```js
for( const i of tf.arange(100) )
for( const j of tf.arange(100) )
  // ...
```
Is way more idiomatic than:
```
for( let i=0; i < 100; i++ )
for( let j=0; j < 100; i++ )
  // ...
```

(I hid an bug in the `for(;;)` loop just to point out how problematic it is)

If that is what @caisq had in mind, using a generator would be preferrable (Python first introduced [xrange](https://www.geeksforgeeks.org/range-vs-xrange-python/) and then changed `range` to a generator in Python 3 because creating a `list`, everytime `range()` is called, turned out to be a memory and performance issue).

nsthorat:
Gentle ping on this -- going through all the tfjs-core PRs.

nsthorat:
Gentle ping, what's going on with this?

caisq:
The use case isn't super clear. Closing this PR for now.

PR #1058, opened 2018-05-25
change the regex check for git origin

pyu10055:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1058)

PR #1473, opened 2019-01-02
Fixed division by zero in QR decomposition. Issue #1058

jarno-r:
#### Description
https://github.com/tensorflow/tfjs/issues/1058
The sign() function returns 0 on 0, which causes a division by zero in the QR decomposition function qr() if there is a zero on the diagonal.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1473)


jarno-r:
FYI: https://github.com/tensorflow/tfjs-core/pull/1366 has a new implementation of qr() using an entirely difference algorithm, which might also fix this issue.

DirkToewe:
I'm not sure I understand, why there is a sign there in first place. Isn't it completely arbitrary wether or not we Householder reflect to `[0, ..., 0, -norm, 0, ..., 0]` or `[0, ..., 0, +norm, 0, ..., 0]`? It makes the gradient smother when the input is close to triangular but:
  * The gradient cannot be computed as is anyways, since `dispose()` is called manually
  * The gradient becomes quite volatile if the diagonal entry is close to zero, since the sign of the diagonal might change on tiny pertubations while the norm might still be huge.

jarno-r:
@caisq Change committed.
@DirkToewe The code appears to be based directly on the lecture notes mentioned in the comments. There's no explanation there, why the sign is chosen as is. There's a somewhat cryptic note on the Wikipedia page on QR decomposition via Householder reflections claiming that the sign should be chosen to avoid losing significance, but that doesn't make any sense to me.
I really don't know much about how tensorflow.js works, so I can't say anything about the use of dispose().

DirkToewe:
Oh, that actually makes sense. If You subtract a large float form another equally large float the result is a small float that only represents the last few digits of the two large numbers so there is a loss in significance. So it makes sense from an accuracy standpoint. Since gradients can't be computed anyways yet, there is no valid argument against it either.

DirkToewe:
@caisq I believe, You meant to thank @jarno-r

caisq:
That's correct, @DirkToewe :)

DirkToewe:
@jarno-r I believe it's up to You to commit this PR. @caisq had given You the GTG.

jarno-r:
@DirkToewe It says 'Only those with write access to this repository can merge pull requests.' I don't think I have write access.

dsmilkov:
CLA is ok since the commits are mine due to sync with master.

jarno-r:
@googlebot I consent.

PR #1486, opened 2019-01-08
Make Engine explicit implement TensorTracker.

mprobst:
Previously, Engine would only structurally match TensorTracker, by
implementing the right methods. This works on the TypeScript level, but
fails when compiling with Closure Compiler. The optimizer never sees an
assignment of an `Engine` type into a `TensorTracker` due to the
indirection in the code, and then removes and disambiguates properties
incorrectly. This is a known limitation of Closure Compiler.

The fix is simple: explicitly implement the interface that's used to
refer to the type, and Closure Compiler is aware of the property
use/name collision.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1486)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -18,7 +18,7 @@
 import {BackendTimingInfo, DataMover, KernelBackend} from './kernels/backend';
 import {Profiler} from './profiler';
 import {backpropagateGradients, getFilteredNodesXToY, NamedGradientMap, TapeNode} from './tape';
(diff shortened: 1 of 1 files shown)


dsmilkov:
Hey Martin, your fix is already in - did it earlier today: https://github.com/tensorflow/tfjs-core/pull/1488

I'll close this PR. Thanks!

mprobst:
SGTM, thanks @dsmilkov.

PR #1491, opened 2019-01-09
Test Windows and MacOS for tfjs-core.

nkreeger:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1491)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ os:
   - osx
   - windows
 env:
-  - TO_TEST=regular
-  - TO_TEST=integration
+  - TO_TEST=regular YARN_GPG=no
+  - TO_TEST=integration YARN_GPG=no
 notifications:
   email:
     recipients:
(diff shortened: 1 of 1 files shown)


nsthorat:
Looking back at this -- I wonder if this is the right place to do this testing. tfjs-core is mostly browser unit testing. This would greatly increase the size of every travis run, I wonder if we could do this in the tfjs-node package instead so we don't bloat this?

dsmilkov:
+1 we won't get much benefit from calling `yarn/yarn test-node` in windows here. We will get much more benefit from running browserstack on Windows (to test our shaders on D3D via Angle), but that doesn't require a new travis OS (travis linux is sufficient for that)

I think tfjs-node would be the best place for this.

nsthorat:
Apologies, closed the wrong PR :)

nsthorat:
But I do think this should be closed anyways haha, let us know your thoughts :)

nsthorat:
Closing this out since we're doing this with cloud CI now.

PR #1500, opened 2019-01-22
Avoid repeating texture upload and expensive reshape

astojilj:
When there is reshaped texture for packed tensor already, avoid
repeating upload (compileAndRun -> dataSync -> uploadToGPU) to
GPU and packed reshape in every inference.

This is noticeable in e.g. conv2d implementation using matmul,
where 4D filter needs to be reshaped to 2D before matMul.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1500)


astojilj:
@annxingyuan PTAL

PR #1503, opened 2019-01-23
Potential fix for tfjs issue 545

ClementVidal:
When testing TensorFlow models using jest and **tfjs-node**, errors like the one reported by [tfjs issue 545](https://github.com/tensorflow/tfjs/issues/545) occurs.

I think they comes from the fact that jest patch several default objects like Array which make subsequent call to instanceof fail.

Probably because the Array object available at "creation time" is not the same that the one used at "test time".

Note that **this fix actually resolve the `tensor1d() requires values to be a flat/TypedArray` error** of me and @smith-kyle in [tfjs issue 545](https://github.com/tensorflow/tfjs/issues/545)

_Sorry for this approximative description, i'm not a tensor flow expert, i'm just trying to make our jest unit test work correctly with tensorflow._

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1503)

### DIFF
--- a/src/ops/norm.ts
+++ b/src/ops/norm.ts
@@ -87,7 +87,7 @@ function normImpl(
 
   // vector
   if (x.rank === 1 || typeof axis === 'number' ||
-      axis instanceof Array && axis.length === 1) {
+      Array.isArray(axis) && axis.length === 1) {
     if (p === 1) {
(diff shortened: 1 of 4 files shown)


ClementVidal:
@dsmilkov This should be done.

PR #1448, opened 2018-12-12
Packed batch<=>space ND: transpose, slice, pad, 5-D and 6-D packed sampling.

astojilj:
Packed transpose, slice, pad and 5-D and 6-D packed sampling, in this PR, implement support for packed batchToSpaceND and packed spaceToBatchND.
It is verified using DeepLabV3+ MobileNetV2 segmentation, with exclusion of packed batch normalization (erroneous).

Instead of adding additional tests, existing transpose, slice, pad, batchToSpaceND and spaceToBatchND tests are reused for packed operations using activateWebGLPackedTestEnv. activateWebGLPackedTestEnv and deactivateWebGLPackedTestEnv delimit set of tests to be run in both packed and non-packed environment mode. When majority of operations get their packed variants, this could be replaced by running additional packed env for all tests.

Didn't define separate WEBGL_PACK_XXXX for transpose, slice and pad - using master switch WEBGL_PACK for all of them.

PERF

---

##### For repository owners only:

Please remember to apply all applicable tags to your pull request.
Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY

For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1448)

PR #1504, opened 2019-01-23
Packed transpose.

astojilj:
This PR is a split from PR #1448 and implements packed transpose
operation. Together with packed slice and packed pad, this completes
packed batchToSpace and spaceToBatch operations.

Compiled shader's userCode for transpose tensor of shape [3, 1001, 1] with perm=[0, 2, 1] is:

```
    void main() {
      ivec3 rc = getOutputCoords();
      vec4 result = vec4(0.);
      result[0] = getChannel(getA(rc.x,rc.z,rc.y), vec2(rc.z,rc.y));
      if(++rc.z < 1001) {
        result[1] = getChannel(getA(rc.x,rc.z,rc.y), vec2(rc.z,rc.y));
      }
      --rc.z;
      if(++rc.y < 1) {
        result[2] = getChannel(getA(rc.x,rc.z,rc.y), vec2(rc.z,rc.y));
        if(++rc.z < 1001) {
          result[3] = getChannel(getA(rc.x,rc.z,rc.y), vec2(rc.z,rc.y));
        }
      }  
      setOutput(result);
    }
```

[Benchmark](https://gist.github.com/astojilj/971be3037c115a9fa5fbb5d204d4da9e) on discrete GPU on [ASUS ROG-GL702VM](https://www.asus.com/us/ROG-Republic-Of-Gamers/ROG-GL702VM/) show performance improvement:

|          transpose 6D tensor of shape            | ** non packed ** | ** packed ** |
|:--------------------:|:-------------------------:|:------------------------:|
|       [1, 19, 4, 19, 4, 1]       |     ~5.2ms                      |         ~5.2ms                 |
| [1, 19, 4, 19, 4, 96] |           ~23ms                |             ~12ms             |
| [1, 19, 4, 19, 4, 960] |           ~152ms                |             ~56ms             |

Note: [1, 19, 4, 19, 4, 960] transpose happens in DeepLabV3+ MobileNetV2.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1504)

### DIFF
--- a/src/kernels/webgl/transpose_packed_gpu.ts
+++ b/src/kernels/webgl/transpose_packed_gpu.ts
@@ -1,0 +1,84 @@
+/**
+ * @license
+ * Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+import {GPGPUProgram} from './gpgpu_math';
+import {getCoordsDataType} from './shader_compiler';
+import { getVecChannels } from '../packing_util';
+
+export class TransposePackedProgram implements GPGPUProgram {
+  variableNames = ['A'];
+  outputShape: number[];
+  userCode: string;
+  rank: number;
+  usesPackedTextures = true;
+
+  constructor(aShape: number[], newDim: number[]) {
+    const outputShape: number[] = new Array(aShape.length);
+    for (let i = 0; i < outputShape.length; i++) {
+      outputShape[i] = aShape[newDim[i]];
+    }
+    this.outputShape = outputShape;
+    this.rank = outputShape.length;
+    if (this.rank > 6) {
+      throw Error(
+          `Packed transpose for rank ${this.rank} is not yet supported.`);
+    }
+    const dtype = getCoordsDataType(this.rank);
+
+    const outputOrder = getVecChannels('rc', this.rank);
+    const switchedOrder = new Array(this.rank);
+    for (let i = 0; i < newDim.length; i++) {
+      switchedOrder[newDim[i]] = outputOrder[i];
+    }
+    const innerDims = `vec2(${switchedOrder.slice(-2).join()})`;
+    const cLimit =
+        `${outputOrder[this.rank - 1]} < ${outputShape[this.rank - 1]}`;
+    let main = '';
+    
+    const componentSetup = [
+      `${dtype} rc = resRC;`,
+      `${outputOrder[this.rank - 1]} += 1;
+       if(${cLimit}) {
+      `,
+      `}
+       rc = resRC;
+       ${outputOrder[this.rank - 2]} += 1;
+       if(${outputOrder[this.rank - 2]} < ${outputShape[this.rank - 2]}) {`,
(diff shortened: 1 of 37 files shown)


src/kernels/webgl/transpose_packed_gpu.ts
```diff
@@ -0,0 +1,84 @@
...
+       if(${outputOrder[this.rank - 2]} < ${outputShape[this.rank - 2]}) {`,
+      `  ${outputOrder[this.rank - 1]} += 1;
+         if(${cLimit}) {`
+    ];
+    for (let i = 0; i < 4; i++) {
+      main += `
+        ${componentSetup[i]}
+          result[${i}] =
+            getChannel(getA(${switchedOrder.join()}), ${innerDims});
+      `;
```
annxingyuan: This is a bit hard to read - would it make sense to just write the iteration through the rgba channels out?

astojilj:
@annxingyuan 
>      Run a benchmark on repeatedly transposing a large tensor 

Should the benchmark compare "pack + packed transpose + unpack" vs unpack transpose?

So far, performance improvement is verified with batchToSpace and spaceToBatch packed ([slice, pad and transpose](https://github.com/tensorflow/tfjs-core/blob/2e3142bdfd40173d4dcc125a5dfeeda7c8670aa7/src/kernels/backend_webgl.ts#L953) all packed make batch<->space packed).

If "pack + packed transpose + unpack" doesn't perform faster than unpacked transpose, we could make the choice, to use packed ot to use unpacked version, dependent on transpose operation input + environment flag?

Or you'd like to verify something else about packed transpose.

>     * Paste the compiled shader code into the description for this PR
Of course. I plan to rewrite the shader - expand it to make it easier to read.

annxingyuan:
@astojilj 

I would say that as long as the packed transpose by itself is clearly faster than unpacked transpose, we can consider this an improvement (this is the standard we've been using for introducing other packed ops). Unlike with other ops we've packed (binary ops, unary ops, batchnorm, etc.), the additional work per fragment shader in packed transpose can't be vectorized, so we're essentially de-parallelizing work. Performance should still improve since the unpacked transpose work is well within the startup cost of a shader, but it would be good to have some measurements.

> So far, performance improvement is verified with batchToSpace and spaceToBatch packed (slice, pad and transpose all packed make batch<->space packed).

Are you saying you ran a benchmark on batchToSpace and saw an improvement? Could you include those measurements in this PR?

> If "pack + packed transpose + unpack" doesn't perform faster than unpacked transpose, we could make the choice, to use packed ot to use unpacked version, dependent on transpose operation input + environment flag?

Yes it might be a good idea to introduce heuristics like this at some point - let's not worry about it for now though :)

src/kernels/webgl/transpose_packed_gpu.ts
astojilj: Done.

astojilj:
@annxingyuan 
Here is a [live benchmark](https://01org.github.io/depth-camera-web-demo/nn/using-deeplab/index.html) of DeepLabV3+ MobileNetV2; use checkbox "Use packed array ops" to toggle environment flag WEBGL_PACK_ARRAY_OPERATIONS.

Setting WEBGL_PACK_ARRAY_OPERATIONS activates packed slice, transpose and pad, and inherently, batchToSpaceND and spaceToBatchND are packed, too.

Feel free to run it on different hardware - I've checked on [MacBook Pro (Retina, 15-inch, Mid 2014) with integrated card](https://support.apple.com/kb/sp704?locale=en_US): when all checkboxes are checked (also WEBGL_PACK_ARRAY_OPERATIONS), inference takes ~125ms.
When unchecking  "Use packed array ops" (WEBGL_PACK_ARRAY_OPERATIONS=false), inference takes ~150ms..

>https://github.com/tensorflow/tfjs-core/tree/master/integration_tests/benchmarks - see https://github.com/tensorflow/tfjs-core/blob/master/integration_tests/benchmarks/matmul_benchmarks.ts for an example) and compare to the unpacked version

I still need to check this.

annxingyuan:
Thanks @astojilj

PR #1506, opened 2019-01-23
Wrap cropAndResize with an `op` decorator

dsmilkov:
Because we weren't wrapping cropAndResize using the op decorator, a user was seeing this stack trace when executing a model that had cropAndResize as an op:
```
engine.ts:197 Uncaught (in promise) TypeError: Cannot read property 'name' of null
    at e.runKernel (engine.ts:197)
    at Object.cropAndResize_ [as cropAndResize] (image_ops.ts:296)
    at executeOp$8 (image_executor.ts:65)
    at executeOp$16 (operation_executor.ts:65)
    at p (graph_executor.ts:323)
    at e.processStack (graph_executor.ts:309)
    at e.<anonymous> (graph_executor.ts:294)
    at callbacks.ts:17
    at Object.next (callbacks.ts:17)
    at i (callbacks.ts:17)
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1506)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -194,7 +194,7 @@ export class Engine implements TensorManager, TensorTracker, DataMover {
       saved.push(x);
       return x;
     };
-    const scopeName = this.activeScope.name;
+    const scopeName = this.activeScope != null ? this.activeScope.name : '';
     const startingBytecount = this.numBytes;
     const startingNumTensors = this.numTensors;
 
--- a/src/ops/image_ops.ts
+++ b/src/ops/image_ops.ts
@@ -301,4 +301,4 @@ export const resizeBilinear = op({resizeBilinear_});
 export const resizeNearestNeighbor = op({resizeNearestNeighbor_});
 export const nonMaxSuppression = op({nonMaxSuppression_});
 export const nonMaxSuppressionAsync = nonMaxSuppressionAsync_;
-export const cropAndResize = cropAndResize_;
+export const cropAndResize = op({cropAndResize_});


timekarma:
After taking https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.14.2/dist/tf.min.js
I am now getting this error when 
loading the Model  ok
then error is seen here
 const pixels = tf.fromPixels(image);
  const res1 = await model.executeAsync(pixels.reshape([1, ...pixels.shape]));

tensor_array.ts:116 Uncaught (in promise) Error: TensorArray : Could not write to TensorArray index 0,
          because the value dtype is int32, but TensorArray dtype is float32.
    at e.write (tensor_array.ts:116)
    at tensor_array.ts:162
    at Array.forEach (<anonymous>)
    at e.writeMany (tensor_array.ts:162)
    at e.scatter (tensor_array.ts:252)
    at control_executor.ts:127
    at callbacks.ts:17
    at Object.next (callbacks.ts:17)
    at callbacks.ts:17
    at new Promise (<anonymous>)

timekarma:
The above error is coming in 0.14.2 release.
In 0.14.1 it works fine. I tested it using the standard object detection example

nsthorat:
@timekarma would you mind filing an issue here: https://github.com/tensorflow/tfjs/issues/new

I think this PR is unrelated.

Also, if you could host your model for us to take a look at we can fix it ASAP!

PR #1510, opened 2019-01-24
Support WeChat mini app environment

pyu10055:
To compensate the differences between browser and WeChat mini app:

- WeChat mini app runs on JS core (ios) which does not have document, window, and setImmediate function or objects.
- When creating a GPGUContext with a existing context, it needs to store the context for the GL version, otherwise it would be picked later.
- expose setTensorTracker function to allow external backend registration to use.

This PR also fix the inconsistency issue with GPGPUContext constructor, it should always cache the rendering context.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1510)


pyu10055:
> > _[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1510)_ status: 0 of 1 approvals obtained (waiting on @pyu10055)
> 
> _[src/environment.ts, line 466 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1510#-LX5R-KU-X0t_kXYcV7U:-LXLZTZlByCUX06gb4Ql:bgmp7av) ([raw file](https://github.com/tensorflow/tfjs-core/blob/a95581e2f6c6e7ce5d879af458b532b4e84b13eb/src/environment.ts#L466)):_
> 
> Previously, pyu10055 (Ping Yu) wrote…
> Remove the _global variable since you don't really need it.
This cache speeds up the lookup, since now this method can be called multiple times.

PR #1514, opened 2019-01-25
Update to 0.15.0 and update the publish npm script to allow publishing from a release branch.

nsthorat:
This PR is PRing into the 0.15.x branch as well.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1514)


nsthorat:
The errors are in node.js and the converter which need a release before being fixed.

PR #1515, opened 2019-01-27
Make doc of stridedSlice consistent to the API

Lewuathe:
MISC

Since it seems to be copied from the [TensorFlow core API reference](https://www.tensorflow.org/api_docs/python/tf/strided_slice), the parameter name is not consistent with the TensorFlow.js implementation. We should be able to use the consistent parameter name with the argument.

See: https://js.tensorflow.org/api/0.14.2/#stridedSlice

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1515)


dsmilkov:
Thank you!

PR #1523, opened 2019-02-01
implement isNaN, isInf, isFinite

AndreasMadsen:
also implement NAN, INFINITY, isnan, isinf for the WebGL backend.

~~_I suppose my `isnan()` implementation might not work. The previous was okay, I just wanted to try this one as it is a bit simpler._~~

---

Some odd things I have noticed:

* The `getNaN()` didn't appear to be used anywhere.

* The caching mechanics for the previous implementation of `NAN`, didn't appear to work. As every call to an operator would initialize a new `GPUProgram`. There is now no caching implementation for `NAN`, but in WebGL 2.0 it is defined as a constant.

* The `PROD` env handling of NAN, gives a false sense of security and causes undefined behavior. In `PROD`, NAN's are still produced and propagated through arithmetic operators. It is simply the specialized NAN handling that is disabled, which means the behavior is determined by the arithmetic operators. These often have undefined behavior, since some WebGL environments don't support NAN natively.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1523)

### DIFF
--- a/src/kernels/backend.ts
+++ b/src/kernels/backend.ts
@@ -275,6 +275,16 @@ export class KernelBackend implements TensorStorage, Backend, BackendTimer {
     throw new Error('Not yet implemented');
   }
 
+  isNaN<T extends Tensor>(x: T): T {
+    throw new Error('Not yet implemented');
+  }
+  isInf<T extends Tensor>(x: T): T {
+    throw new Error('Not yet implemented');
+  }
+  isFinite<T extends Tensor>(x: T): T {
+    throw new Error('Not yet implemented');
+  }
+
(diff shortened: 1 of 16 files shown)


AndreasMadsen:
@nsthorat what can I do to get some activity on this? Sorry, if I /cc'ed the wrong person.

nsthorat:
Looking at this now, sorry for the delay.

AndreasMadsen:
@nsthorat Thanks :)

nsthorat:
Yep, you are totally right about the queried bit! Last comment then I will merge this, thanks for the awesome contribution :)

nsthorat:
Thanks again for the PR!

PR #1527, opened 2019-02-03
Update master version to 1.0.0-alpha1.

nsthorat:
Also update the publish-npm script to use `--tag next` so we don't accidentally publish 1.0 under the `@latest` tag.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1527)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-core",
-  "version": "0.14.5",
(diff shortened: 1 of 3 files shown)


nsthorat:
And to be clear: we don't have to do the first alpha release yet. Let's check in some of the breaking changes to core first, then I'll update the other packages to depend on alpha.

PR #1531, opened 2019-02-04
cherry-pick to 0.15.x: Always test for a fence with gl2.clientWaitSync(sync, 0, 0) (#1529)

nsthorat:
This is a cherry pick of https://github.com/tensorflow/tfjs-core/pull/1529 into 0.15.x.

Fixes https://github.com/tensorflow/tfjs/issues/1145

The smallest repro is calling .data() 3 times without waiting:
```ts
tf.ENV.set('WEBGL_CPU_FORWARD', false)
tf.ENV.set('WEBGL_SIZE_UPLOAD_UNIFORM', 0);
tf.square(3).data();
tf.square(3).data();
tf.square(3).data();
```

There are 3 fences that are in parallel. And on the next tick the binary search tests only the middle fence and sees that it is done, and assumes the first fence is done too. But since we never explicitly tested fence[0].isDone(), Chrome gives a warning.

The warnings started showing up more often when Layers started doing parallel reads of the scalar loss values: https://github.com/tensorflow/tfjs-layers/pull/398

I changed the binary search to linear search so we also test all fences up to the one that is not done.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1531)

### DIFF
--- a/src/kernels/webgl/gpgpu_context.ts
+++ b/src/kernels/webgl/gpgpu_context.ts
@@ -518,9 +518,8 @@ export class GPGPUContext {
   private itemsToPoll: PollItem[] = [];
 
   pollItems(): void {
-    // Find the last query that has finished using binary search.
-    // All other queries before it are also done.
-    const index = binSearchLastTrue(this.itemsToPoll.map(x => x.isDoneFn));
+    // Find the last query that has finished.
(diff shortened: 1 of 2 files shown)


dsmilkov:
I signed it!

PR #1538, opened 2019-02-05
Optimize addN op by the multiple upload of textures

Lewuathe:
PERF

As well as the optimization for concat op, we can optimize `addN` op by uploading multiple textures up to ` WEBGL_MAX_TEXTURES_IN_SHADER` so that we can reduce the overhead of texture upload.

Here is the benchmark of optimized `addN` ops. We can see the significant performance improvement in case of over 4000 tensors.

Environment
- macOS: 10.13.6
- Chrome: 71.0.3578.98

```js
const nums = [1, 2, 3, 4, 5, 6, 7, 8, 9];
const results = nums.map(n => {
  const tensors = [];
  for (let i = 0; i < n * 1000; i++) {
    tensors.push(tf.ones([100, 100]));
  }
  const res = tf.addN(tensors);
  res.dataSync();
  return performance.now() - start;
});
console.log(results);
```

chart

See: https://github.com/tensorflow/tfjs/issues/989

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1538)


nsthorat:
Looks good overall, one small note is that this isn't going to be supported when packing is enabled. If you feel motivated you can add the packed version of this, otherwise we can do this in a follow up and this LGTM :)

Lewuathe:
@nsthorat Thanks! Is there any good reference (or code) to implement packed version?

nsthorat:
Check out the code here for binary ops that are packed: https://github.com/tensorflow/tfjs-core/blob/master/src/kernels/webgl/binaryop_packed_gpu.ts

Lewuathe:
@nsthorat Thank you! Try to take a look.

nsthorat:
Let me know when this is ready to review :)

Lewuathe:
@nsthorat Ah, sorry. It's basically ready for the review.

src/kernels/webgl/backend_webgl.ts
```diff
@@ -1468,11 +1470,31 @@ export class MathBackendWebGL implements KernelBackend {
...
+      const rightSide = this.addN(tensors.slice(midIndex));
+      return this.addN([leftSide, rightSide]);
+    }
+
+    const dtype = tensors
+        .map(t => t.dtype)
+        .reduce((d1, d2) => upcastType(d1, d2));
+    const shapes = tensors.map(t => t.shape);
+    // We can make sure shapes are identical in op level.
+    const usePackedOp = ENV.getBool('WEBGL_PACK_BINARY_OPERATIONS');
```
annxingyuan: Let's just check `WEBGL_PACK` here rather than checking `WEBGL_PACK_BINARY_OPERATIONS` or creating a new flag.

Lewuathe: Thanks for pointing out!

Lewuathe:
@nsthorat @annxingyuan I updated it accordingly. Please take a look when you get a chance.

PR #1543, opened 2019-02-05
Remove delayedStorage mode. Also add internal format for packed half float textures and pass in backend at a few more tensor construction sites.

annxingyuan:
This PR fixes https://github.com/tensorflow/tfjs/issues/1194

#### Changes
- Remove `delayedStorage` as configurable parameter of WebGL backend constructor. Replace `delayedStorage` with private `keepCopyOfTextureOnGPUAfterRead` flag that can be flipped to enable reading from a texture without releasing it. 
- Add a new internal format `internalFormatPackedHalfFloat` (`gl.RGBA16F`) to support packed half float textures.
- Passing in backend in a few more places where we construct tensors for good measure.

BUG

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1543)

### DIFF
--- a/src/engine.ts
+++ b/src/engine.ts
@@ -262,7 +262,7 @@ export class Engine implements TensorManager, TensorTracker, DataMover {
         bytes = util.sizeFromShape(a.shape) * util.bytesPerElement(a.dtype);
       }
       this.tensorInfo.set(a.dataId, {
-        backend: this.backend,
+        backend: backend != null ? backend : this.backend,
         dtype: a.dtype,
         shape: a.shape,
         bytes,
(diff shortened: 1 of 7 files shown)


nsthorat:
lmk when I should review

annxingyuan:
Hey @nsthorat @dsmilkov - this PR is ready for another look.

I ended up not quite completely removing delayed storage... we still need to be able to internally control whether we are releasing textures after we read them for a special packing-related case :/

annxingyuan:
Hey @nsthorat @dsmilkov - this PR is ready for another look. Unfortunately I wasn't able to get rid of the weird condition altogether (for that we need to wait until packed uploading is the default - currently it's not because we set the `isPacked` bit on the texture lazily to match the program it's going into). But I was able to remove the dataSync.

PR #1495, opened 2019-01-17
Avoid expensive reshape for odd number of rows in conv2D

astojilj:
This PR addresses tensorflow/tfjs#949; conv2D is the main source of expensive reshapes in e.g. DeepLabV3 + MobileNetV2 model: before the patch, with environmental flags 'WEBGL_PACK:true,WEBGL_PACK_DEPTHWISECONV:false', there are 70 expensive reshapes per inference - with the patch, there are only 4.

Performance is measured using [modified conv2d benchmark](https://gist.github.com/astojilj/4ce6b26807b9f9636dbe81c7efba4166) in such way that it uses conv2d by matmul implementation: 

##### master branch, unpacked; [(tf.ENV.set('WEBGL_PACK', false)](https://gist.github.com/astojilj/4ce6b26807b9f9636dbe81c7efba4166#file-packedconv2dbenchmark-patch-L25);
```
    "N=227 {\"inDepth\":16,\"outDepth\":32,\"filterSize\":1,\"stride\":1,\"pad\":\"same\"}": {
      "averageTimeMs": 103.76119999957155
    },
    "N=513 {\"inDepth\":16,\"outDepth\":32,\"filterSize\":1,\"stride\":1,\"pad\":\"same\"}": {
      "averageTimeMs": 616.3369999990391
    }
```
##### master branch, packed ([tf.ENV.set('WEBGL_PACK', true)](https://gist.github.com/astojilj/4ce6b26807b9f9636dbe81c7efba4166#file-packedconv2dbenchmark-patch-L25)) 
```
    "N=227 {\"inDepth\":16,\"outDepth\":32,\"filterSize\":1,\"stride\":1,\"pad\":\"same\"}": {
      "averageTimeMs": 49.42674999969313
    },
    "N=513 {\"inDepth\":16,\"outDepth\":32,\"filterSize\":1,\"stride\":1,\"pad\":\"same\"}": {
      "averageTimeMs": 216.8936500007112
    }
```

##### astojilj:conv-avoid-expensive-reshape branch, packed ([tf.ENV.set('WEBGL_PACK', true)](https://gist.github.com/astojilj/4ce6b26807b9f9636dbe81c7efba4166#file-packedconv2dbenchmark-patch-L25)) 
```
    "N=227 {\"inDepth\":16,\"outDepth\":32,\"filterSize\":1,\"stride\":1,\"pad\":\"same\"}": {
      "averageTimeMs": 32.91375000029802
    },
[...]

PR #1553, opened 2019-02-08
Fix reference issue in free-packed-reshape-conv2D approach

astojilj:
This fix relates to tensorflow/tfjs#949 and fix committed in PR #1495.

xTexData.shape gets referenced from GPGPUBinary.inShapeInfos. Decrementing
row count, after batchMatMul->...->compileProgram leads to invalid row
count within the reference in GPGPUBinary.inShapeInfos. Proper way would be
to provide a copy to GPGPUBinary.inShapeInfos but that would affect
all the program compilation - so, provide a  copy, with even row cound,
when calling batchMatMul->...->compileProgram and after that, the original
xTexData.shape is restored.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1553)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -1681,6 +1681,11 @@ export class MathBackendWebGL implements KernelBackend {
             [1, xShape[0] * xShape[1] * (xShape[2] + 1), convInfo.inChannels],
             {dataId: x.dataId}, x.dtype) as Tensor3D;
 
+    // xTexData.shape gets referenced from GPGPUBinary.inShapeInfos. Provide a
+    // copy to it with even row count. After the method, the original
+    // xTexData.shape needs to be restored.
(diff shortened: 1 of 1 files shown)


astojilj:
@annxingyuan 
PTAL
I wasn't able to create a test that would repproduce the issue, but was able to verify incorrect state using instrumentation code on deeplab model.

annxingyuan:
Thanks @astojilj ! 

Can you explain why it's not possible to create a test case?

astojilj:
> Thanks @astojilj !
> 
> Can you explain why it's not possible to create a test case?
@annxingyuan 

Not yet :) I'm able to reproduce it with other, invalid changes within backend_webgl.ts I use to  benchmark performance, but wasn't able to create such state from unit tests with original backend_webgl.ts. I'll continue trying to do it.

astojilj:
Closing as merged to PR #1599

PR #1558, opened 2019-02-09
Add fill kernel in backend

Lewuathe:
PERF

That is necessary in order to make use of TensorFlow core via tfjs-node.
Fill op in tfjs-core can change the backend kernel to optimize the performance.

And we can improve the performance by itself. Here is the result of micro benchmark I ran. Using GPU for fill kernel is slightly faster along with the size of tensor is increasing.

Environment
- macOS: 10.13.6
- Chrome: 71.0.3578.98

```js
const nums = [10, 20, 30, 40, 50, 60, 70, 80, 90, 100];
const results = nums.map(n => {
  const tensors = [];
  for (let i = 0; i < 100; i++) {
    const res = tf.ones([n * n, n * n]);
    res.dataSync();
  }
  return performance.now() - start;
});
```

screen shot 2019-02-09 at 11 26 49

See: https://github.com/tensorflow/tfjs/issues/1211

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1558)

### DIFF
--- a/src/kernels/backend.ts
+++ b/src/kernels/backend.ts
@@ -583,6 +583,12 @@ export class KernelBackend implements TensorStorage, Backend, BackendTimer {
       defaultValue: Scalar): Tensor<R> {
     throw new Error('Not yet implemented');
   }
+
+  fill<R extends Rank>(
+      shape: ShapeMap[R], value: number|string, dtype?: DataType): Tensor<R> {
+    throw new Error('Not yet implemented.');
+  }
+
   /**
    * Sets the data mover for this backend. Backends should use the mover to
    * move data from other backends to this backend.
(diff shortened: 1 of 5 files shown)


caisq:
cc myself. Thanks, @Lewuathe !

caisq:
@dsmilkov @Lewuathe  I played with tfjs-node by linking it against this PR. I can confirm that this change, together with some other changes I'll submit, can increase Model.fit() speed by 8% in some case (GPU). Can we prioritize reviewing and merging this PR?

Lewuathe:
@caisq Great to hear that!

dsmilkov:
Done! Thank you Kai! Feel free to do the rest of the ops mentioned in https://github.com/tensorflow/tfjs/issues/1211#issuecomment-461480625 in a single PR (or combine several in a few PRs).

caisq:
@dsmilkov @Lewuathe I have a PR almost ready for onesLike and zerosLike, but feel free to do the remaining ops, including range, linspace, randomUniform and randomNormal

PR #1559, opened 2019-02-09
Fix wrong paging accounting for pointwise conv2D

astojilj:
- pointwise conv2D implementation caused wrong calculation of MathBackendWebGL.numBytesInGPU, in time increasing it and leading to unecessary paging activation.
- added unit test exposing the issue.
- simplify computeBytes computation to use texShape:
  - texShape is not equal to logical shape,
  - no need for call to generic utils.sizeFromShape
    since we always have texShape[0] * texShape[1]
- save one this.texData.get lookup per call.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1559)

### DIFF
--- a/src/kernels/backend_webgl.ts
+++ b/src/kernels/backend_webgl.ts
@@ -2219,8 +2219,8 @@ export class MathBackendWebGL implements KernelBackend {
       let numBytesToPage = this.numBytesInGPU - numBytesBeforePaging;
       while (numBytesToPage > 0 && this.lruDataGPU.length > 0) {
         const dataId = this.lruDataGPU.shift();
-        const {texShape, dtype, isPacked} = this.texData.get(dataId);
-        numBytesToPage -= this.computeBytes(texShape, dtype, isPacked);
+        const {texShape, dtype} = this.texData.get(dataId);
+        numBytesToPage -= this.computeBytes(texShape, dtype);
         this.read(dataId);
       }
     }
@@ -2364,7 +2364,7 @@ export class MathBackendWebGL implements KernelBackend {
         this.lruDataGPU.splice(idx, 1);
       }
     }
-    this.numBytesInGPU -= this.computeBytes(texShape, dtype, isPacked);
+    this.numBytesInGPU -= this.computeBytes(texShape, dtype);
     this.textureManager.releaseTexture(texture, texShape, texType, isPacked);
   }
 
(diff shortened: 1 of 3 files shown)


astojilj:
@annxingyuan 
PTAL.
FYI, this is a fix to [critical issue merged via PR #1495](https://github.com/tensorflow/tfjs-core/pull/1495#issuecomment-462035765)

annxingyuan:
Hey @astojilj - thank you for addressing this issue. I have some questions before we merge:

- It seems like both this PR and [PR 1553](https://github.com/tensorflow/tfjs-core/pull/1553) address the paging issue - is that right? If so could you combine the fix in PR 1553 with this one and change this PR name to mention fixing the memory leak in pointwise convolution? 
- Could you add a test to `conv2d_test.ts` that pointwise convolution does not leak memory in the WebGL backend? And could you make sure that this test fails on master?
- Could you add a test to `engine_test.ts` that `tf.profile` returns the correct byte usage for packed vs unpacked textures?
- Why does `dtype` need to be passed into `acquireTexture` and `releaseTexture` now? Does the passed-in `dtype` ever differ from `texData.get(dataId).dtype`? If so could you include a test or at least a comment that explains why they would ever differ? If not could you remove the `dtype` argument?

Thanks again for looking into this!

astojilj:
@annxingyuan 
> It seems like both this PR and PR 1553 address the paging issue - is that right? If so could you combine the fix in PR 1553 with this one and change this PR name to mention fixing the memory leak in pointwise convolution?

Combined. Thanks. It is not about memory leak - it is about not restoring numBytesInGPU to original value after tensor gets disposed. 

> Could you add a test to conv2d_test.ts that pointwise convolution does not leak memory in the WebGL backend? And could you make sure that this test fails on master?

I've added a test for both numBytes and numBytesInGPU. Verified that in master numBytesInGPU check fails.

>Could you add a test to engine_test.ts that tf.profile returns the correct byte usage for packed vs unpacked textures?

That would be orthogonal issue to this patch as numBytes are not affected. Only numBytesInGPU (used internally with 'WEBGL_NUM_MB_BEFORE_PAGING' for paging: " Number of megabytes bytes allocated on the GPU before we start moving data to cpu).

That said, I would need to change numBytesInGPU tests in backend_webgl.ts.

> Why does dtype need to be passed into acquireTexture and releaseTexture now? Does the passed-in dtype ever differ from texData.get(dataId).dtype? If so could you include a test or at least a comment that explains why they would ever differ? If not could you remove the dtype argument?

No, it doesn't differ - callers of acquireTexture and releaseTexture already call texData.get(dataId).dtype so I found it better to pass the parameter then to call texData.get(dataId).dtype again. It is the same logic as with texData.get(dataId) parameter: it is fetched by acquireTexture caller and passed.

astojilj:
@annxingyuan 

>That said, I would need to change numBytesInGPU tests in backend_webgl.ts.

Extended the existing lazy packing and unpacking test. Fixed there also that add is non packed operation.
With it, addressed all the comments (and fixed computeBytesOnGPU method) you had. PTAL.
Thanks.

astojilj:
> LGTM - thanks @astojilj !

@annxingyuan, thank you.

dsmilkov:
Thank you @astojilj for this fix!!

astojilj:
@dsmilkov no problem and thank you, too - I enjoy work on this project.

PR #1583, opened 2019-02-21
Ensure uniform cache is updated when only maxLoc is null

Lewuathe:
MISC

Even if only `maxLoc` is null, the uniform cache should be updated.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1583)


nsthorat:
Closing this out for inactivity. If this is still an issue in the future let's open a new PR.

PR #1585, opened 2019-02-23
backend: Add ops onesLike and zerosLike

caisq:
- Also add CPU and WebGL backend implementations of these
  based on fill()
- This will be used in tfjs-node to bind to the libtensorflow
  implementations of onesLike and zerosLike to save memory
  copying between CPU and GPU, which will lead to performance
  improvements.

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1585)


caisq:
cc @Lewuathe

PR #1590, opened 2019-02-26
Delete tf.fromPixels and tf.toPixels.

nsthorat:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1590)


nsthorat:
Actually it was just moved: https://github.com/tensorflow/tfjs-core/blob/master/src/ops/browser.ts#L152

PR #1594, opened 2019-02-27
Fix upload of half-float textures in "WebGL 1"

nkreeger:
Looks like this was regressed a few months ago here: https://github.com/tensorflow/tfjs-core/pull/1375

I'm curious how this worked on iOS devices? From what I was told - Safari only supported `OES_texture_half_float` which enables creating `gl.OES_HALF_FLOAT` textures but allows them to write f32 data in `gl.texSubImage2D()`.

Happy to discuss offline. But this essentially makes devices that only support f16 not work with the regression in texture manager.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1594)

### DIFF
--- a/src/kernels/webgl/texture_manager.ts
+++ b/src/kernels/webgl/texture_manager.ts
@@ -145,8 +145,13 @@ export class TextureManager {
 function getPhysicalFromLogicalTextureType(
     logicalTexType: TextureUsage, isPacked: boolean): PhysicalTextureType {
   if (logicalTexType === TextureUsage.UPLOAD) {
-    return isPacked ? PhysicalTextureType.PACKED_2X2_FLOAT32 :
-                      PhysicalTextureType.UNPACKED_FLOAT32;
+    if (ENV.get('WEBGL_RENDER_FLOAT32_ENABLED')) {
+      return isPacked ? PhysicalTextureType.PACKED_2X2_FLOAT32 :
+                        PhysicalTextureType.UNPACKED_FLOAT32;
+    } else {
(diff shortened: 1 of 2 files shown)


nsthorat:
I don't think this will work on an iPhone -- we've always uploaded to 32 bit textures there (there is no way to upload to a 16 bit texture, I read the WebKit code :P)

nkreeger:
Upload yes - from my understanding things work like this:

```js
const oes = gl.getExtension('OES_texture_half_float');

// Create texture (for upload):
const t = gl.createTexture();
gl.bindTexture(gl.TEXTURE_2D, t);
...
gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, border, gl.RGBA, ext.HALF_FLOAT_OES, null);

// Now actually send values to texture:
const values = new Float32Array([0.5, 1.5, 2.5, 3.5]);
gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, 1, 1, gl.RGBA, gl.FLOAT, values);
```

Note that the texture has to be created as f16, but data can be uploaded via f32.

nsthorat:
Have you tried that on an iPhone out of curiosity? That is the WebGL spec, but Apple iPhones fail most WebGL conformance tests.

nsthorat:
By the way, that seems completely fine if this works on iOS :)

nsthorat:
Actually, if this does work on iOS, we need to do a flag refactor. We don't have a flag for the upload type since it's unconditionally 32 bits right now (this is why you may be confused by the render / download flags).

nsthorat:
Closing this out since it's old.

PR #1596, opened 2019-02-28
Add fields to io.ModelArtifacts and surrounding save/load logic

caisq:
Towards: https://github.com/tensorflow/tfjs/issues/1285
Towards: https://github.com/tensorflow/tfjs/issues/1286

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1596)

PR #1597, opened 2019-03-01
Provide types to enforce the structure of model.json files.

davidsoergel:
We were reminded of this by discussion on https://github.com/tensorflow/tfjs-core/pull/1596 (and, related, https://github.com/tensorflow/tfjs-converter/pull/320).

My view is that those two PRs should make it into 1.0.  This one is less urgent because it's a non-breaking change that just helps enforce our intentions.

Obviously, this PR can't enforce anything on the Python side.  Ultimately we should probably do all this via a language-independent JSON Schema.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1597)


nsthorat:
Sry, I just merged a change to the file that has to do with error messaging - the merge should be simple: https://github.com/tensorflow/tfjs-core/pull/1599

Let me know if you have problems

PR #1599, opened 2019-03-01
Add error messaging when the model.json parsing fails. 

nsthorat:
If it has a .pb extension, send a nice error message about how to convert to json.

Also:
- Removes the ability to load binary files.
- Removes the string[] option to `io.browserHTTPRequest`

On a request to a .pb file:
image

An invalid request with OK status: 
image

OLD request to url with invalid status error:
image

NEW request to url with invalid status error:
image

Verified we can load models from GCP by linking this to union.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1599)


dsmilkov:
So much better! Thanks for improving the UX of loading models.

PR #1602, opened 2019-03-04
Remove command to log benchmarks since it's happening in union now.

annxingyuan:
We are moving this functionality to union so other tfjs packages can use the same infrastructure: https://github.com/tensorflow/tfjs/pull/1319

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1602)


dsmilkov:
Remove the benchmark code too?

annxingyuan:
@dsmilkov good point!

PR #1603, opened 2019-03-05
Check explicitly for equals 0 in packed division kernel.

annxingyuan:
This PR fixes https://github.com/tensorflow/tfjs/issues/1321

I benchmarked repeated division calls given this change and there is no performance impact compared to the previous kernel. 

I don't know how to reproduce this issue outside of the LSTM text generation example. I also don't understand why straightforward equality check fails. I thought that maybe the problem was mixing in unused channels that experienced division by 0, so I tried a fix where we check whether a channel is out of bounds and explicitly set it to zero, but that didn't seem to fix Shanqing's bug. 

Hopefully this is enough of a band-aid for now. I created an issue (https://github.com/tensorflow/tfjs/issues/1324) for better understanding the root cause of this. 

This change integrates with the tfjs-layers test suite on my MBP. 

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1603)

### DIFF
--- a/src/kernels/webgl/binaryop_packed_gpu.ts
+++ b/src/kernels/webgl/binaryop_packed_gpu.ts
@@ -28,14 +28,33 @@ const CHECK_NAN_SNIPPET = `
 
 // We do the same as in ./binaryop_gpu, with vec4 and ivec4.
 // On Linux, the vectorized implementation produces NaNs when a and b are 0.
+// We must also check for a / b channels equal to 0 in case they are empty.
+// Equality check fails, so checking epsilon.
+// TODO: Find a vectorized implementation of epsilon check that works on Linux.
 export const DIV = `
   // vec4 one = vec4(equal(a, b));
   // return one + (vec4(1.0) - one) * a / b;
+  float epsilon = 0.00000001;
   vec4 result = a / b;
+
   result.x = a.x == b.x ? 1. : result.x;
   result.y = a.y == b.y ? 1. : result.y;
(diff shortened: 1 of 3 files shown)


annxingyuan:
cc @caisq

annxingyuan:
@nsthorat yes - environment epsilon is not available at that point :/

annxingyuan:
@nsthorat I ended up removing the test because (1) it doesn't fail on master so does not actually expose the root cause and (2) I was struggling to get readPixels to work with bytes for Safari (somehow just evaluating the environment flag 'is_download_float_enabled' within the test was enough to cause Chrome to fail?)

caisq:
@annxingyuan thanks for investigating the issue and putting together a fix quickly. To make sure I understand the problem in the lstm-text-generation example: do division-by-zero actually happen in that example? From the fact that this works with older versions of TF.js and with the WEBGL_PACK flag gets rid of the problem, it seems that the division-by-zero isn't due to that particular model or the data, is it? (Just making sure there isn't a bug in the tfjs-example)

annxingyuan:
@caisq No it does not appear that division by 0 occurs in the lstm-text-generation example. Or at least, if the denominator is 0 somewhere, the numerator is also 0 (which means they are equal and we return 1). So the example seems fine.

dsmilkov:
Let's close this one and use https://github.com/tensorflow/tfjs-core/pull/1605 instead.

PR #1609, opened 2019-03-06
Support complex type in onesLike

Lewuathe:
FEATURE

As TensorFlow supports complex type in `ones_like` ops, we can support `complex64` tensor in `onesLike` ops.

See: https://www.tensorflow.org/api_docs/python/tf/ones_like

```
>>> tf.ones_like([[1,2,3], [1,2,3]], dtype='complex64').eval()
array([[1.+0.j, 1.+0.j, 1.+0.j],
       [1.+0.j, 1.+0.j, 1.+0.j]], dtype=complex64)
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1609)

### DIFF
--- a/src/ops/array_ops_test.ts
+++ b/src/ops/array_ops_test.ts
@@ -511,6 +511,16 @@ describeWithFlags('onesLike', ALL_ENVS, () => {
     expectArraysEqual(b, [1, 1, 1]);
   });
 
+  it('1D complex dtype', () => {
+    const real = tf.tensor1d([1, 2, 3], 'float32');
+    const imag = tf.tensor1d([1, 2, 3], 'float32');
+    const a = tf.complex(real, imag);
+    const b = tf.onesLike(a);
+    expect(b.dtype).toBe('complex64');
+    expect(b.shape).toEqual([3]);
+    expectArraysEqual(b, [1, 0, 1, 0, 1, 0]);
+  });
+
   it('2D default dtype', () => {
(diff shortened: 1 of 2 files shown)


src/ops/array_ops_test.ts
```diff
@@ -511,6 +511,16 @@ describeWithFlags('onesLike', ALL_ENVS, () => {
...
   });
 
+  it('1D complex dtype', () => {
+    const real = tf.tensor1d([1, 2, 3], 'float32');
+    const imag = tf.tensor1d([1, 2, 3], 'float32');
+    const a = tf.complex(real, imag);
+    const b = tf.onesLike(a);
+    expect(b.dtype).toBe('complex64');
+    expect(b.shape).toEqual([3]);
+    expectArraysEqual(b, [1, 0, 1, 0, 1, 0]);
```
nsthorat: shouldn't these all be ones?

Lewuathe: `ones_like` of TensorFlow core generates a complex number with real 1 and imaginary 0. So I think it's consistent with TensorFlow core implementation. 

```
>>> tf.ones_like([1,2,3], dtype='complex64')
<tf.Tensor: id=18, shape=(3,), dtype=complex64, numpy=array([1.+0.j, 1.+0.j, 1.+0.j], dtype=complex64)>
```

PR #861, opened 2018-03-18
Consolidate redundant backend creation in tests

Lewuathe:
# Purpose

Creating backend object every test cases is redundant. We can make test faster by eliminating this redundant object creation.

# Result

Before

```
Chrome 65.0.3325 (Mac OS X 10.12.6): Executed 4089 of 4089 SUCCESS (2 mins 18.853 secs / 2 mins 17.784 secs)
Firefox 58.0.0 (Mac OS X 10.12.0): Executed 4089 of 4089 SUCCESS (9 mins 7.102 secs / 8 mins 25.818 secs)
```

After

```
Chrome 65.0.3325 (Mac OS X 10.12.6): Executed 4089 of 4089 SUCCESS (1 min 15.038 secs / 1 min 1.975 secs)
Firefox 59.0.0 (Mac OS X 10.12.0): Executed 4089 of 4089 SUCCESS (5 mins 6.947 secs / 4 mins 25.21 secs)
```

---
This change is [Reviewable](https://reviewable.io/reviews/pair-code/deeplearnjs/861)

PR #1444, opened 2018-12-11
Fix a firefox download issue 

rthadur:
#### Description

---

##### For repository owners only:

Please remember to apply all applicable tags to your pull request.
Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY

For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1444)

PR #1613, opened 2019-03-07
Fix download model in Firefox (#861)

kaesve:
Fixes downloading models in Firefox. Tested in Firefox, Chrome and Edge.

Fix demo: https://codepen.io/kaesve/pen/bZqEGL?editors=1011

Aside from dispatching a click event instead of calling `click()`, I added a
setTimeout to each download. Without making these calls asynchronous, Firefox
would only save the last of possibly two downloads.

Fixes tensorflow/tfjs#861

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1613)


kaesve:
I just realized there already is an open pull request for this issue at https://github.com/tensorflow/tfjs-core/pull/1444
My pull request is very similar but there are a few small differences:

* I'm using `new MouseEvent()` instead of `document.createEvent()` (could be a compatibility issue with IE)
* I don't have a special path for firefox, but use `anchor.dispatchEvent` for all browsers
* I included a fix for the unit test
* I added a setTimeout to make sure all files are downloaded, instead of just the last one

nsthorat:
Thanks for the PR, sorry for the delay!

PR #1617, opened 2019-03-10
Packed reverse

jgartman:
Adds a packed kernel for reverse:

| shape     |Packed  (ms)        | Unpacked (ms)  |
| ------------- |:-------------:| -----:|
| [100, 100]      | 5.814 | 5.542 |
| [200, 200]      | 6.002      |   5.640 |
| [500, 500]      | 7.109      |   16.257 |
| [1000, 1000]      | 18.289   |   44.871|

Benchmark code:
```
export class ReverseGPUBenchmark implements BenchmarkTest {
  async run(size: number): Promise<number> {
    tf.setBackend('webgl');
    tf.ENV.set('WEBGL_PACK', false);
    const t: tf.Tensor<tf.Rank.R2> = tf.randomNormal([size, size]);

    const benchmark = () => tf.reverse(t, [0]) as tf.Tensor;

    const time = await util.warmupAndBenchmarkGPU(benchmark);

    t.dispose();

    return time;
  }
}
```

Compiled shader code for reverse of tensor of `shape = [2, 3, 4`] along `axis = 1`:

```
     void main() {
          ivec3 rc = getOutputCoords();
          vec4 result = vec4(0.);
          result.r = getChannel(getX(rc.x,3 - rc.y - 1,rc.z), vec2(3 - rc.y - 1,rc.z));
          if(rc.z + 1 < 4){
            result.g = getChannel(getX(rc.x,3 - rc.y - 1,(rc.z + 1)), vec2(3 - rc.y - 1,(rc.z + 1)));
          }
          if(rc.y + 1 < 3) {
            result.b = getChannel(getX(rc.x,3 - (rc.y + 1) - 1,rc.z), vec2(3 - (rc.y + 1) - 1,rc.z));
            if(rc.z + 1 < 4) {
              result.a = getChannel(getX(rc.x,3 - (rc.y + 1) - 1,(rc.z + 1)), vec2(3 - (rc.y + 1) - 1,(rc.z + 1)));
            }
          }
          setOutput(result);
        }
      
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1617)


src/kernels/webgl/reverse_packed_gpu.ts
```diff
@@ -0,0 +1,80 @@
...
+        result.z = ${getc};
+        if(${nextColumn}) {
+          result.w = ${getc};
+        }
+      }
+    `;
+
+    this.userCode = `
+      void main() {
+      ${type} rc = getOutputCoords();
```
annxingyuan: Indent these lines.

src/kernels/webgl/reverse_packed_gpu.ts
```diff
@@ -0,0 +1,80 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc. All Rights Reserved.
```
annxingyuan: LLC.

src/kernels/webgl/reverse_packed_gpu.ts
```diff
@@ -0,0 +1,80 @@
...
+  constructor(xShape: number[], axis: number[]) {
+    const rank = xShape.length;
+    if (rank > 4) {
+      throw new Error(
+          `WebGL backend: Reverse of rank-${rank} tensor is not yet supported`);
+    }
+    this.outputShape = xShape;
+    const channels = getChannels('rc', rank);
+    const nextColumn =
+        `++${channels[rank - 1]} < ${this.outputShape[rank - 1]}`;
```
annxingyuan: I realize that we have this pattern elsewhere in the packed shaders, but could we instead test for adding 1 to the coordinate rather than mutating the coordinate? That way we don't have to remember to clean up after ourselves and decrement the coordinate. I think it'll be a little bit easier to read as well.

src/kernels/webgl/reverse_packed_gpu.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    };
+    const inCoords = xShape.map((_, i) => getInCoord(i)).join(',');
+    const innerDims = rank === 1 ?
+        'rc' :
+        `vec2(${xShape.map((_, i) => getInCoord(i)).slice(-2).join(',')})`;
+    const type = getCoordsDataType(rank);
+    const getc = `getChannel(getX(${inCoords}), ${innerDims})`;
+
+    const upperRow = `
```
annxingyuan: What do you think about restructuring this shader to match the unpacked shader so that all the logic for rank = 1 is nested under one if-statement? I wonder whether it might be easier to read that way - avoids having to test for rank = 1 in so many places. If we did it this way we should also be able to inline `upperRow` and `lowerRow` in the body of `void main` below.

src/kernels/webgl/reverse_packed_gpu.ts
jgartman: Done

src/kernels/webgl/reverse_packed_gpu.ts
jgartman: Done

src/kernels/webgl/reverse_packed_gpu.ts
jgartman: Thanks @annxingyuan, I refactored the code a bit and updated the compiled shader code at the top of the PR as well.  Let me know what you think.

src/kernels/webgl/reverse_packed_gpu.ts
jgartman: Done

annxingyuan:
Thanks @jgartman !

PR #1622, opened 2019-03-12
support begin, end and strides with length smaller than the input rank

pyu10055:
fixes https://github.com/tensorflow/tfjs/issues/1355

When the length of begin, end or strides array is small than the input rank, the output size calculation loop never end.

The fix is to set the default values for all of above if any of them is not set for a given axis.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1622)


leemengtaiwan:
Sorry I know you're very busy, but I'm wondering when the fix will be merged and be available in upcoming tfjs version?

The issue described in #1355 still exist though it's being closed.

dsmilkov:
Hi @leemengtaiwan, we should get a patch version of tfjs in a week or so. I reopened https://github.com/tensorflow/tfjs/issues/1355 which this PR will close.

PR #1627, opened 2019-03-16
ADD: maxpool3d op and cpu kernel

kaczmarj:
Hello, this PR proposes adding 3D max pooling. I have implemented the CPU kernel (without the gradients) and tests.

This PR is driven by my group's desire to use the three-dimensional U-Net in TensorFlow JS. This architecture uses max-pooling in the encoding process.

I have one question about the changes I have made. I found two implementations of `eitherStridesOrDilationsAreOne`, one private in `ops/conv` and one exported in `ops/conv_util`. The private implementation works with 1D, 2D, and 3D conv layers, whereas the exported version in `/ops/conv_util` only works with 1D and 2D layers. Would it be acceptable to have one implementation that works with 1D, 2D, and 3D layers? Basically the single implementation being:

```typescript
export function eitherStridesOrDilationsAreOne(
    strides: number|[number, number]|[number, number, number],
    dilations: number|[number, number]|[number, number, number]): boolean {
  return tupleValuesAreOne(strides) || tupleValuesAreOne(dilations);
}
```

Thank you,
Jakub

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1627)


dsmilkov:
Hi @kaczmarj,

Thank you for the invaluable contribution. Unfortunately, we already merged another PR: https://github.com/tensorflow/tfjs-core/pull/1778 that adds the same functionality. We merged that first because we processed PRs out-of-order. We will make best effort next time to make sure that we can review PRs in-order to avoid duplicate work. Cheers!

PR #1634, opened 2019-03-19
Export MathBackendCPU used in tfjs-data node test

kangyizhang:
Export MathBackendCPU from kernels/backend_cpu. This is used to set test environment in node test.

https://github.com/tensorflow/tfjs/issues/1391

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1634)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,6 +91,7 @@ export {
 
 // Backend specific.
 export {KernelBackend, BackendTimingInfo, DataMover, DataStorage} from './kernels/backend';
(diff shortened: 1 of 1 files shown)


dsmilkov:
Hi, exposing the MathBackendCPU to the public API is not a good long term solution since we want to be able to fully modularize core (decouple the webgl and cpu backend). I sent you a workaround in https://github.com/tensorflow/tfjs-core/pull/1635 which you can do in data after we release a patch of core.

kangyizhang:
Got it. Thanks for sending the other PR.

PR #1638, opened 2019-03-20
Clean up : update webgl cropAndResize code for readability

piercus:
While reading the code, I've noticed x and y was switched in `src/kernels/webgl/crop_and_resize_gpu.ts`.

For readability, i suggest to use x for width/right/left and y for height/top/bottom parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1638)


piercus:
> RC = RowColumn and CR = ColumnRow

Ok, now i understand why this x <-> y switch was not a problem !

I have updated it anyway, to make it easier for other newbies like me :-)

PR #1639, opened 2019-03-20
show better error message for not converted TFHub module

pyu10055:
moved TFHub url update logic to browser http handler and show module not converted message when 404 is returned.
fix issue tensorflow/tfjs#1408

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1639)


nsthorat:
Is this something we could ask for from the TF Hub team so that we can show a better error message? It's unfortunate we can't differentiate between a 404 (wrong URL) and an unconverted model

nsthorat:
Gentle ping

PR #1640, opened 2019-03-21
Modify im2col to use packed texture inputs.

annxingyuan:
Previously there were two problems:
- Failed to explicitly pass packed texture as output to `im2col`. This means `compileAndRun` marked the output texture as unpacked, so matMul would wastefully try to pack it before running (so we ran the pack shader even though it did nothing because the data was already packed).
- There was no version of `im2col` that consumed packed textures, so we had to unpack textures coming from previous packed ops.

I created a packed version of `im2col` in this PR, which solves both problems.

### Benchmarks

```
const x = tf.randomNormal([224, 224, 8]);
const w = tf.randomNormal([5, 5, 8, 16]);
tf.conv2d(x, w, 1, 'same');
```

Running the previous conv2d on pixel3: `104 / 90` to `63 / 42` *(1.6x)*

MobileNet pixel3: `126 / 91` to `113 / 82` *(1.1x)*
ResNet MacBook: `62 / 56` to `34 / 31` *(1.8x)*
ResNet iPhone: `165 / 159` to `83 / 79` *(2x)*
ResNet Pixel3: `199 / 182` to `131 / 93` *(1.5x)*

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1640)


dsmilkov:
Really nice work! I can't believe we got ResNet to be ~30ms on a laptop and ~80 on iPhone.

PR #1642, opened 2019-03-21
Remove redundant 'break'

rkinabhi:
The while iteration will stop when the flag `done` is set to _true_. The requirement of a break statement here doesn't seem necessary. 

An alternative fix would be to remove the assignment of `done` to _true_ instead. 😄

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1642)


nsthorat:
Thanks for the PR!

PR #1643, opened 2019-03-23
Packed LRN

jgartman:
Adds a packed kernel for localResponseNormalization:

Benchmark results:

| input shape        | unpacked (ms)           | packed (ms)  | update 3/31/19 packed (ms)| update 4/6/19| update 4/14/19|
| ------------- |:-------------:| -----:| -----:|-----:|-----:|
| [2, 32, 32, 3]      | 5.357 | 6.028 |5.310|5.504|5.397|
| [2, 64, 64, 3]      | 5.887      |   5.726 | 5.857|5.671|5.939|
| [2, 128, 128, 3]| 9.934     |    10.721 |  10.733|10.953|11.278|
| [2, 256, 256, 3]      | 25.528      |   16.705 | 16.715|16.726|16.311|
| [2, 512, 512, 3]      | 77.978      |   48.720 |48.142|48.875|48.600|

System info:

```
Graphics/Displays:

    Radeon Pro 560:

      Chipset Model: Radeon Pro 560
      Type: GPU
      Bus: PCIe
      PCIe Lane Width: x8
      VRAM (Dynamic, Max): 4096 MB
      Vendor: AMD (0x1002)
      Device ID: 0x67ef
      Revision ID: 0x00c0
      ROM Revision: 113-C980AJ-927
      VBIOS Version: 113-C9801AU-A02
      EFI Driver Version: 01.00.927
      Automatic Graphics Switching: Supported
      gMux Version: 4.0.29 [3.2.8]
      Metal: Supported, feature set macOS GPUFamily2 v1
      Displays:
        Color LCD:
          Display Type: Built-In Retina LCD
          Resolution: 2880 x 1800 Retina
          Framebuffer Depth: 30-Bit Color (ARGB2101010)
          Main Display: Yes
          Mirror: Off
          Online: Yes
          Rotation: Supported
          Automatically Adjust Brightness: No

    Intel HD Graphics 630:

      Chipset Model: Intel HD Graphics 630
      Type: GPU
      Bus: Built-In
      VRAM (Dynamic, Max): 1536 MB
      Vendor: Intel
      Device ID: 0x591b
      Revision ID: 0x0004
      Automatic Graphics Switching: Supported
      gMux Version: 4.0.29 [3.2.8]
      Metal: Supported, feature set macOS GPUFamily2 v1
```

Benchmark code:

```
export class LRNBenchmark implements BenchmarkTest {
  async run(size: number): Promise<number> {
    tf.setBackend('webgl');
    tf.ENV.set('WEBGL_PACK', true);
    const t: tf.Tensor<tf.Rank.R4> = tf.randomNormal([2, size, size, 3]);

    const benchmark = () => tf.localResponseNormalization(t, 2) as tf.Tensor;

    const time = await util.benchmark(benchmark);

    t.dispose();

    return time;
  }
}
```

Compiled shader code for LRN of tensor with input shape of `[2, 3, 3, 8]` and a depth radius of 2:
```
      void main() {
        ivec4 coords = getOutputCoords();
        int b = coords.x;
        int r = coords.y;
        int c = coords.z;
        int d = coords.w;

        bool hasNextCol = d < 8;
        bool hasNextRow = c < 3;

        vec4 sum = vec4(0.);
        vec4 xFragAtOutputCoords = getX(b, r, c, d);

        vec4 xAtOutputCoords =
          xFragAtOutputCoords *
            vec4(1, hasNextCol, hasNextRow, hasNextRow && hasNextCol);

        int firstChannel = d - 2;
        vec2 cache = vec2(0.);
        if(firstChannel >= 0){
          vec4 firstChannelFrag = getX(b, r, c, firstChannel);
          cache.x = firstChannelFrag.x;
            if(hasNextRow){
              cache.y = firstChannelFrag.z;
            }
        }

        ivec2 depth = ivec2(d, d + 1);
        for (int j = - 2; j <= 2; j++) {
          ivec2 idx = depth + j;
          bvec2 aboveLowerBound = greaterThanEqual(idx, ivec2(0));
          bvec2 belowUpperBound = lessThanEqual(idx, ivec2(7));

          bool depthInRange = aboveLowerBound.x && belowUpperBound.x;
          bool depthPlusOneInRange = aboveLowerBound.y && belowUpperBound.y;

          if(depthInRange || depthPlusOneInRange){
            vec4 z = vec4(0.);
            vec4 xFragAtCurrentDepth;
            z.xz = cache.xy;
            if(depthPlusOneInRange && hasNextCol){
              xFragAtCurrentDepth = (idx.y != d || idx.y != d + 1) ?
                getX(b, r, c, idx.y) : xFragAtOutputCoords;
              bool idxYEven = imod(idx.y, 2) == 0;
              z.y = idxYEven ? xFragAtCurrentDepth.x :
                xFragAtCurrentDepth.y;
              if(hasNextRow){
                z.w = idxYEven ? xFragAtCurrentDepth.z :
                  xFragAtCurrentDepth.w;
              }
            }
            cache.xy = z.yw;
            sum += z * z;
          }
        }
        vec4 result = xAtOutputCoords * inversesqrt(float(1) + float(1) * sum);
        setOutput(result);
      }
    

```

Just had a couple of questions:

1) Since it looks like according to the benchmark above performance is the same or possibly even has a slight regression up until N=128, should I add a heuristic to run the unpacked kernel for smaller input shapes?
2) I lumped LRN in with batch normalization in the environment flags but not sure if this is appropriate.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1643)


annxingyuan:
Hey @jgartman thanks for this contribution! Looking at your packed shader implementation, it seems possible that many of the sampling calls in the inner loop can actually be shared. For example, these two lines (lrn_packed_gpu.ts:80):

```
z.x = depthInRange ?
              getChannel(getX(b, r, c, idx.x), vec2(c, idx.x)) : 0.0;
z.y = depthPlusOneInRange && hasNextCol ?
              getChannel(getX(b, r, c, idx.y), vec2(c, idx.y)) : 0.0;
```

This might actually be sampling the same texel twice (depending on `idx.x` and `idx.y`). Same goes for the 3rd and 4th sampling calls in the inner loop.

Is it possible that by caching these calls so you never sample the same texel twice you might get better benchmark numbers? If so, we may want to write some specialized packed logic so we sample the minimum amount possible - our matmul (https://github.com/tensorflow/tfjs-core/blob/master/src/kernels/webgl/mulmat_packed_gpu.ts) and depthwiseConv (https://github.com/tensorflow/tfjs-core/blob/master/src/kernels/webgl/conv_packed_gpu_depthwise.ts) shaders both do this. 

Let me know what you think. Thanks!

jgartman:
Thanks @annxingyuan, you're right.  I removed unnecessary calls to the sampler and updated the benchmark above with the results and also updated the compiled shader code.  Surprisingly on my machine it doesn't seem to make much difference.  I didn't post the code for this above but as an experiment I also tried rewriting the shader with fewer conditionals with the tradeoff of more logic in each conditional branch and that didn't seem to help performance either.

Just to give you a bit more context, I don't have a specific use case that's motivating this PR so if it wouldn't add much value then we can always just close it for now.

annxingyuan:
Hey @jgartman - thanks for going down this rabbit hole :)

I still think there may still be a few redundant sampling calls in the shader. Basically every time we call `getChannel` we incur a cost, even if we're sampling the same location. That means, for example, in these lines:

```
vec4 xAtOutputCoords = vec4(
          getChannel(getX(b, r, c, d), vec2(c, d)),
          hasNextCol ?
            getChannel(getX(b, r, c, d + 1), vec2(c, d + 1)) : 0.0,
          hasNextRow ?
            getChannel(getX(b, r, c + 1, d), vec2(c + 1, d)) : 0.0,
          (hasNextRow && hasNextCol) ?
            getChannel(getX(b, r, c + 1, d + 1), vec2(c + 1, d + 1)) : 0.0
        );
```

If `d` is even, then the first two `getChannel` calls are sampling the same location. If `c` is even, then all four `getChannel` calls are sampling the same location. To get around this we would check for even / odd-ness of those last two dimensions before calling `getChannel` again.

The same goes for the other `getChannel` calls in this shader. 

Does this make sense? We store data in packed textures such that each texel contains only values from the same two innermost dimensions (I'll refer to these as the rows and columns) of the underlying tensor. This means the `r` channel of a packed texel always contains the value from an even row and even column, the `g` channel contains the value from the same row but the next column over, the `b` channel contains the value from the next row and the same column, and the `a` channel contains the value from the next row and the next column. That's why it matters whether the last two inputs to `getChannel` are even vs odd in determining whether it's a redundant sample.

Hopefully if you include logic to avoid redundant `getChannel` calls you'll see a performance improvement. 

Thanks for the context - would still be awesome to get this in - someone will use it :)

jgartman:
Thanks again for your input @annxingyuan and the explanation in your previous post, I really appreciate it.  I updated the benchmark and the compiled shader code again.  Unless I'm misunderstanding, the inner dims of the coordinates returned by `getOutputCoords` should always be even which simplifies things a bit in this case but let me know if I'm wrong there.  Unfortunately, the performance was about the same as before. I think I was a bit hasty in assuming that the packed version of this op was going to be clearly better in all cases.

src/kernels/webgl/lrn_packed_gpu.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        vec4 xAtOutputCoords = vec4(
+          getChannel(xFragAtOutputCoords, vec2(c, d)),
+          hasNextCol ?
+            getChannel(xFragAtOutputCoords, vec2(c, d + 1)) : 0.0,
+          hasNextRow ?
+            getChannel(xFragAtOutputCoords , vec2(c + 1, d)) : 0.0,
+          (hasNextRow && hasNextCol) ?
+            getChannel(xFragAtOutputCoords, vec2(c + 1, d + 1)) : 0.0
+        );
```
annxingyuan: Could you try this:

```
vec4 xAtOutputCoords = xFragAtOutputCoords * vec4(1, hasNextCol, hasNextRow, hasNextRow && hasNextCol);
```

src/kernels/webgl/lrn_packed_gpu.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        int firstChannel = d - ${rad};
+        vec2 cache = vec2(0.);
+        if(firstChannel >= 0){
+          vec4 firstChannelFrag = getX(b, r, c, firstChannel);
+          cache.x = getChannel(firstChannelFrag, vec2(c, firstChannel));
```
annxingyuan: You should know at compile time whether `firstChannel` is odd or even - so you could splice in a value for `cache.x` without calling `getChannel`. e.g. if `rad` is even, then `firstChannel` is even (because `d` is always even). In which case you could do:

`cache.x = firstChannelFrag.x`

If it's odd, then it should be:

`cache.x = firstChannelFrag.y`

Same goes for computing `cache.y` - it will be `firstChannelFrag.z` if `firstChannel` is even, and `firstChannelFrag.w` if `firstChannel` is odd.

src/kernels/webgl/lrn_packed_gpu.ts
```diff
@@ -0,0 +1,111 @@
...
+          bool depthInRange = aboveLowerBound.x && belowUpperBound.x;
+          bool depthPlusOneInRange = aboveLowerBound.y && belowUpperBound.y;
+
+          if(depthInRange || depthPlusOneInRange){
+            vec4 z = vec4(0.);
+            vec4 xFragAtCurrentDepth;
+            z.xz = cache.xy;
+            if(depthPlusOneInRange && hasNextCol){
+              xFragAtCurrentDepth = idx.y != d ?
+                getX(b, r, c, idx.y) : xFragAtOutputCoords;
```
annxingyuan: If `idx.y` equals `d` plus one, this will be a redundant sampling call.

src/kernels/webgl/lrn_packed_gpu.ts
```diff
@@ -0,0 +1,111 @@
...
+              z.y = getChannel(xFragAtCurrentDepth, vec2(c, idx.y));
```
annxingyuan: Here since `c` is always even, this will always be either `xFragAtCurrentDepth.x` or `xFragAtCurrentDepth.y`. If `idx.y` is even, then it's x, if it's odd, then it's y. You can compute that conditional directly rather than using `getChannel`. Same logic goes for computing `z.w` below.

jgartman:
Thanks @annxingyuan I went ahead and tried out the remaining suggestions but unfortunately I wasn't able to get better performance than before.  I updated the benchmark and shader code.  As far as readability, would we want to use the version from commit 556502b that still has all of the optimizations except for the most recent ones?  Hope this review didn't end up being a bunch of work for you and thanks for all your input.

annxingyuan:
Hey @jgartman - thank you so much for trying out those optimizations! It's too bad we didn't see any performance gains. Yes, version 556502b sounds good! I'll merge this once you revert this last change.

Thanks again!

jgartman:
@annxingyuan I think this should be ready now

annxingyuan:
Thank you @jgartman !

PR #1644, opened 2019-03-25
Make the save function in a gradient be a Tensor[] instead of a NamedTensorMap

nsthorat:
This makes Google Closure happy because index types cannot be accessed with a property accessor.

Also update property accessor that have string index types to use string indexing (also to make closure happy).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1644)


nsthorat:
Verified this syncs and works in google3.

PR #1663, opened 2019-04-05
Add tf.util.fetch.

annxingyuan:
This PR fixes https://github.com/tensorflow/tfjs/issues/1439

#### Changes
- Added `tf.util.fetch`
- Add `node-fetch` as dependency
- Remove code that looks for global fetch in `BrowserHTTPRequest`
- Spy on `tf.util` rather than `global` for IO tests

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1663)


annxingyuan:
@nsthorat Would you mind taking a look at the node-fetch test? Not sure whether it's what you had in mind. I'm spying on `global.fetch` - I'm not sure how to spy on `node-fetch` since it just exports a single function without a parent object - but the only way `global.fetch` gets defined as a function is if `node-fetch` is imported.

annxingyuan:
@nsthorat I added the code for the test but it fails because `node-fetch` just exports a function - there's no parent to spy on. Or am I getting the syntax wrong?

PR #1666, opened 2019-04-05
Add a script to measure the bundle size.

nsthorat:
This adds a CI that measures the impact of the change against master for the bundle size.

The results below are for dropping backend CPU from index.ts.

```
yarn test-bundle-size

~~~~minified bundle~~~~
==> post-gzip
  diff:    -12.25 K  (-10.96%)
  master:  111.73 K
  change:   99.48 K

==> pre-gzip
  diff:    -45.68 K  (-9.83%)
  master:  464.95 K
  change:  419.26 K


~~~~unminified bundle~~~~
==> post-gzip
  diff:    -20.00 K  (-8.67%)
  master:  230.60 K
  change:  210.60 K

==> pre-gzip
  diff:   -145.45 K  (-9.91%)
  master: 1468.07 K
  change: 1322.62 K


~~~~esm bundle~~~~
==> post-gzip
  diff:    -12.31 K  (-11.05%)
  master:  111.39 K
  change:   99.08 K

==> pre-gzip
  diff:    -45.67 K  (-9.84%)
  master:  464.35 K
  change:  418.67 K

Done in 119.29s.
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1666)


nsthorat:
At the publish level as well as PR-level I think diffing would be useful inside Cloud CI.

I can set that up if people think it's useful.

Also, I can put it in build-npm, that makes sense. Let's first answer whether we want this to just be an automatic check on cloud instead.

dsmilkov:
Getting the gzip numbers as an FYI upon explicitly calling a script is not so useful since we have bundlephobia.comCloud. However automatically doing it on CI + diffing with master (your PR added x.x KB to the bundle) sounds very useful.

nsthorat:
Yeah I was thinking it would be done as you're trying to drop things from the library & measure without waiting for publishing to NPM. But a CI solves that, so I'll do it there.

nsthorat:
Okay, updated this so it runs on a CI as a separate thing. Here is an example with me adding two methods (dropping the backends doesn't quite work with CI yet):

https://pantheon.corp.google.com/cloud-build/builds/61efaaed-60d3-4ae5-8308-8243cfd3b472?project=learnjs-174218

nsthorat:
The only *slight* possible issue is this takes a really long time, about the same length as the tests themselves :/

nsthorat:
On the last run it's now taking a minute longer than the tests themselves.

Argh.

PR #1668, opened 2019-04-06
Update rollup to latest and add ability to visualize bundle

dsmilkov:
Add ability to visualize our browser bundle.

Screen Shot 2019-04-06 at 1 15 05 PM

To produce a visualization, call `yarn rollup -c --visualize`
. It will output `dist/tf-core.min.js.html` with the breakdown. 

Details:
- Update rollup and its plugins to latest so we have this ability.
- Change the `build-npm.sh` script to visualize so we always publish the breakdown on npm. That way we can visit http://unpkg.com/@tensorflow/tfjs/dist/tf-core.min.js.html and see the breakdown for any published version.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1668)


nsthorat:
BTW - post-gzip is an important number (since some stuff is very repetitive in the library). I wonder if we can just drop my change and add a way to see the post-gzip total size with this flow. See this change: https://github.com/tensorflow/tfjs-core/pull/1666

nsthorat:
One more question: do you know if unpkg will render the HTML file as HTML or as text? I wonder if we could instead add this as a build artifact in GCP

PR #1670, opened 2019-04-09
Slice the result of read and readSync to avoid mutating internal state.

nsthorat:
See: https://github.com/shiffman/Tensorflow-JS-Examples/issues/7

This avoid users dataSyncing and mutating internal state. Slice is shallow so is very cheap. I ran the benchmarks to make sure, there is no change.

I also updated the benchmarks script to use 1.0 code and hit TF Hub.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1670)

### DIFF
--- a/benchmarks/index.html
+++ b/benchmarks/index.html
@@ -166,7 +166,7 @@
       <tbody></tbody>
     </table>
   </div>
-  <script src="https://unpkg.com/@tensorflow/tfjs-core/dist/tf-core.js"></script>
+  <script src="../dist/tf-core.js"></script>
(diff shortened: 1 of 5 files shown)


nsthorat:
Everything's broken in CPU backend, lots of assumptions about mutating internal state. Working on the fix.

nsthorat:
Okay, fixed. Ready for review.

nsthorat:
Also this doesn't slice the call into backend.write -- what do you think about that?

nsthorat:
Dropping this change since it has performance impacts. Slice is shallow but not cheap.

PR #1683, opened 2019-04-18
Fix packed 1d reverse bug

jgartman:

Fixes https://github.com/tensorflow/tfjs/issues/1518.

Root cause: Passing the wrong inner dims to the `getChannel` function.  As an example to demonstrate the problem suppose we have `tf.tensor1d([0, 1, 2, 3])` and we're trying to find the value that goes at coordinate 0 on the output texture.  The input tensor will get stored on a packed texture as follows:
```
0|1  2|3
---  ---
x|x  x|x
```
The reversed coordinates were being incorrectly calculate as:
```
          result.r = getChannel(getX(4 - 0 - 1), 0);
          if(0 + 1 < 4){
              result.g = getChannel(getX(4 - (0  + 1) - 1), (0  + 1));
          }
```
For the r channel this would sample from the first column of the second texel and return 2 instead of 3.  After this fix is applied `result.r` would be `getChannel(getX(4 - 0 - 1), 4 - 0 - 1)` which will sample 3.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1683)

### DIFF
--- a/src/backends/webgl/reverse_packed_gpu.ts
+++ b/src/backends/webgl/reverse_packed_gpu.ts
@@ -42,9 +42,11 @@ export class ReversePackedProgram implements GPGPUProgram {
         void main(){
           int rc = getOutputCoords();
           vec4 result = vec4(0.);
-          result.r = getChannel(getX(${xShape[0]} - rc - 1), rc);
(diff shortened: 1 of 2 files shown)


nsthorat:
Thanks @jgartman!

PR #1693, opened 2019-04-20
Fix the result of ones with complex type

Lewuathe:
`ones` ops should return the value with zeros in the imaginary part to be consistent with TensorFlow core.

```python
a = tf.ones([2, 2], dtype='complex64')
print(a.eval())

# [[1.+0.j 1.+0.j]
#  [1.+0.j 1.+0.j]]
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1693)


nsthorat:
Thanks!

PR #1694, opened 2019-04-21
Add linspace kernel

Lewuathe:
`linspace` ops should be implemented as kernel to make it bindable with various kind of backend.

See: https://github.com/tensorflow/tfjs/issues/1211

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1694)


Lewuathe:
It seems to fail in Safari test cases due to numerical precision.

```
Step #1 - "test": Safari 11.1.2 (Mac OS X 10.13.6) linspace webgl {} start stop FAILED
Step #1 - "test": Error: Arrays differ: actual[5] = 6.000000476837158, expected[5] = 6.
Step #1 - "test": Actual: 1,2,3,4,5,6.000000476837158,7,8,9,10.
Step #1 - "test": Expected: 1,2,3,4,5,6,7,8,9,10. in /tmp/9a7710e58516efb4655f1e59f3cbd6d0.browserify.js (line 69669)
```

nsthorat:
How about for now we add this as a kernel but in WebGL land we just do it on the CPU? This way we can bind to the libtensorflow kernel for linspace but we don't have precision problems on safari.

Lewuathe:
Sounds good. Will do.

Lewuathe:
But still not sure why this kind of precision problem happens in Safari...

Lewuathe:
@nsthorat I updated. Could you take a look when you get a chance?

Lewuathe:
Thanks for reviewing.

> Once you've merged this, do you think you could also add this code in the Node.js backend after the release?

Okay, I'll try to do that.

PR #1695, opened 2019-04-22
WebGPU: Increase work per thread for matMul.

annxingyuan:
#### Changes
- increase work per thread in matMul kernel

#### Benchmarks

These are minimum times across 20 runs for `AOuter` = `shared` = `BOuter` = 500.

tile size = 8, work per thread = 4: 86ms

tile size = 8, work per thread = 2: 52ms

tile size = 8, work per thread = 1: 65ms

tile size = 16, work per thread = 4: 38ms

tile size = 16, work per thread = 2: 119ms

tile size = 16, work per thread = 1: 100ms

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1695)


austinEng:
Here are my thoughts on why adding `workPerThread` doesn't have a consistent effect on performance. You're currently using `workPerThread` so that each thread loads a `workPerThread x workPerThread` 2d sub-matrix into the shared memory region. Since the matrices are row-major the accesses to main memory are probably hitting different cache lines for different rows.

You could try having workPerThread only iterate over loads across a single row which might help. Then your local threadgroups would be more like:
```
layout (
        local_size_x = TileSize, 
        local_size_y = TileSize / WorkPerThread, 
        local_size_z = 1) in;
```

dsmilkov:
Curious, what are the WebGL numbers for 500x500? (no need to do benchmarks for this, just curious if you know on top of your head)

PR #1712, opened 2019-04-28
Fix docs argument order for fused.matMut

iislucas:
The argument order in the comment for [`fused.matMut`](https://js.tensorflow.org/api/latest/#fused.matMul) doesn't match the argument order in the code, resulting in this error when you try to run the example code: 

```
An error occured on line: 5
Argument 'bias' passed to 'fused matMul' must be numeric tensor, but got string tensor
```

This PR fixes the argument order in the comment to match the code, and renames 'c' to 'bias' in the comment to make it clearer that that argument is the bias argument. It also adds `.print();` to the computed tensor so that we can see the result.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1712)


nsthorat:
Thanks for the PR @iislucas!

PR #1713, opened 2019-04-29
Add tf.signal.frame() op.

Lewuathe:
Add frame op as provided in [TensorFlow core](https://www.tensorflow.org/api_docs/python/tf/signal/frame). It's also needed to implement stft, istft ops to make slices of the sample.

See: https://github.com/tensorflow/tfjs/issues/1362

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1713)


nsthorat:
See this for failures: https://pantheon.corp.google.com/cloud-build/builds/ecf44c74-fa85-4c83-ae90-7f42c9d0c603?project=learnjs-174218

Lewuathe:
> See this for failures: https://pantheon.corp.google.com/cloud-build/builds/ecf44c74-fa85-4c83-ae90-7f42c9d0c603?project=learnjs-174218

Thanks but unfortunately I do not have permission to look into that. But anyway, I guess the problem is fixed.

nsthorat:
Ah apologies that's the internal version of the link, you can see the failure by clicking details:

image

nsthorat:
Ah looks like we may have a bug, I think this will fix it: https://github.com/tensorflow/tfjs-core/pull/1737

nsthorat:
Thanks for the awesome PR as always @Lewuathe :)

PR #1714, opened 2019-04-29
Save optimizers

caisq:
FEATURE 

- Change the `className`s of the optimizers align with TensorFlow (Python)
- Use arrays to explicitly order the optimizers' weights to facilitate weight loading and saving
- Implement `getWeights()` for weight saving
- Implement `setWeights()` for weight loading

Towards https://github.com/tensorflow/tfjs/issues/83

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1714)


caisq:
s/are both offline/are both async/

PR #1722, opened 2019-05-01
webgpu: add a large matmul test

kainino0x:
Passes on both webgl and webgpu backends, so apparently they match.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1722)


annxingyuan:
@kainino0x wouldn't 32x32 be enough to test >1 tile?

kainino0x:
Right now we have 16x16 tiles and WPT=2, which means 33x33 would be enough. But if we raise WPT to 4 or tile size to 32 on hardware that supports it, 65x65 would be necessary.

PR #1723, opened 2019-05-01
webgpu: matmul_packed: double buffer reads

kainino0x:
With 16x16 tile sizes:

| version | WPT=2 | WPT=4
| --- | --- | ---
| single buffered (1 shared mem) (#1731) | 3.1ms | 2.5ms
| double buffered (2 shared mem) (#1731+#1723) | 3.1ms | 2.9ms

WebGL: 1.7ms.

MacBookPro15,2/Intel Iris Plus 655

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1723)

### DIFF
--- a/src/backends/webgpu/src/flags_webgpu.ts
+++ b/src/backends/webgpu/src/flags_webgpu.ts
@@ -24,4 +24,4 @@ ENV.registerFlag('WEBGPU_IMMEDIATE_EXECUTION_ENABLED', () => true);
  * Thread register block size for matmul kernel. If 0, we use the version of
  * matMul without register blocking.
  */
-ENV.registerFlag('WEBGPU_MATMUL_WORK_PER_THREAD', () => 2);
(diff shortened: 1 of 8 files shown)


annxingyuan:
@kainino0x I assume these are for 500x500 * 500x500 and 50 reps per readback?

kainino0x:
Yes, for the current version of the benchmark: 5 trials, 50 reps, 500x500 * 500x500
https://github.com/tensorflow/tfjs-core/blob/794758f2f71544577c0fccdac915610a41c44e7d/src/backends/webgpu/src/benchmark_ops_test.ts#L75

kainino0x:
After #1731, this is looking like it's making things worse. Closing for now.

kainino0x:
Had a thought. I think we don't need double buffering because our workgroup sizes (256) are considerably bigger than the hardware wavefront size (32). That's pretty much equivalent to double buffering (8x buffering, really) as long as the driver schedules it well (which I think we can rely on).

PR #1742, opened 2019-05-06
Use core tests in webgpu

dsmilkov:
Use core tests for WebGPU and remove any duplicated tests.

- Remove duplicated webgpu tests now that we expose core.
- Move any newly added webgpu tests to core so any backend can benefit.
- Exclude failing tests along with a comment on the reason of failure.
- Fix a bug in our vanilla CPU backend due to a newly added test in WebGPU which I moved to core (Yay test reuse!)
- Reduce console warnings due to debug mode in tests

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1742)


kainino0x:
FYI, I wrote some new tests in the conv2d_tests - 4 or so that aren't in the original tests (you should be able to tell by name). When you do this refactor you should probably copy those back out to the original test file, if you weren't already doing this.

dsmilkov:
@kainino0x Sounds good. I moved any original tests in core so any backend (not just webpgu) can benefit. This PR is ready for review.

PR #1749, opened 2019-05-08
Add booleanMask op

syt123450:
FEATURE

Add tf.booleanMask op. Feature requested in [tensorflow/tfjs#380](https://github.com/tensorflow/tfjs/issues/380).

Reference:
* [tf.boolean_mask documentation](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/boolean_mask?hl=en)
* [tf.boolean_mask tensorflow python implementation](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/ops/array_ops.py#L1274)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1749)


nsthorat:
Thanks for the PR! Do you know about our GCP test check? If you look above you will see "Trigger: ed916b23-8e01-4fda-85af-b7c8627d15a1 Failing after 1m — Summary". If you click "Details" you will see what the failure is.

You should get that to pass before I can merge, thanks!

nsthorat:
Once you fix this LGTM and I will merge!

In the future though it would be good to bug us about ops since we don't want to just add ops for ops sake, they should be used in a model :)

nsthorat:
Oh looks like someone asked for it, nevermind :)

mx601595686:
Is it`s possible to add set function, this method can only get elements, but I want to get and modify and save back to its original place, this will be very useful when doing image processing.

syt123450:
Sorry for late reply, thanks for your following PR @nsthorat !

PR #1750, opened 2019-05-10
WIP Shader cache improvements

tafsiri:
Not ready for review.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1750)


tafsiri:
PTAL at this PR. In particular if you notice anything that would clearly be a hit to shader run time that would be great to point out (correctness stuff too).

## Results

[This spreadsheet](https://docs.google.com/spreadsheets/d/1tIYbM-LSOOV-lGFIDeY5KtBeESdCByJ2dQfi1hvV3Q8/edit?usp=sharing) has some of the results of how this change would impact existing models like mobilenet and posenet assuming no changes to the cache key (so just using uniforms in the programs). There is some non trivial performance regression. (Not included is the potential benefit to universal-sentence-encoder). 

## Discussion

This PR changes the snippets generated by the shader compiler to use uniforms to hold shape (and shape related) information but leaves the key the same, this is to allow us to more clearly evaluate the impact of each change (moving to the new key would improve cache hits, but for models that 'warmup' quickly that isn't all that helpful). There are still a few todos and some commented out code. But before going any further the general approach needs to be evaluated since it is causing a penalty to existing models.

From profiling in Chrome most of the new time is spent in readPixels, suggesting that shader execution time is increasing. From initial discussion @dsmilkov suspects that swapping literals for variable access removes some of the optimizations the compiler can do to the program before it is ever run. 

If there is an obvious error/issue in the PR then that would be great to try and fix. Else we are left with looking at this tradeoff in cost (and potentially deciding what if any penalty we would be willing to accept). Its possible that there is a sweet spot between which values are inlined and which as put in as uniforms, but its hard to tell off the bat. Any thoughts on that would also be helpful.

Even if we don't move forward with this. I'd be happy to extract the updated benchmark code and make a separate PR with that.

nsthorat:
Can we close this out?

PR #1753, opened 2019-05-12
Fix gradients for onesLikes / zerosLike

caisq:
Fixes https://github.com/tensorflow/tfjs/issues/1398

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1753)


nsthorat:
FYI I am on vacation this week, this fix is fine for now but we should really fix this properly by stopping gradient propagation when we have null gradients @dsmilkov

dsmilkov:
https://github.com/tensorflow/tfjs/issues/619 tracks the special "null" for gradients.

PR #1754, opened 2019-05-13
Update tfjs-core to 1.2.0

dsmilkov:
INTERNAL

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1754)


dsmilkov:
Integration fails for layers, as expected. Fix is coming asap.

dsmilkov:
Good point. will bump the minor version

PR #1755, opened 2019-05-13
call arraybuffer API for ImageData

pyu10055:
This is to demonstrate the alternative options for fixing the tensor CPU copy slowness for GPU upload.
call 
void gl.texImage2D(target, level, internalformat, width, height, border, format, type, ArrayBufferView? pixels);

for ImageData object when the DESTRUCT_IMAGE_DATA flag is set.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1755)


dsmilkov:
Lint fails for circular deps. To avoid it, move `PixelData` to `src/types.ts`

nsthorat:
Looks like you merged without responding to my comments, can you send a follow up PR to address them? Thanks!

PR #1756, opened 2019-05-14
Publish .js files with test_snippets so that ts-node doesn't choke in dependent packages.

nsthorat:
This will put test_snippets.js and test_snippets.d.ts in dist.

The slightly weird thing about this PR is the *new* file structure of dist, but making dist only contain files as part of the public API means we need a new tsconfig for scripts.

```
./dist/src/index.js
./dist/src/ops/...
./dist/scripts/test_snippets/...
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1756)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,8 +3,8 @@
 src/**/*_test.ts
 integration_tests/
 src/backends/webgpu/
-dist/backends/webgpu/
-dist/backends/**/*_test.js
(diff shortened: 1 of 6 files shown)


nsthorat:
Closing this PR, I don't like the new structure of dist (we rely very heavily on this).

PR #1763, opened 2019-05-19
[tests] Add iOS 12 to BrowserStack test envs

sdll:
Resolves https://github.com/tensorflow/tfjs/issues/836.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1763)


nsthorat:
Thanks for the PR -- we're going to revisit this in the future but for now I'm going to close this because we're trying to move over to a monorepo on tfjs.

PR #1765, opened 2019-05-20
[webgpu] Add readSync just when data lives on the CPU so getParamValue works during model conversion.

annxingyuan:
The converter calls `readSync` to get parameter values but the tensors should always be CPU resident at that point.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1765)


nsthorat:
High-level question, is this possible to do in the converter directly? I don't think we want to assume this is possible for all backends.

annxingyuan:
@nsthorat I see your point - but I also think it makes sense that we would make `readSync` available for WebGPU in case there's a synchronous way to get the data, which there is when it's on the CPU. As a bonus, this would conveniently enable model conversion. 

What do you think?

PR #1769, opened 2019-05-25
Fix onProgress not passed in getLoadHandlers

syt123450:
BUG

* Fixes [tensorflow/tfjs#1610](https://github.com/tensorflow/tfjs/issues/1610).
* Adds a test case to ensure `onProgress` works well with `getLoadHandlers`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1769)


caisq:
@syt123450 I see. Let's get the tfjs-core PR reviewed and checked in first. In the meantime, you can add the test, we'll run it again once your fix in tfjs-core is released.

nsthorat:
Thanks for the fix!

PR #1771, opened 2019-05-27
Add initial support for loading tfjs in WebWorkers

WenheLI:
This PR is about a Webworker support feature.

## Features
- It adds support for Webwoker over CPU backend.
- It adds support for Webwoker over WEBGL backend.
- It adds support for webworker on `fromPixels` and `save`.

## Question
- How should we properly import `Webworker` related Objects inside `typescript`? Since the `Webworker` related objects have internal conflicts with original `dom` objects, we may need to find out a way to add them or just use `@ts-ignore` for now.
- How should I do the unit test in this case, should I build up a new set of test in the `Webworker` environment for all operations or just the effected operations? Or, is there any better way of achieving it?

After this PR is merged, I will move on to the tfjs-node side to add similar support for it. Just let me know, if I miss anything!

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1771)


WenheLI:
Hi - @tafsiri @dsmilkov 
Thanks for your suggestion! I have removed the ``webworker`` flag which surely makes the logic clearer. 
> would typeof(self) !== 'undefined' work in this case (to be consistent with the rest of the checks)?

For this, both of  ``webworker`` and ``browser`` environments have the ``self``  global variable but points to different instances. Thus, `` typeof(self) !== 'undefined'`` may not work in this case. Currently, I used the ``typeof(self) !== typeof(document)`` to detect if it is under ``webworker``, since ``document`` will be ``'undifined'`` in webworker.

PR #1775, opened 2019-05-29
Enable flag changes in exposed API functions

GantMan:
These public APIs were created back in this ticket: https://github.com/tensorflow/tfjs/issues/760

But the code assumes a person would never want to turn these off once triggered.  I don't want them turned on all the time, so exposing an optional boolean to set them back to false.

----------------

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1775)

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -26,10 +26,12 @@ import {getTensorsInContainer} from './tensor_util';
 /**
  * Enables production mode which disables correctness checks in favor of
  * performance.
+ *
+ * @param active The optional parameter if prod mode should be active.
  */
 /** @doc {heading: 'Environment'} */
-export function enableProdMode(): void {
-  ENV.set('PROD', true);
+export function enableProdMode(active = true): void {
+  ENV.set('PROD', active);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


nsthorat:
Hi @GantMan thanks for the PR.

Just curious what your use case for disabling these are? We're really not trying to use these flags as mutable state, they're supposed just be set at module initialization time (and I'm going to enforce that soon).

GantMan:
Hi @nsthorat - It's a noise to signal thing.

I'd like to enable verbose logging for a set of operations (a single one could, of course, use `profile`), to assist in potential debugging.  It's hypothetical, but reasonable that we wouldn't want EVERY operation to be logged, only a subset of my choice.

For perspective:  I'm providing a UI for enabling this.  And philosophically... what UI element can be turned on and never off?  Just hoping ot use it as a way to dig deeper if needed, and then turned back off.

nsthorat:
I don't think this completely solves the issue you want to be solved. If you only want a few ops to be logged I would definitely recommend using the programmatic `tf.profile()` and filtering the response (which should be a JSON object). This is our blessed way of doing what you want programatically. The global flags aren't meant to be switched in the middle of the program, only to setup TensorFlow.js at the beginning (in fact flags are cached aggressively so there's no guarantee this will even work).

GantMan:
That's fair.   I can deal with that.

Do you have a way for me to PR the docs (I didn't see it in a glance) where I could put this information for people after me?

nsthorat:
Here are the API docs, is this what you meant? https://js.tensorflow.org/api/latest/#profile

tafsiri:
@GantMan there are also some flag related docs here https://www.tensorflow.org/js/guide/platform_environment if you want to take a look and see if they need any improvement. The source for that is located [here](https://github.com/tensorflow/tfjs-website/blob/master/docs/guide/platform_environment.md#flags)

GantMan:
Thanks!!!

PR #1780, opened 2019-06-08
remove the variable exports in environment.ts

pyu10055:
currently the ENV variable is exported from environment.ts, and set in the engine.ts.
This is causing issues on some bundling system that copies the module on require('environment.ts') call, in which the set exports.ENV in setEnbironmentGlobal method is setting the variable in the origin module object not the copied module object.

```
export function setEnvironmentGlobal(environment: Environment) {
  ENV = environment;
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1780)


nsthorat:
For posterity, here is a post talking about immutable exports: http://2ality.com/2015/07/es6-module-exports.html

So this PR seems reasonable. We need to have two top level objects stuck on NS because:
- ENV needs to be available at file load time, thus it needs to check NS.
- ENGINE needs to be available at file load time. It can see ENV, but cannot stick itself on ENV because we want ENV to be a leaf.

Therefore, we must have two top-level objects stuck on NS.

Ping, let me know when you're ready to review.

nsthorat:
Just to clarify, that's only for commonjs modules. Any es6 modules will allow the current functionality, but we're not going to rely on that.

nsthorat:
Let's hold off on this PR until we understand where the build error is coming from and can reproduce it. All the current configs seem to work just fine :)

nsthorat:
Closing this since we solved this with the webpack config.

PR #1789, opened 2019-06-14
react native platform

tafsiri:
Adds an initial implementation of the react native platform adapter. Major components of platform are present and it has been tried out with the cpu backend in an example react native application (not included in this repo). Not finished, but feels like a decent chunk of code for a first PR.

Included:
 - Platform implementation (fetch)
 - Two IOHandlers
 - Unit test setup with mocks for react-native platform stuff. (see design doc for more details on the approach to using unit tests with mocks vs e2e testing in the host platform).

Changes to core include:
  - Exporting types related to the `tf.io` namespace
  - The code used to test if we are running in a browser.
  - A `responseType` header set to `arrayBuffer` is added to requests for weights. 

Things this does not include
 - Full unit tests, more will be added if we are generally happy with the mocking approach
 - The integration/e2e tests (in form of a react native app). Outlined in the design doc, not implemented yet. (Though as mentioned above I have tested this in a sample react native app I have outside this repo).
 - ModelManager registration or registration of save/load handers. Currently save and load handlers are created directly via a function. ModelManager registration can be added later as per the design doc.
 - Tests are not added to CI build yet as they need a new release of tfjs-core.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1789)


tafsiri:
I also moved this to the top level as per in person discussion of repo organization.

PR #1795, opened 2019-06-18
Add support for string weights

dsmilkov:
- Add support for deserializing weights of dtype `string`
- Allow string tensors in `tile` and `expandDims` ops (these ops are used by AutoML)
- Add utf-8 encoding/decoding capability to `Platform`

This is needed to enable execution of AutoML models, which store the vocab as a weight (`Const` tensor) of dtype `string`.

Corresponding change in tfjs-converter: https://github.com/tensorflow/tfjs-converter/pull/386

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1795)


dsmilkov:
I'm a googler and don't need a CLA

nsthorat:
LGTM, but wait for approval from Nick / Ping on the doc.

dsmilkov:
Thanks, yup will do

PR #1797, opened 2019-06-19
Move webgpu code to a top-level tfjs-webgpu folder.

nsthorat:
This makes the directory structure align with the NPM package structure more closely.

This has the added effect of making syncing to google3 much saner (one BUILD rule for each of the top-level NPM packages).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1797)


annxingyuan:
@nsthorat Looks like we might just need to add 'RFFT' to the webgpu exclude list to get tests to pass.

PR #1809, opened 2019-06-23
Fix bug in which seed for UniformRandom cannot be null

caisq:
Fixes https://github.com/tensorflow/tfjs/issues/1687
Fixes https://github.com/tensorflow/tfjs/issues/1685

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1809)


caisq:
@nsthorat I hope this bug fix can be included in 1.2.2. Without it, tf.layers.dropout() won't work.

PR #1812, opened 2019-06-25
No need outDepth when compute output shape in conv_util.computeOutputShape3D

axinging:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1812)


nsthorat:
I checked all the other repos for usage of this and couldn't find it. It's exported through backend_util which is new, but doesn't seem to be used anywhere except internally.

nsthorat:
Thanks for the PR!

PR #1822, opened 2019-06-28
[webgpu] Add buffer manager for buffer reuse.

annxingyuan:
This PR fixes https://github.com/tensorflow/tfjs/issues/1640

#### Changes
- Add buffer manager, analogous to WebGL texture manager, that recycles buffers
- Change disposal logic - immediately dispose if texture is not part of a command queue, otherwise wait until submitting the queue before disposing

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1822)


kainino0x:
___
*[tfjs-webgpu/src/backend_webgpu_test.ts, line 82 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1822#-LjMx0WsCokMyPkDRFRe:-LjRtabe7oUjK_RjDaC9:b9yje9f) ([raw file](https://github.com/tensorflow/tfjs-core/blob/9ba07250cb05a19c7d7712b3c05aea071b515dae/tfjs-webgpu/src/backend_webgpu_test.ts#L82)):*

async unit tests don't actually interleave, they run serially (but yeah, async can be removed here)
</blockquote></details>

Hm, I thought they did, but you're right. I had problems with this when I was trying to benchmark. I wonder if it was this timeout issue: https://github.com/jasmine/jasmine/issues/703#issuecomment-62965757

kainino0x:
___
*[tfjs-webgpu/src/backend_webgpu.ts, line 149 at r2](https://reviewable.io/reviews/tensorflow/tfjs-core/1822#-LjRqQSB9frxD2Pl8qkn:-LjT4KAZ6aQDGjFBl8eo:b71c6nw) ([raw file](https://github.com/tensorflow/tfjs-core/blob/dc04ffa8bd1505611bda91bfbd5f035657565103/tfjs-webgpu/src/backend_webgpu.ts#L149)):*
<details><summary><i>Previously, annxingyuan (Ann Yuan) wrote…</i></summary><blockquote>

It looks like there are different WebGPU buffer types depending on whether they need to be mappable, whether they are storing uniforms, etc. 
</blockquote></details>

Usages are used in WebGPU implementations to figure out the location and layout of some memory. E.g. mappable memory might be in GPU-visible CPU memory, or CPU-visible GPU memory, while non-mappable memory would be in GPU-private memory which is faster for many things.

PR #1842, opened 2019-07-16
Update tfjs-core to 1.2.3.

dsmilkov:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1842)


dsmilkov:
The tfjs-node integration fails as expected. Will be fixed in https://github.com/tensorflow/tfjs-node/pull/279 after the release

PR #1843, opened 2019-07-16
fixes camelcase CrossEntropy -> Crossentropy.

bileschi:
Fixes https://github.com/tensorflow/tfjs/issues/1743

Adds shims for existing API so as not to break existing users.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1843)

### DIFF
--- a/src/ops/loss_ops.ts
+++ b/src/ops/loss_ops.ts
@@ -252,14 +252,14 @@ function logLoss_<T extends Tensor, O extends Tensor>(
   return computeWeightedLoss(losses, $weights, reduction);
 }
 
-function sigmoidCrossEntropyWithLogits_<T extends Tensor, O extends Tensor>(
+function sigmoidCrossentropyWithLogits_<T extends Tensor, O extends Tensor>(
     labels: T|TensorLike, logits: T|TensorLike): O {
   const $labels =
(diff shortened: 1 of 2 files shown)


bileschi:
___
*[src/ops/loss_ops.ts, line 496 at r1](https://reviewable.io/reviews/tensorflow/tfjs-core/1843#-LjvqTW0BBiLWF9Bf4Qd:-LjvvQdD1f7jD0-0kDYD:bk02g9p) ([raw file](https://github.com/tensorflow/tfjs-core/blob/3fb910cbda7b7459960823a20c9a0348b6fad814/src/ops/loss_ops.ts#L496)):*

Ah - sorry. I overlooked that fact when commenting on the referred bug. 

If tf.losses.sigmoidCrossEntropy is the equivalent of https://www.tensorflow.org/api_docs/python/tf/losses/sigmoid_cross_entropy, then Nikhil is right, the current spelling is correct. I was thinking about this in the context of the keras API, which has crossentropy as one word.


</blockquote></details>

It seems we want to align with both TensorFlow and Keras, and thus are stuck in a pickle, since they don't align with each other.  Our options are:

1.  Drop this and keep the misalignment.  Close the bug as WAI
2.  Prefer one or the other
3.  Support both spellings of Crossentropy
4.  Replace all references to `xentropy`.  (jk)

PR #1844, opened 2019-07-16
add experimental webgl support for RN platform

tafsiri:
This PR adds experimental support for the WebGL backend on react native platform. It is experimental because it mocks out extension testing in the platform itself. This is needed until the upstream GL context supports the extension checking.

Additionally, it makes core check for EXT_color_buffer_half_float if EXT_color_buffer_float is not present.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1844)


tfjs-react-native/src/platform_react_native.ts
```diff
@@ -117,11 +120,68 @@ export class PlatformReactNative implements Platform {
...
+      const glContext = await GLView.createContextAsync();
+
+      // TODO Remove. TEMPORARY. Mock extension support for
+      // EXT_color_buffer_float and EXT_color_buffer_half_float
+      //@ts-ignore
+      const getExt = glContext.getExtension.bind(glContext);
+      // tslint:disable-next-line:only-arrow-functions
+      const shimGetExt = function(name: string) {
+        if (name === 'EXT_color_buffer_float') {
+          if (RNPlatform.OS === 'ios') {
```
nkreeger: Might be good to add a comment here about why this case is handled.

PR #1847, opened 2019-07-17
Security fix for cloud func mailer and react native

tafsiri:
Note that I took a different approach here for RN and removed it as a dev dependency (left it as a peer dependency) due to some surprising behaviour with yarn.lock when upgrading (easier to demonstrate in person). This gives us more time to look into rn .60 and lock file issues. 

Also i noticed we still had a warning pending for cloud funcs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-core/1847)

### DIFF
--- a/scripts/cloud_funcs/send_email/package-lock.json
+++ b/scripts/cloud_funcs/send_email/package-lock.json
@@ -450,9 +450,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.11",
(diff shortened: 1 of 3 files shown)


tafsiri:
@nsthorat yep (though note we don't install RN for them so they can use whatever versions that support expo-gl really). However i don't think its notable because it is a component used by the CLI not the app itself, so we can stay compatible with older versions that RN 6

